PR #11, opened 2014-04-30
PageUp/Down, Home/End key in Table(#12879)

imcharsi:
Hi.

At first, my english is not good.

This is just workaround about #12879.

Refer to http://dev.vaadin.com/ticket/12879

Follow is a testproject about this with comparing between original and modification.
https://github.com/imcharsi/vaadin_test_table20140429

There is still something strange behavior about PageUp/Down key even though with this workaround.

So I think it would be required more precise fix.

Thanks.


Artur-:

This might also be the same issue as http://dev.vaadin.com/ticket/15332, which was fixed this week

Thanks

PR #13, opened 2014-06-12
Provide a way to obtain Container Item from CalendarEvent

eXsio:
When overriding methods such as moveEvent or resizeEvent, or implementing EventClickHandler, EventRangeSelectHandler, one gets an instance of appropriate CalendarComponentEvent, which holds a reference to an instance of ContainerCalendarEvent, which is a private inner class (so one can't cast the event object to it). When using this provider combined with (for exapmle) JPAContainer, there is no way to obtain an instance of the real event (which can be an entity, managed by the container). This severely limits the possibilities of extending ContainerEventProvider.


Artur-:
Hi, as we are using Gerrit we cannot accept pull requests but need the changes to go through Gerrit. Please see https://vaadin.com/wiki/-/wiki/Main/Contributing+Code and follow the instructions there.

I'll close this pull request, please open one in Gerrit instead!

Thanks

PR #16, opened 2014-11-03
Update Table.java

kamiszczu:
Small typo


Artur-:

As this is a minor javadoc change I can also submit it through Gerrit myself but then you would not get credit for the fix as it would be in my name.

Thanks

elmot:
Already fixed in master branch. To be closed here.

PR #19, opened 2014-12-08
given my the set --> given by the set

rtack:
given by the set seems more correct even though it doesn't sound quite right.


Artur-:

As this is a minor javadoc change I can also submit it through Gerrit myself but then you would not get credit for the fix as it would be in my name.

Thanks

rtack:
Done with https://dev.vaadin.com/review/#/c/8259/

PR #23, opened 2015-01-26
Fix issue #16499: UI.getPushConfiguration().getTransport() *always* returns null

wodencafe:
Hello Vaadin team,

I stumbled upon this in the lastest Vaadin 7.3.9, Tomcat 8 and 7, etc.

UI.getPushConfiguration().getTransport() will always return null.

The problem has to do with using "Transport.valueOf" in the getTransport() method that is trying to look up the Transport enum based on it's identifier "websocket" rather than it's string equivalent name "WEBSOCKET". This applies to all of the transports.

Here is the relevant Vaadin ticket created in response to this issue: http://dev.vaadin.com/ticket/16499

Please try in your own project with Vaadin, any version, call UI.getPushConfiguration().getTransport() and see for yourself that it is always null.

The problem is caused by PushConfigurationImpl.getTransport() falling into the catch (IllegalArgumentException) that is then swallowed silently, and null is returned.

### DIFF
--- a/server/src/com/vaadin/ui/PushConfiguration.java
+++ b/server/src/com/vaadin/ui/PushConfiguration.java
@@ -208,7 +208,7 @@ class PushConfigurationImpl implements PushConfiguration {
     public Transport getTransport() {
         try {
             return Transport
-                    .valueOf(getParameter(PushConfigurationState.TRANSPORT_PARAM));
+                    .getByIdentifier(getParameter(PushConfigurationState.TRANSPORT_PARAM));
         } catch (IllegalArgumentException e) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


Artur-:
Hi, as we are using Gerrit we cannot accept pull requests but need the changes to go through Gerrit. Please see https://vaadin.com/wiki/-/wiki/Main/Contributing+Code and follow the instructions there.

Thanks

wodencafe:
Hi Artur,

I submitted the change and a Unit Test to the Gerrit system here:
https://dev.vaadin.com/review/#/c/7080/

Artur-:
Thanks, merged it!

PR #24, opened 2015-03-28
To get the correct path information

carrchang:
Sometimes can not get the correct path information, try to get from the servlet path.


Legioth:

carrchang:
I've tried to convert VaadinServlet to a SpringMVC Controller. When I use DispatcherServlet as my app's root servlet, request.getPathInfo() always equals null, so I made this change. 

I have pushed a demo in https://github.com/carrchang/vaadin-spring-integration, to demonstrate how the controller works.

I wish Vaadin UI as a spring view option, maybe this is a stupid idea. If you have time, please review the demo, and hope to give some advice.

Thank you!

PR #25, opened 2015-04-04
[TableQuery] RowId returned in RowIdChangeListener uses wrong primary

Flamenco:
This issue applies to at least MySql:

When creating a new row with a Long type primary key, the RowIdChangeListener returns a newRowId containing a Long type.

Calling container.getItem(newRowId) returns null instead of the item.

This is because SqlContainer creates a RowId with an Integer type.  The hashCode of the Integer does not match the hashCode of the Long and the lookup fails.

The newItemId should store the key as the same type that is stored in the database.

Here is the relevant code

``` java

/* Generate itemId for the row based on primary key(s) */

 Object[] itemId = new Object[pKeys.size()];

 for (int i = 0; i < pKeys.size(); i++) {

    itemId[i] = rs.getObject(pKeys.get(i));

 }

 id = new RowId(itemId);
```

rs.getObject is returning an Integer in my case, while RowIdChangeListener:newRowId is holding a Long.

So, at least for my database driver, the Statement.getGeneratedKeys result set does not return the same type as the key field.

In my opinion, the newRowId is incorrect and should downcast the Long to an Integer, as used by the SqlContainer.

### DIFF
--- a/server/src/com/vaadin/data/util/sqlcontainer/connection/SimpleJDBCConnectionPool.java
+++ b/server/src/com/vaadin/data/util/sqlcontainer/connection/SimpleJDBCConnectionPool.java
@@ -33,6 +33,7 @@ public class SimpleJDBCConnectionPool implements JDBCConnectionPool {
 
     private int initialConnections = 5;
     private int maxConnections = 20;
+    private Thread pingThread = null;
 
     private String driverName;
     private String connectionUri;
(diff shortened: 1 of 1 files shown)


Legioth:

elmot:
This patch does something completely different to a declared problem. To be closed here.

PR #27, opened 2015-05-05
fix setValue for TextFileProperty

travisfw:
This I/O operation was handled pretty poorly. Extra calls to close() and
not even in a finally block. Furthermore, it attempted to write into
the same file. Close now happens within finally block, and the value is
first written into a new file, which is then moved over the old file.
This is more in line with common best practice.

Not sure if I have fixed all the issues with this, and honestly, I haven't used Ivy so I don't know how to run the tests. Still, I imagine what I have written will work. You can probably just pull and test and it'll be fine. If not, I'm still open to helping.


travisfw:
Just bumping this. Is this contribution useful? Feel free to close the pull request and/or give feedback.

Legioth:

We are using Gerrit so we cannot accept pull requests here but need the changes to go through Gerrit. Please see https://vaadin.com/wiki/-/wiki/Main/Contributing+Code and follow the instructions there.

I also had a quick look at your patch and saw that you were using e.g. java.nio.file.Files from Java 7. We can unfortunately not accept patches that do not compile with Java 6 since we have customers that have not yet been able to upgrade their servers.

travisfw:
Thanks for getting back to me. I'll read that wiki page.

PR #28, opened 2015-05-12
Fix for trac-14802 : Correct style name for ValoTheme.PANEL_SCROLL_INDICATOR

robeden:
The correct style name for ValoTheme.PANEL_SCROLL_INDICATOR should be "scroll-divider".

https://dev.vaadin.com/ticket/14802


robeden:
The file is right, but let me clean up the commits shown. Needs a rebase...

PR #30, opened 2015-06-30
vaadin-client changed to 'provided' scope

banterCZ:
Since vaadin-client is not needed at runtime, its default scope could be provided, couldn't be? Thanks to that, it will not get to the war file by accident.


Legioth:

We are using Gerrit so we cannot accept pull requests here but need the changes to go through Gerrit. Please see https://vaadin.com/wiki/-/wiki/Main/Contributing+Code and follow the instructions there.

elmot:
Ported to Gerrit, to be closed here
https://dev.vaadin.com/review/#/c/12550/

PR #32, opened 2015-09-17
7.5

ola-mattsson:
- Since grid was not cancelling scrolling when multiple touches,
  zooming was not working, this patch fixes "zoom trap".
- Prevent default when moving the scroll in touch devices to avoid
  parent scrolling

Change-Id: I47856ef20bc835bb2b310b45f6d2749e950e8e7a


Artur-:
wat

PR #44, opened 2016-04-08
Update clientside-compiling.asciidoc

enver-haase:


magi42:
The `[guibutton]` style indeed is not quite suitable for the button in question, as the name refers to the textual name of the button (which can be seen in the tooltip), not the button caption as such.

However, I'd rather not have it as suggested in this PR either. The button caption is not `*.` and the appearance of a graphical element shouldn't really be described in text. I'll try to think something more suitable; it's probably best to refer to the button by its (tooltip) name and perhaps have a screenshot to illustrate it visually. There already is an annotated screenshot in the server-side integration chapter: `gwt/img/widgetset-compiling-toolbar-hi.png`.

Someone please close this.

PR #50, opened 2016-04-12
Fixed ConcurrentModificationException

ArneLimburg:
When contiuously writing sessions (i.e. for session replication) this code contiuously throws ConcurrentModificationExceptions. The proposed change fixes this.


Legioth:
I think the root cause of this issue should be fixed by https://dev.vaadin.com/ticket/19310 which was released about two months ago in Vaadin 7.6.3.

Could you please verify that your fix is still needed with Vaadin 7.6.3 or newer?

ArneLimburg:
We'll try and give you feedback

Legioth:
Any progress with trying out with Vaadin Framework 7.6.3 or newer?

ArneLimburg:
7.6.3 seems to work for us. For now I'll close this pull request.

PR #54, opened 2016-06-03
Update DefaultFieldFactory.java

adihubba:
In the Vaadin book there is a suggestion to extend from this class to implement an own factory. There is absolutely no reason why the two methods are static. If you remove the two static modifiers, it's a lot easier to extend from this class.

### DIFF
--- a/server/src/main/java/com/vaadin/ui/DefaultFieldFactory.java
+++ b/server/src/main/java/com/vaadin/ui/DefaultFieldFactory.java
@@ -75,7 +75,7 @@ public class DefaultFieldFactory implements FormFieldFactory, TableFieldFactory 
      * @param propertyId
      * @return the formatted caption string
      */
-    public static String createCaptionByPropertyId(Object propertyId) {
+    public String createCaptionByPropertyId(Object propertyId) {
         return SharedUtil.propertyIdToHumanFriendly(propertyId);
     }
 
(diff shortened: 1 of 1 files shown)


elmot:
Please follow the https://github.com/vaadin/vaadin/blob/master/CONTRIBUTING.md for code contribution. If you are implementing the patch for Vaadin 7, please push your change for "7.7" branch, not master, i.e.
use git push gerrit HEAD:refs/for/7.7. Also don't forget to file a ticket in trac and note its number in the commit message. I close the push request here.

hesara:
@adihubba  Thanks for the PR! Sorry us being a bit in the middle of our migration from Gerrit to GitHub, and not having the proper time to process this pull request.

Our contribution guide now looks quite different than last week. We now work completely in GitHub, and accept outside PR’s here as well 🎉

The tickets from our previous system (Trac) are also being migrated to GitHub Issues, but that’s probably going to happen later this week at the earliest.

So, coming back to this particular PR, it concerns DefaultFieldFactory which is actively used in Vaadin 7 but only in the compatibility package in Vaadin 8. Thus, I believe this change should probably go to the 7.7 branch in case it doesn't break API compatibility too much.

Could you please make a PR with these changes picked on the latest 7.7 branch, formulate the commit message as recommended in https://github.com/vaadin/framework/blob/master/CONTRIBUTING.md and accept the contribution agreement on GitHub (it will show up on the PR page after your next push). We could then give it a proper review and hopefully integrate the changes.

PR #58, opened 2016-07-06
Add null-check to setCellFocus Method

djbowman1:
Added if conditional in setCellFocus method to short circuit in the case that container is null, full reason available on ticket #20031

Change-Id: I646dd92dd3440440a0de39b1d7d3a66c543c31b4

### DIFF
--- a/client/src/main/java/com/vaadin/client/widgets/Grid.java
+++ b/client/src/main/java/com/vaadin/client/widgets/Grid.java
@@ -2526,6 +2526,9 @@ public class Grid<T> extends ResizeComposite implements
          */
         private void setCellFocus(int rowIndex, int columnIndexDOM,
                 RowContainer container) {
+            if (container == null){
+                return;
+            }
             if (rowIndex == rowWithFocus
                     && cellFocusRange.contains(columnIndexDOM)
                     && container == this.containerWithFocus) {


manolo:
@djbowman1 The process seems complicated because:
- You are sending a patch to master branch of Vaadin Framework, but vaadin-grid has it's own branch named feature/elements 
- Vaadin Framework only accepts patches via gerrit instead of PR in github

So there are two options:
- You send your patch directly to gerrit, which is could not be the best UX. You can take this patch I've sent today as reference: https://dev.vaadin.com/review/#/c/13798/
- Or I can cherry-pick this patch and send to gerrit in your behalf

djbowman1:
@manolo Hey thanks for your help, i got the changes put into a new pull request up on gerrit now. I would like to suggest revisiting your documentation on contributing to code, there are various parts that contradict that you may want to revise.

manolo:
@djbowman1 you are right we should say something about that we depend on the Grid implementation in the Vaadin Framework, and they have their own contribution rules and guides. 

Let close this PR in favour of the gerrit one. 
Thanks

manolo:
This PR has been moved to Gerrit: https://dev.vaadin.com/review/#/c/13807/

PR #67, opened 2016-08-29
Fixed event name to match example code

marcushellberg:
Value change event was called clickEvent, fixed to match code below.


Artur-:
Thanks, merged in 6214afcd5ff1c82f62ac2d1c56ff455f0e15b3c4

PR #81, opened 2016-12-02
Refresh in debug mode if servlet context is reinitialized

Legioth:
When developing a Vaadin application in an IDE, the development is often
done on a local development server configured to redeploy the
application whenever a .java file has been modified.

This patch introduces a refresh request that is used to detect when the
server has been restarted, and then subsequently reload the application
to use the newly deployed code.

The new functionality is only activated when the application is run in
debug mode, and it only works for requests originating from a localhost
address.

Fixes https://dev.vaadin.com/ticket/19102

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/81)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ApplicationConnection.java
+++ b/client/src/main/java/com/vaadin/client/ApplicationConnection.java
@@ -420,6 +420,9 @@ public class ApplicationConnection implements HasHandlers {
                 getUIConnector().getState().overlayContainerLabel);
         Roles.getAlertRole().setAriaRelevantProperty(overlayContainer,
                 RelevantValue.ADDITIONS);
+        if (ApplicationConfiguration.isDebugMode()) {
+            new RefreshRequestHandler(this).startRefreshRequest();
+        }
     }
 
     /**
--- a/client/src/main/java/com/vaadin/client/RefreshRequestHandler.java
+++ b/client/src/main/java/com/vaadin/client/RefreshRequestHandler.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2000-2016 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.client;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.google.gwt.http.client.Request;
+import com.google.gwt.http.client.RequestBuilder;
+import com.google.gwt.http.client.RequestCallback;
+import com.google.gwt.http.client.RequestException;
+import com.google.gwt.http.client.Response;
+import com.google.gwt.user.client.Timer;
+import com.google.gwt.user.client.Window;
+import com.vaadin.shared.ApplicationConstants;
+
+/**
+ * Handles automatic reloading after a server restart or redeploy when not
+ * running in production mode.
+ *
+ * @author Vaadin Ltd
+ */
+public class RefreshRequestHandler {
+    private ApplicationConnection connection;
+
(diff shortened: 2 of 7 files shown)


denis-anisimov:
Reviewed 7 of 7 files at r1.

---

*[server/src/main/java/com/vaadin/server/RefreshHandler.java, line 43 at r1](https://reviewable.io:443/reviews/vaadin/framework/81#-KXzomtqGmmLqyEt7L7L:-KXzomtqGmmLqyEt7L7M:b-j5z608) ([raw file](https://github.com/vaadin/framework/blob/ceb7354c7a5439a4286f3aa7331420972cfd1a5c/server/src/main/java/com/vaadin/server/RefreshHandler.java#L43)):*
> ```Java
> 
>     /*
>      * The token changes every time the refreshhandler is reloaded. This should
> ```

Does it ?
Most likely yes, but there is no guarantee that each time when `nextInt()` is called it returns a different value.

---

*[server/src/main/java/com/vaadin/server/RefreshHandler.java, line 67 at r1](https://reviewable.io:443/reviews/vaadin/framework/81#-KXzndZ3Kek7DfwED4aF:-KXzndZ3Kek7DfwED4aG:b-np6y0b) ([raw file](https://github.com/vaadin/framework/blob/ceb7354c7a5439a4286f3aa7331420972cfd1a5c/server/src/main/java/com/vaadin/server/RefreshHandler.java#L67)):*
> ```Java
>         // interface is used
>         if (!localhostAddresses
>                 .contains(remoteAddr.replaceAll("%[0-9]+$", ""))) {
> ```

Network interface is always the last part of the address, right ?
Isn't easier to remove a suffix from the address ?

Anyway, this code is executed on each request as I understand and each time the RegExp is compiled .
It's more effective to extract compiled Pattern as a constant and use pattern here.

---

*[server/src/main/java/com/vaadin/server/RefreshHandler.java, line 87 at r1](https://reviewable.io:443/reviews/vaadin/framework/81#-KXzq_mpUs0PHaBkTpmm:-KXzq_mpUs0PHaBkTpmn:b-gsahgw) ([raw file](https://github.com/vaadin/framework/blob/ceb7354c7a5439a4286f3aa7331420972cfd1a5c/server/src/main/java/com/vaadin/server/RefreshHandler.java#L87)):*
> ```Java
>                  * amount of time we're willing to let this request stay open
>                  */
>                 Thread.sleep(1000 * 60 * 5);
> ```

Apparently I don't understand this part of the code.

This just sleeps for 5 minutes each time when server token is the same as on the client side.
Doesn't it block all the response dispatching for 5 minutes ?

As I understand the workflow is :
- client has `null` token.
- it sends request to the server.
- server respond with some not `null` token.
- client saves it and request server one more time with token provided 
- server gets the token and sees that it's the same 
- now it sleeps for 5 minutes. It means that a response will be hanging for 5 minutes....

Am I wrong ?

---

*[server/src/main/java/com/vaadin/server/RefreshHandler.java, line 123 at r1](https://reviewable.io:443/reviews/vaadin/framework/81#-KXznYNHsAs3yTher9Kx:-KXznYNHsAs3yTher9Ky:b-d0wxhj) ([raw file](https://github.com/vaadin/framework/blob/ceb7354c7a5439a4286f3aa7331420972cfd1a5c/server/src/main/java/com/vaadin/server/RefreshHandler.java#L123)):*
> ```Java
>         return Logger.getLogger(RefreshHandler.class.getName());
>     }
> }
> ```

new line

---

*[uitest/src/main/java/com/vaadin/launcher/DevelopmentServerLauncher.java, line 191 at r1](https://reviewable.io:443/reviews/vaadin/framework/81#-KXzmzrLXvklk9ZFJECK:-KXzmzrLXvklk9ZFJECL:b9657n3) ([raw file](https://github.com/vaadin/framework/blob/ceb7354c7a5439a4286f3aa7331420972cfd1a5c/uitest/src/main/java/com/vaadin/launcher/DevelopmentServerLauncher.java#L191)):*
> ```Java
>          * session is persisted between restarts.
>          */
>         SessionHandler sessionHandler = webappcontext.getSessionHandler();
> ```

Why is this block of the code needed ?
Is it used anywhere ?

Is it possible by the way to write an automatic test for the PR ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/81)*

Legioth:
I removed the `waiting for author` label from this since the author won't do anything about this PR for now. Unless adopted by another author, this PR should simply be closed.

tsuoanttila:
Closing this PR, not merging as-is. Ticket described the issue as "annoyance", not urgent to get fixed.

PR #7037, opened 2016-12-10
Use thread-safe collections for VaadinService listeners (#19153)

Legioth:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/7037)

### DIFF
--- a/server/src/main/java/com/vaadin/server/VaadinService.java
+++ b/server/src/main/java/com/vaadin/server/VaadinService.java
@@ -25,7 +25,6 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
@@ -37,6 +36,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
@@ -51,7 +52,6 @@ import javax.servlet.ServletContext;
 import javax.servlet.http.HttpServletResponse;
 
 import com.vaadin.annotations.PreserveOnRefresh;
-import com.vaadin.event.EventRouter;
 import com.vaadin.server.VaadinSession.FutureAccess;
 import com.vaadin.server.VaadinSession.State;
 import com.vaadin.server.communication.AtmospherePushConnection;
@@ -66,7 +66,6 @@ import com.vaadin.shared.Registration;
 import com.vaadin.shared.ui.ui.UIConstants;
 import com.vaadin.ui.UI;
 import com.vaadin.util.CurrentInstance;
-import com.vaadin.util.ReflectTools;
 
 import elemental.json.Json;
 import elemental.json.JsonException;
@@ -101,17 +100,6 @@ public abstract class VaadinService implements Serializable {
     @Deprecated
     static final String REINITIALIZING_SESSION_MARKER = PRESERVE_UNBOUND_SESSION_ATTRIBUTE;
 
-    private static final Method SESSION_INIT_METHOD = ReflectTools.findMethod(
-            SessionInitListener.class, "sessionInit", SessionInitEvent.class);
-
-    private static final Method SESSION_DESTROY_METHOD = ReflectTools
-            .findMethod(SessionDestroyListener.class, "sessionDestroy",
-                    SessionDestroyEvent.class);
-
-    private static final Method SERVICE_DESTROY_METHOD = ReflectTools
-            .findMethod(ServiceDestroyListener.class, "serviceDestroy",
-                    ServiceDestroyEvent.class);
-
     /**
      * @deprecated As of 7.0. Only supported for {@link LegacyApplication}.
      */
@@ -128,7 +116,18 @@ public abstract class VaadinService implements Serializable {
 
     private final DeploymentConfiguration deploymentConfiguration;
 
-    private final EventRouter eventRouter = new EventRouter();
+    /*
+     * Can't use EventRouter for these listeners since it's not thread safe. One
+     * option would be to use an EventRouter instance guarded with a lock, but
+     * then we would needlessly hold a "global" lock while invoking potentially
+     * slow listener implementations.
+     */
+    private final Set<ServiceDestroyListener> serviceDestroyListeners = Collections
+            .newSetFromMap(new ConcurrentHashMap<>());
+
+    private final List<SessionInitListener> sessionInitListeners = new CopyOnWriteArrayList<>();
+
+    private final List<SessionDestroyListener> sessionDestroyListeners = new CopyOnWriteArrayList<>();
 
     private SystemMessagesProvider systemMessagesProvider = DefaultSystemMessagesProvider
             .get();
@@ -445,8 +444,8 @@ public abstract class VaadinService implements Serializable {
      * @return a registration object for removing the listener
      */
     public Registration addSessionInitListener(SessionInitListener listener) {
-        return eventRouter.addListener(SessionInitEvent.class, listener,
-                SESSION_INIT_METHOD);
+        sessionInitListeners.add(listener);
+        return () -> sessionInitListeners.remove(listener);
     }
 
     /**
@@ -463,8 +462,7 @@ public abstract class VaadinService implements Serializable {
      */
     @Deprecated
     public void removeSessionInitListener(SessionInitListener listener) {
-        eventRouter.removeListener(SessionInitEvent.class, listener,
-                SESSION_INIT_METHOD);
+        sessionInitListeners.remove(listener);
     }
 
     /**
@@ -482,8 +480,8 @@ public abstract class VaadinService implements Serializable {
      */
     public Registration addSessionDestroyListener(
             SessionDestroyListener listener) {
-        return eventRouter.addListener(SessionDestroyEvent.class, listener,
-                SESSION_DESTROY_METHOD);
+        sessionDestroyListeners.add(listener);
+        return () -> sessionDestroyListeners.remove(listener);
     }
 
     /**
@@ -517,12 +515,20 @@ public abstract class VaadinService implements Serializable {
                     session.removeUI(ui);
                 });
             }
-            // for now, use the session error handler; in the future, could
-            // have an API for using some other handler for session init and
-            // destroy listeners
-            eventRouter.fireEvent(
-                    new SessionDestroyEvent(VaadinService.this, session),
-                    session.getErrorHandler());
+            SessionDestroyEvent event = new SessionDestroyEvent(
+                    VaadinService.this, session);
+            for (SessionDestroyListener listener : sessionDestroyListeners) {
+                try {
+                    listener.sessionDestroy(event);
+                } catch (Exception e) {
+                    /*
+                     * for now, use the session error handler; in the future,
+                     * could have an API for using some other handler for
+                     * session init and destroy listeners
+                     */
+                    session.getErrorHandler().error(new ErrorEvent(e));
+                }
+            }
             session.setState(State.CLOSED);
         });
     }
@@ -540,8 +546,7 @@ public abstract class VaadinService implements Serializable {
      */
     @Deprecated
     public void removeSessionDestroyListener(SessionDestroyListener listener) {
-        eventRouter.removeListener(SessionDestroyEvent.class, listener,
-                SESSION_DESTROY_METHOD);
+        sessionDestroyListeners.remove(listener);
     }
 
     /**
@@ -851,12 +856,19 @@ public abstract class VaadinService implements Serializable {
 
     private void onVaadinSessionStarted(VaadinRequest request,
             VaadinSession session) throws ServiceException {
-        // for now, use the session error handler; in the future, could have an
-        // API for using some other handler for session init and destroy
-        // listeners
-
-        eventRouter.fireEvent(new SessionInitEvent(this, session, request),
-                session.getErrorHandler());
+        SessionInitEvent event = new SessionInitEvent(this, session, request);
+        for (SessionInitListener listener : sessionInitListeners) {
+            try {
+                listener.sessionInit(event);
+            } catch (Exception e) {
+                /*
+                 * for now, use the session error handler; in the future, could
+                 * have an API for using some other handler for session init and
+                 * destroy listeners
+                 */
+                session.getErrorHandler().error(new ErrorEvent(e));
+            }
+        }
 
         ServletPortletHelper.checkUiProviders(session, this);
     }
@@ -1888,6 +1900,10 @@ public abstract class VaadinService implements Serializable {
     /**
      * Adds a service destroy listener that gets notified when this service is
      * destroyed.
+     * <p>
+     * The listeners may be invoked in a non-deterministic order. In particular,
+     * it is not guaranteed that listeners will be invoked in the order they
+     * were added.
      *
      * @since 7.2
      * @param listener
@@ -1900,8 +1916,8 @@ public abstract class VaadinService implements Serializable {
      */
     public Registration addServiceDestroyListener(
             ServiceDestroyListener listener) {
-        return eventRouter.addListener(ServiceDestroyEvent.class, listener,
-                SERVICE_DESTROY_METHOD);
+        serviceDestroyListeners.add(listener);
+        return () -> serviceDestroyListeners.remove(listener);
     }
 
     /**
@@ -1917,8 +1933,7 @@ public abstract class VaadinService implements Serializable {
      */
     @Deprecated
     public void removeServiceDestroyListener(ServiceDestroyListener listener) {
-        eventRouter.removeListener(ServiceDestroyEvent.class, listener,
-                SERVICE_DESTROY_METHOD);
+        serviceDestroyListeners.remove(serviceDestroyListeners);
     }
 
     /**
@@ -1933,7 +1948,9 @@ public abstract class VaadinService implements Serializable {
      * @since 7.2
      */
     public void destroy() {
-        eventRouter.fireEvent(new ServiceDestroyEvent(this));
+        ServiceDestroyEvent event = new ServiceDestroyEvent(this);
+        serviceDestroyListeners
+                .forEach(listener -> listener.serviceDestroy(event));
     }
 
     /**


Legioth:
Original reivew: https://dev.vaadin.com/review/#/c/12351

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037#-:-KYdiFSoNzgdwZPCTax3:b-7z5q2q)*

pleku:
:lgtm: , but @denis-anisimov you had some comments open in review ? 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037#-:-KYlg6JT4INEmAKboBX8:b-kpst18)*

denis-anisimov:
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 521 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYlvJrUY6X2iZJjoVZL:-KYlvJrVv7Hk8WC9zcSL:bkmi5nm) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L521)):*
> ```Java
>             SessionDestroyEvent event = new SessionDestroyEvent(
>                     VaadinService.this, session);
>             for (SessionDestroyListener l : new ArrayList<>(
> ```

Just a note: I don't like the approach used almost everywhere in event firing methods.
Adding/removal listeners is a rare operation in most of cases comparing firing an event.
So copy (copy-on-write) should be done there (listener mutation) instead of here.

In this specific case there will be no ConcurrentModificationException at all if listener has been added/removed during navigation through the set (because of the nature of the set). 
But theoretically it may lead  to infinite execution of such navigation.......

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 525 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYltXYggk1yE4gO2EuM:-KYltXYhTbfRBfQmp7uG:b-lum1x0) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L525)):*
> ```Java
>                 try {
>                     l.sessionDestroy(event);
>                 } catch (Exception e) {
> ```

`Throwable` then.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 865 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYlv-gZU4dcQZE-Z90u:-KYlv-gZU4dcQZE-Z90v:b-2y29hh) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L865)):*
> ```Java
>             try {
>                 l.sessionInit(event);
>             } catch (Exception e) {
> ```

`Throwable`

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1953 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYlv9k3TKggnsIChckb:-KYlv9k4DLluxb2jzPBj:b4bhj8x) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L1953)):*
> ```Java
>     public void destroy() {
>         ServiceDestroyEvent event = new ServiceDestroyEvent(this);
>         new ArrayList<>(serviceDestroyListeners)
> ```

Has no sense at this point to catch throwables ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

Legioth:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 521 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYlvJrUY6X2iZJjoVZL:-KYo4WNva4Dc7B9hUDVi:b-qy2iqd) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L521)):*

Just a note: I don't like the approach used almost everywhere in event firing methods.
Adding/removal listeners is a rare operation in most of cases comparing firing an event.
So copy (copy-on-write) should be done there (listener mutation) instead of here.

In this specific case there will be no ConcurrentModificationException at all if listener has been added/removed during navigation through the set (because of the nature of the set). 
But theoretically it may lead  to infinite execution of such navigation.......
</blockquote></details>

You're right. There's no benefit at all from doing copies thanks to `ConcurrentHashMap`

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 525 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYltXYggk1yE4gO2EuM:-KYo4jprU-7m5UJ3ox68:b-6zazxu) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L525)):*

`Throwable` then.
</blockquote></details>

`EventRouter.fireEvent` with a non-null `errorHandler` catches `Exception`, so to be maintain old behavior, this should also catch `Exception`.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 865 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYlv-gZU4dcQZE-Z90u:-KYo55bfns9lc8hIAHxk:b-6zazxu) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L865)):*

`Throwable`
</blockquote></details>

`EventRouter.fireEvent` with a non-null `errorHandler` catches `Exception`, so to be maintain old behavior, this should also catch `Exception`.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1953 at r1](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYlv9k3TKggnsIChckb:-KYo571BzRdrtMyvHAML:b-ljplnt) ([raw file](https://github.com/vaadin/framework/blob/e7c0151360892a95e93c3903b06e91ab4be18536/server/src/main/java/com/vaadin/server/VaadinService.java#L1953)):*

Has no sense at this point to catch throwables ?
</blockquote></details>

Would maybe make sense, but the old implementation didn't work in that way since it used the `fireEvent` overload that didn't receive any `errorHandler`. That behavior is retained by this bugfix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

denis-anisimov:
Reviewed 1 of 1 files at r2.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 521 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYr2d7STaBoKcgVKSe7:-KYr2d7TMr_OS-4MoyJP:b-ky6ae4) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L521)):*
> ```Java
>             SessionDestroyEvent event = new SessionDestroyEvent(
>                     VaadinService.this, session);
>             for (SessionDestroyListener l : sessionDestroyListeners) {
> ```

One letter variable name.
Please don't.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 861 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYr2yvs4qNR6fZQGl2-:-KYr2yvs4qNR6fZQGl20:b-jab486) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L861)):*
> ```Java
>             VaadinSession session) throws ServiceException {
>         SessionInitEvent event = new SessionInitEvent(this, session, request);
>         for (SessionInitListener l : sessionInitListeners) {
> ```

One letter variable name.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1952 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYr2tSB_ULotWW2wMe1:-KYr2tSB_ULotWW2wMe2:b-jab486) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L1952)):*
> ```Java
>     public void destroy() {
>         ServiceDestroyEvent event = new ServiceDestroyEvent(this);
>         serviceDestroyListeners.forEach(l -> l.serviceDestroy(event));
> ```

One letter variable name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

mstahv:
I found this pull request via an email from customer who has faced this issue with their V7 prototype. The original issue behind this is a really critical to fix, I'd hurry the fix to V7 series as well.

I quickly looked through the patch. It looks to me that you are making the framework work in indeterministic manner. That looks very scary to me, I bet my good old friend @jani-laakso would rip his pants because of this. This kind of framework should work in deterministic way, always, even if there was a tiny performance hit. I know there is already some non-deterministic behaviour in client side rendering, please, let's not introduce such to the more critical server side code.

elmot:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ1hJqgrlhar5cFJiaj:b-kwkg0i) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*
> ```Java
> 
>     private final Set<SessionDestroyListener> sessionDestroyListeners = Collections
>             .newSetFromMap(new ConcurrentHashMap<>());
> ```

ConcurrentHashMap does not keep listeners in order of addition. That could have heavy consequences in customer applications where listeners are supposed to be called in particular order.
I think, we can use CopyOnWriteArraySet here. @mstahv, is that enough for you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ1lcHwtDdYozskFlWX:b-y5tfum) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

ConcurrentHashMap does not keep listeners in order of addition. That could have heavy consequences in customer applications where listeners are supposed to be called in particular order.
I think, we can use CopyOnWriteArraySet here. @mstahv, is that enough for you?
</blockquote></details>

No one ever can assume listeners order. This is totally wrong assumption. And totally bad design of the application.

If the order matter then special explicit design has to be introduced which express this fact.

Just because listeners cannot know about each other. If they don't know about each other then the order does't matter.
If they know then it should be one listener.
Everything else is just a hack.

It can be considered as a compatibility issue but there should be very serious reasons to take this into account than just theoretical reasoning.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

elmot:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ1y15KpRUgduLuFZ24:b-yjuqcc) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

No one ever can assume listeners order. This is totally wrong assumption. And totally bad design of the application.

If the order matter then special explicit design has to be introduced which express this fact.

Just because listeners cannot know about each other. If they don't know about each other then the order does't matter.
If they know then it should be one listener.
Everything else is just a hack.

It can be considered as a compatibility issue but there should be very serious reasons to take this into account than just theoretical reasoning.
</blockquote></details>

In reality, there are lots of cases when app developers initialize stuff one-by-one, relying some other objects were initialized before. Sad but true.
 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ1ykbPTwD10jXqD3UV:b-emkl3d) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

In reality, there are lots of cases when app developers initialize stuff one-by-one, relying some other objects were initialized before. Sad but true.
 
</blockquote></details>

Agreed.

But this is their fault definitely.

And again: to be able to support this feature it should be designed appropriately ( there are a lot questions. e.g. if the order matters then there should be a way to INSERT. Otherwise why it preserve the order but I cannot influence on it ? A lot of logical issues).

But here there is one more strong argument against it: the whole idea here is the  THREAD SAFETY.
How can you guarantee ANY order at all if I can add them from any thread ?

You cannot assume that two listeners added one after another from THE SAME thread will be one after another in the listeners collection.
It's just not possible exactly because of threads.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

elmot:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ229uvjFljteZe9MQf:b-9r8g3c) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

Agreed.

But this is their fault definitely.

And again: to be able to support this feature it should be designed appropriately ( there are a lot questions. e.g. if the order matters then there should be a way to INSERT. Otherwise why it preserve the order but I cannot influence on it ? A lot of logical issues).

But here there is one more strong argument against it: the whole idea here is the  THREAD SAFETY.
How can you guarantee ANY order at all if I can add them from any thread ?

You cannot assume that two listeners added one after another from THE SAME thread will be one after another in the listeners collection.
It's just not possible exactly because of threads.
</blockquote></details>

In case of inter-listener dependencies, its enough to suppose if a developer
 added listener1 then listener2 from the same thread, then listener2 is always called after listener1. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ5NrIlL6Wl1e_1REgL:b-7xmxjl) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

In case of inter-listener dependencies, its enough to suppose if a developer
 added listener1 then listener2 from the same thread, then listener2 is always called after listener1. 
</blockquote></details>

It means that he definitely can combine those listeners as a one.

It makes some sense but very little.
I'm still not convinced to provide reliable bad behavior for those who are writing a bad code.

I'm arguing here not against the final solution for this one issue but rather about the common approach: if you are going to support the order then provide an explicit API design for that.

I'm OK with CopyOnWrite.

Up to Leif.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

Legioth:
The actual problem that this doesn't even address is that our Spring add-on adds a separate listener for every user.

Please see the original discussion in gerrit (https://dev.vaadin.com/review/#/c/12351) about the non-determinism. The main issue is that you'd need to use `CopyOnWriteArrayList` instead of `ConcurrentHashMap` to keep the order deterministic, and that in turn has some unintended performance consequences, especially with the kind of abuse that the Spring addon does.

---

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ5ayPmQtWFTgFvNssd:btmuyln) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

It means that he definitely can combine those listeners as a one.

It makes some sense but very little.
I'm still not convinced to provide reliable bad behavior for those who are writing a bad code.

I'm arguing here not against the final solution for this one issue but rather about the common approach: if you are going to support the order then provide an explicit API design for that.

I'm OK with CopyOnWrite.

Up to Leif.
</blockquote></details>

I'm not ok with `CopyOnWriteArrayList` because it slows down with many listeners, which is exactly the case in the Spring add-on that triggered this ticket to be filed in the first place.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037#-:-KZ5aRJjpeCBo7dh6I8N:b-gmn4r3)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-KZ5cgk6Lw-U6aS9d27Z:bzgf2rz) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

I'm not ok with `CopyOnWriteArrayList` because it slows down with many listeners, which is exactly the case in the Spring add-on that triggered this ticket to be filed in the first place.
</blockquote></details>

https://github.com/ben-manes/concurrentlinkedhashmap ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

mstahv:
Sorry I didn't answer sooner here, had the repo on ignoring during github transition.

@denis-anisimov, deterministic behaviour is not important "feature" for developer when everything goes ok, but it is a lifesafer for them and you when starting to investigating issues. Multithreading in this case indeed throws even more randomness for the hassle. 

Has somebody actaully tried the performance characteristics of various options: ConcurrentSkipListSet, ConcurrentHashSet, CopyOnWriteArraySet? I'd bet the difference is irrelevant in most cases. More relevant would be to get this critical bug and related optimisation in Vaadin Spring landed before 8.0. 

CopyOnWriteArraySet would most likely be fast solution in most apps as the related listeners are mostly static during the application lifecycle. The way how Vaadin Spring uses is probably really rare. If it isn't, there should probably be some sort of destroyed listener one could add directly to a specific VaadinSession.

Legioth:
The Spring add-on can easily be fixed to not add and remove session init and destroy listeners individually for each session, but there is no good way to avoid doing the same for the service destroy listeners.

Based on that, I'd suggest that `sessionInitListeners` and `sessionDestoryListeners` are implemented to use a copy-on-write implementation with deterministic iteration order but `O(n)` for additions and removes, while `serviceDestroyListeners` would use `ConcurrentHashMap`that doesn't offer deterministic ordering, but `O(1)` for additions and removes. The difference between those would have to be very clearly documented.

What do you think?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037#-:-KawkT8hZKxGUCRtGPyD:bl7adf1)*

denis-anisimov:
I'm OK.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037#-:-Kb9bxcN6zfZkieJ0ScL:b-9c6udg)*

Legioth:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-Kb_CPduROo1gFtDo9FS:b-ioulcj) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*
> https://github.com/ben-manes/concurrentlinkedhashmap ?

Hasn't been updated in almost two years and there's a ticket about compatibility issues with JDK9. The readme recommends using Caffeine (https://github.com/ben-manes/caffeine) instead. The latest Caffeine .jar file on Maven central is 972kb, which would directly affect the size of the .war file for all our users.

As mentioned in the top-level discussion, I'd suggest using `CopyOnWriteArrayList` for the session listeners but continue relying on `ConcurrentHashMap` for the service destroy listener.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 521 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYr2d7STaBoKcgVKSe7:-Kb_BTqtvFeUGPccZ-PG:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L521)):*

One letter variable name.
Please don't.
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 861 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYr2yvs4qNR6fZQGl2-:-Kb_BX4g9tu4JP0qhDHA:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L861)):*

One letter variable name.
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1952 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KYr2tSB_ULotWW2wMe1:-Kb_BZe-wjY56IurCm7V:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L1952)):*

One letter variable name.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

denis-anisimov:
Reviewed 1 of 1 files at r3.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 131 at r2](https://reviewable.io:443/reviews/vaadin/framework/7037#-KZ1hJqgrlhar5cFJiai:-Kbi6XnI2UjillGvVHAV:b32w0x) ([raw file](https://github.com/vaadin/framework/blob/795ecb061fdcd913475a9fda41c922b0bc9e6bb1/server/src/main/java/com/vaadin/server/VaadinService.java#L131)):*

> https://github.com/ben-manes/concurrentlinkedhashmap ?

Hasn't been updated in almost two years and there's a ticket about compatibility issues with JDK9. The readme recommends using Caffeine (https://github.com/ben-manes/caffeine) instead. The latest Caffeine .jar file on Maven central is 972kb, which would directly affect the size of the .war file for all our users.

As mentioned in the top-level discussion, I'd suggest using `CopyOnWriteArrayList` for the session listeners but continue relying on `ConcurrentHashMap` for the service destroy listener.
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/7037#-KcCusScfUVOEn-XCl_b:-KcCusScfUVOEn-XCl_c:bfwufv4) (no related file):*
Ticket id in the commit message is for dev.vaadin.com. The corresponding ticket on github as id #7250.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037)*

elmot:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7037#-:-KcM_hYISxhr4o7WIw-r:bnfp4nl)*

PR #7962, opened 2016-12-12
Unified listeners

elmot:
DRAFT Listeners unification proposal

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/7962)


server/src/main/java/com/vaadin/event/ListenerMethod.java
```diff
@@ -198,7 +198,7 @@ public ListenerMethod(Class<?> eventType, Object target, Method method,
                         .isAssignableFrom(eventType)) {
             throw new java.lang.IllegalArgumentException(
                     "The method " + method.getName()
-                            + " does not accept the given eventType: "
+                            + " does not onEvent the given eventType: "
```
mstahv: I guess somebody has has done some find and replace here.

Legioth:
Review status: 0 of 33 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/7962#-KYqtmxbGfiIo54E9Qxc:-KYqtmxbGfiIo54E9Qxd:b-hlvzl6) (no related file):*
How  does this change deal with the fact that `Component.Listener` is automatically imported in all classes that implement that interface?

For this kind of monster patch, I also feel it's imperative to describe the purpose and the main points of the change in the commit message so that e.g. reviewers and anyone looking at the patch later on won't have to read through lots of changes just to figure out what the patch actually tries to achieve.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7962)*

hesara:
This patch exists to get early feedback about the approach and the naming, mostly from outside the framework team. The team will review this, fix the details and the commit message and submit this if there are no objections to the approach itself (replacing many specific listener interfaces with a generic one parametrized with the event type, which implies moving documentation to the addXyzListener methods and potentially the event classes). Also of interest is whether extending Consumer makes sense or not (the current feeling is that it is not useful in this case). @Legioth @mstahv

Legioth:
A summary of the core ideas of the change together with a couple of examples and a link to a full diff for more details would be much easier to process than to scroll through over 300 lines spread out over 33 different files.

In my opinion, the name `Listener` is completely broken since it conflicts with `Component.Listener` (which is also apparent as fully qualified names throughout this patch).

My gut feeling is also that it doesn't make sense to extend `Consumer` here, nor in almost any other case where a custom callback type is explicitly defined.

---

Review status: 0 of 33 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7962#-:-KYrIX5R1FTcurFUUyOj:b-tmnhkb)*

elmot:
Abandoned. We can't implement this in Vaadin 8 during reasonable time.

PR #7963, opened 2016-12-13
Combine and rename setFilter and applyFilter

Legioth:
setFilter is an unsuitable name since it looks like a setter even though
it isn't one. It's therefore renamed to withFilter.

The difference between setFilter and applyFilter is very confusing. The
original reason for the distinction was that they had incompatible
return types (Void vs F), but this is resolved by changing the
non-appendable version to use ? instead of Void.

One side effect of this change is that setFilter cannot be used on an
appendable filter provider to create a wrapped data provider without
further filtering support. This theorethical case is still supported by
using convertFilter with a Void -> F converter.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/7963)

### DIFF
--- a/server/src/main/java/com/vaadin/server/data/AppendableFilterDataProvider.java
+++ b/server/src/main/java/com/vaadin/server/data/AppendableFilterDataProvider.java
@@ -37,7 +37,8 @@ public interface AppendableFilterDataProvider<T, F> extends DataProvider<T, F> {
      *            the applied filter; not {@code null}
      * @return new data provider with the filter applied
      */
-    public default AppendableFilterDataProvider<T, F> applyFilter(F filter) {
+    @Override
+    public default AppendableFilterDataProvider<T, F> withFilter(F filter) {
         Objects.requireNonNull(filter, "The applied filter can't be null");
         return DataProviderWrapper.chain(this, filter);
     }
--- a/server/src/main/java/com/vaadin/server/data/DataProvider.java
+++ b/server/src/main/java/com/vaadin/server/data/DataProvider.java
@@ -100,7 +100,7 @@ public interface DataProvider<T, F> extends Serializable {
      *            the default filter
      * @return wrapped data provider with provided filter
      */
-    public default DataProvider<T, Void> setFilter(F filter) {
+    public default DataProvider<T, ?> withFilter(F filter) {
         Objects.requireNonNull(filter, "Filter can't be null");
         return DataProviderWrapper.filter(this, filter);
     }
--- a/server/src/main/java/com/vaadin/server/data/DataProviderWrapper.java
+++ b/server/src/main/java/com/vaadin/server/data/DataProviderWrapper.java
@@ -125,7 +125,7 @@ public abstract class DataProviderWrapper<T, F, M>
      * This {@code DataProvider} will deliberately ignore any possible filters
      * from the Query.
      *
-     * @see DataProvider#setFilter(Object)
+     * @see DataProvider#withFilter(Object)
      *
      * @param dataProvider
      *            the underlying data provider
@@ -145,6 +145,8 @@ public abstract class DataProviderWrapper<T, F, M>
 
             @Override
             protected F getFilter(Query<T, Void> query) {
+                assert !query.getFilter()
+                        .isPresent() : "A filter is provided to a data provider that should be unfilterable";
                 return filter;
             }
         };
--- a/server/src/test/java/com/vaadin/server/data/provider/DataProviderTestBase.java
+++ b/server/src/test/java/com/vaadin/server/data/provider/DataProviderTestBase.java
@@ -66,7 +66,7 @@ public abstract class DataProviderTestBase<D extends DataProvider<StrBean, Seria
     private Query<StrBean, SerializablePredicate<StrBean>> createQuery(
             List<SortOrder<String>> sortOrder, Comparator<StrBean> comp,
             SerializablePredicate<StrBean> filter) {
-        return new Query<StrBean, SerializablePredicate<StrBean>>(0,
+        return new Query<>(0,
                 Integer.MAX_VALUE, sortOrder, comp, filter);
     }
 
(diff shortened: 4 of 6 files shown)


hesara:
Reviewed 5 of 6 files at r1.
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[server/src/main/java/com/vaadin/server/data/DataProvider.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYrGsl9M8gdFhH0gg7X:-KYrGsl9M8gdFhH0gg7Y:b-vbvp6v) ([raw file](https://github.com/vaadin/framework/blob/24748ffa83392445986b0dc4e58ad5e93298b83c/server/src/main/java/com/vaadin/server/data/DataProvider.java#L97)):*
> ```Java
>      * Returns a new wrapped data provider with given default filter. Default
>      * filter will apply to each query and further filtering will not be
>      * supported in the wrapped data provider.
> ```

This is where I'm not quite sure if the javadoc and "?" in the return type are enough to indicate that the result cannot be filtered further - any opinion from @tsuoanttila? Could also add a comment about AFDP supporting filtering of the returned provider.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7963)*

Legioth:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[server/src/main/java/com/vaadin/server/data/DataProvider.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYrGsl9M8gdFhH0gg7X:-KYrM85yp_pXkYJzhdUl:b-vm536z) ([raw file](https://github.com/vaadin/framework/blob/24748ffa83392445986b0dc4e58ad5e93298b83c/server/src/main/java/com/vaadin/server/data/DataProvider.java#L97)):*

This is where I'm not quite sure if the javadoc and "?" in the return type are enough to indicate that the result cannot be filtered further - any opinion from @tsuoanttila? Could also add a comment about AFDP supporting filtering of the returned provider.
</blockquote></details>

The way I see it, both `?` and `Void` do in practice mean that further filtering cannot be applied. You're right that `Void` is a clearer marker, but at least to me, the reduced confusion from combining `setFilter` and `applyFilter` is far more important than how unfilterability is marked.

What exactly do you mean with the AFDP comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7963)*

denis-anisimov:
Reviewed 4 of 6 files at r1, 2 of 2 files at r2.

---

*[server/src/main/java/com/vaadin/server/data/DataProvider.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYrGsl9M8gdFhH0gg7X:-KYr_1paMUY0wHcnTdF9:b-eqq5z8) ([raw file](https://github.com/vaadin/framework/blob/24748ffa83392445986b0dc4e58ad5e93298b83c/server/src/main/java/com/vaadin/server/data/DataProvider.java#L97)):*

The way I see it, both `?` and `Void` do in practice mean that further filtering cannot be applied. You're right that `Void` is a clearer marker, but at least to me, the reduced confusion from combining `setFilter` and `applyFilter` is far more important than how unfilterability is marked.

What exactly do you mean with the AFDP comment?
</blockquote></details>

I agree with @hesara .
`?` doesn't mean for me anything except that type is unknown (and can be any).
`Void` has no any optional meaning at all.

If there is no any other reason to have it as `?` then `Void` has to be preserved.

---

*[server/src/test/java/com/vaadin/server/data/provider/DataProviderTestBase.java, line 300 at r2](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYraMOuDK5UTon16kry:-KYraMOvv3W2ME_9cLxD:b-k0hgc) ([raw file](https://github.com/vaadin/framework/blob/2a52bce41799dd1b766a17acc1aaacd45ea9bfdd/server/src/test/java/com/vaadin/server/data/provider/DataProviderTestBase.java#L300)):*
> ```Java
>     }
> 
>     @Test
> ```

Why is this removed ?
No any replacement for the test ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7963)*

Legioth:

---

*[server/src/main/java/com/vaadin/server/data/DataProvider.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYrGsl9M8gdFhH0gg7X:-KYs4KqbRIFeMFdC3DVF:bhhm0a0) ([raw file](https://github.com/vaadin/framework/blob/24748ffa83392445986b0dc4e58ad5e93298b83c/server/src/main/java/com/vaadin/server/data/DataProvider.java#L97)):*

I agree with @hesara .
`?` doesn't mean for me anything except that type is unknown (and can be any).
`Void` has no any optional meaning at all.

If there is no any other reason to have it as `?` then `Void` has to be preserved.
</blockquote></details>

The reason for `?` instead of `Void` is that `?` is compatible with the return type that the more specific implementation in `AppendableFilterDataProvider`. Without loosening this to `?`, the corresponding functionality in `AppendableFilterDataProvider` would have to have a differently named method for doing the same thing. And that would in my opinion be much more confusing than the distinction between `Void` and `?` in this case.

`?` does not mean that anything is accepted, it means that it's not possible to know what would be accepted.

---

*[server/src/test/java/com/vaadin/server/data/provider/DataProviderTestBase.java, line 300 at r2](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYraMOuDK5UTon16kry:-KYs57RMVHEthUcDcKFL:b-kqqz90) ([raw file](https://github.com/vaadin/framework/blob/2a52bce41799dd1b766a17acc1aaacd45ea9bfdd/server/src/test/java/com/vaadin/server/data/provider/DataProviderTestBase.java#L300)):*

Why is this removed ?
No any replacement for the test ?
</blockquote></details>

Explained in the commit message:

```
One side effect of this change is that setFilter cannot be used on an
appendable filter provider to create a wrapped data provider without
further filtering support. This theorethical case is still supported by
using convertFilter with a Void -> F converter.
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7963)*

hesara:
@tsuoanttila had objections to this approach - probably best discussed offline when he is available.

Legioth:

---

*[server/src/main/java/com/vaadin/server/data/DataProvider.java, line 97 at r1](https://reviewable.io:443/reviews/vaadin/framework/7963#-KYrGsl9M8gdFhH0gg7X:-KYwiI8kehj2WUakvKiS:bwwt8pi) ([raw file](https://github.com/vaadin/framework/blob/24748ffa83392445986b0dc4e58ad5e93298b83c/server/src/main/java/com/vaadin/server/data/DataProvider.java#L97)):*

The reason for `?` instead of `Void` is that `?` is compatible with the return type that the more specific implementation in `AppendableFilterDataProvider`. Without loosening this to `?`, the corresponding functionality in `AppendableFilterDataProvider` would have to have a differently named method for doing the same thing. And that would in my opinion be much more confusing than the distinction between `Void` and `?` in this case.

`?` does not mean that anything is accepted, it means that it's not possible to know what would be accepted.
</blockquote></details>

We'll do a slightly bigger redesign that also fixes this issue, but it's not sure we'll get that one finished before the beta.

It would still be good to get this change into the beta since it would make application code almost 100% compatible with the way the application code would look after the redesign.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7963)*

Legioth:
Discussed with Teemu. We'll do a slightly bigger redesign, but this change is still a step in that direction and it will reduce the impact of the redesign once it gets through.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/7963#-:-KYwjeMa3JfJ-Nk6osm-:bjztazp)*

PR #8005, opened 2016-12-15
Extract interfaces from Grid

tsuoanttila:
Closes vaadin/framework8-issues#566

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8005)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -33,7 +33,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
-import java.util.function.BiFunction;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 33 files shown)


pleku:
Reviewed 33 of 33 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8005#-KZ1bMsHrUMBFlDhHNmn:-KZ1bMsHrUMBFlDhHNmo:b9f0van) (no related file):*
Only thing I'm concerned is the `SelectionModel`, `GridSelectionModel` and the sub interfaces of the latter, `SingleSelectionModel`, `MultiSelectionModel` and `NoSelectionModel`.

If/when we add components that have their own selection models that extend `SelectionModel`, I guess those will also prefixed with the Component name, and then the possible subinterfaces might be too, or they will conflict with Grid's Single-, No- and MultiSelectionModel.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8005)*

tsuoanttila:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8005#-KZ1bMsHrUMBFlDhHNmn:-KZ1kAe8sEMwUlslW4-j:b-35d990) (no related file):*

Only thing I'm concerned is the `SelectionModel`, `GridSelectionModel` and the sub interfaces of the latter, `SingleSelectionModel`, `MultiSelectionModel` and `NoSelectionModel`.

If/when we add components that have their own selection models that extend `SelectionModel`, I guess those will also prefixed with the Component name, and then the possible subinterfaces might be too, or they will conflict with Grid's Single-, No- and MultiSelectionModel.
</blockquote></details>

Or we need to take a look if we can somehow make it a common interface for all components at that stage. Since the user usually should not need to use GridSelectionModel anyway (The user should be interested in either Single or MultiSelectionModel) it should be safe to refactor out at a later stage if we unify these APIs somehow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8005)*

PR #8012, opened 2016-12-16
Automatically complete custom bindings in BeanBinder#bindInstanceFields

ahie:
Closes vaadin/framework8-issues#511

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8012)


denis-anisimov:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZ5VFliPmN4EJmDMUHV:b-qxmrz8) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*
> ```Java
>     private final Class<? extends BEAN> beanType;
>     private final Set<String> boundProperties;
>     private final Map<HasValue<?>, BeanBindingImpl<BEAN, ?, ?>> tentativeBindings;
> ```

I would like to avoid this kind of approach whenever it is possible (I'm not sure that it's possible here).
So I would like to ask reconsider this approach completely and try to find another solution (if it's possible).

The issues are:

- it's possible to have something in `tentativeBindings` which will stick there for indefinite time. Normally binder shouldn't be stored for a long time but if it happens then unused bindings might be there as long as binder exists. This is potential memory leak.
- Map which has a `HasValue` as a key. It means you are relying on some specific implementation of `hashCode` (and `equals` but it's usually not a problem). In reality it means that it either should be default or this instance should be immutable. Please never rely on that. 

Map can be fixed to use `IdentityHashMap` instead of `HashMap`.
Potential memory leak can be fixed via `WeakHashMap` but it has two issues: it's not serializable and issue solved with `IdentityHashMap` comes back.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 64 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5UWhJCw27-0UIgcpT:-KZ5UWhJCw27-0UIgcpU:bttjmbl) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L64)):*
> ```Java
>         BeanBinder<TestBean> otherBinder = new BeanBinder<>(TestBean.class);
>         TestClass testClass = new TestClass();
>         otherBinder.forField(testClass.number).withConverter(
> ```

Out of the scope of this change. But since you have modified it:
what does this test check ?

I don't see here any assertion. I guess it shouldn't throw any exception.
But it's not obvious.

Please add either a comment to the method or just name the method so it's obvious from the name.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 71 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5UwhUKED9DP0cpEke:-KZ5UwhUKED9DP0cpEkf:b-m569sa) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L71)):*
> ```Java
> 
>     @Test
>     public void bindInstanceFields_detects_incomplete_custom_bindings() {
> ```

Detects ?

I don't think I understand what it means. 
Should it just automatically bind ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Reviewed 1 of 3 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

Legioth:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 64 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5_zC4u1ZUVCrhIWIN:-KZ5_zC4u1ZUVCrhIWIO:b-a6t2ja) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L64)):*
> ```Java
>         BeanBinder<TestBean> otherBinder = new BeanBinder<>(TestBean.class);
>         TestClass testClass = new TestClass();
>         otherBinder.forField(testClass.number).withConverter(
> ```

Is there any reason for not following the API defined in the ticket, i.e. explicitly using a different method name (`forMemberField`) for bindings that are intended to be completed through `bindInstanceFields`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

ahie:
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZ5b5WL_dHx9RRy6Lla:bg6o957) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*

I would like to avoid this kind of approach whenever it is possible (I'm not sure that it's possible here).
So I would like to ask reconsider this approach completely and try to find another solution (if it's possible).

The issues are:

- it's possible to have something in `tentativeBindings` which will stick there for indefinite time. Normally binder shouldn't be stored for a long time but if it happens then unused bindings might be there as long as binder exists. This is potential memory leak.
- Map which has a `HasValue` as a key. It means you are relying on some specific implementation of `hashCode` (and `equals` but it's usually not a problem). In reality it means that it either should be default or this instance should be immutable. Please never rely on that. 

Map can be fixed to use `IdentityHashMap` instead of `HashMap`.
Potential memory leak can be fixed via `WeakHashMap` but it has two issues: it's not serializable and issue solved with `IdentityHashMap` comes back.
</blockquote></details>

Changed to use IdentityHashMap.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 64 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5UWhJCw27-0UIgcpT:-KZ5b9msqDmlBcwrQ6Gh:b-f8sf1p) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L64)):*

Out of the scope of this change. But since you have modified it:
what does this test check ?

I don't see here any assertion. I guess it shouldn't throw any exception.
But it's not obvious.

Please add either a comment to the method or just name the method so it's obvious from the name.
</blockquote></details>

Added a comment.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 64 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5_zC4u1ZUVCrhIWIN:-KZ5bFXVc3hVdXJ5rRDJ:buu0v46) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L64)):*

Is there any reason for not following the API defined in the ticket, i.e. explicitly using a different method name (`forMemberField`) for bindings that are intended to be completed through `bindInstanceFields`?
</blockquote></details>

I don't see any reason for having a separate API. If you don't want the binding to be automatically completed it can still be bound manually.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 71 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5UwhUKED9DP0cpEke:-KZ5bF-jUy2FRHyerHzi:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L71)):*

Detects ?

I don't think I understand what it means. 
Should it just automatically bind ?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

Legioth:
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 64 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5_zC4u1ZUVCrhIWIN:-KZ5chMvWL2t1Um2L-DL:bmotm9c) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L64)):*

I don't see any reason for having a separate API. If you don't want the binding to be automatically completed it can still be bound manually.
</blockquote></details>

What's the reason for using the same method name for different intentions?

My reason is that I'd like to eventually implement a feature that helps users realize what's wrong if they do `forField` but forget to do a corresponding `bind`. The way of fixing this would be to make most methods in `Binder` throw e.g. `IllegalStateException` if there's currently an unfinished `BindingBuilder`. This would not work if `forField` is also used in cases where the binding is not supposed to be finished.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Reviewed 1 of 2 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5iHLMKSwTTCAowTNq:-KZ5iHLMKSwTTCAowTNr:bc0mvez) (no related file):*
There are still issues.

Let's postpone this after the beta.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Reviewed 1 of 2 files at r2.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZ_QNEVNFwiN3TWgvW4:b-t9mk1e) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*

Changed to use IdentityHashMap.
</blockquote></details>

Now there is still an issue with "unfinished" bindings.
If `bindInstanceFields` is not called at all (developer configures a `Binder` completely manually) then it's possible to have some bindings in `tentativeBindings` "stored forever". Meaning: until the `Binder` instance is not GCed.
As I said: that would be solved WeakHashMap but then other issues appear.

What should be done at least:  developer should aware of "unfinished" bindings. 
So that `setBind()` method should throw an exception if there are unbound bindings (`tentativeBindings` is not empty).
The same should happen for `readBean()` (most likely).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

ahie:

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZ_VsRw4tYXS98AAsIf:bqp02pa) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*

Now there is still an issue with "unfinished" bindings.
If `bindInstanceFields` is not called at all (developer configures a `Binder` completely manually) then it's possible to have some bindings in `tentativeBindings` "stored forever". Meaning: until the `Binder` instance is not GCed.
As I said: that would be solved WeakHashMap but then other issues appear.

What should be done at least:  developer should aware of "unfinished" bindings. 
So that `setBind()` method should throw an exception if there are unbound bindings (`tentativeBindings` is not empty).
The same should happen for `readBean()` (most likely).
</blockquote></details>

Yes, the GC issue still exists. The only way to solve both issues would be to create a WeakIdentityHashMap. This should be done since `initialConverters` in `Binder` also has this same problem.

For the other point, we discussed moving `tentativeBindings` to `Binder` with Leif, so exactly that and other checks could be done. `forMemberField` would then be implemented on top of that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZ_dJ-7P8az0Qh7w3ux:b-80tlqk) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*

Yes, the GC issue still exists. The only way to solve both issues would be to create a WeakIdentityHashMap. This should be done since `initialConverters` in `Binder` also has this same problem.

For the other point, we discussed moving `tentativeBindings` to `Binder` with Leif, so exactly that and other checks could be done. `forMemberField` would then be implemented on top of that.
</blockquote></details>

Sure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

hesara:

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZa63ZYznIlY5UI7hKF:b-63rpln) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*

Sure.
</blockquote></details>

Weak maps cause problems when clustering Vaadin applications. Thus, I'm not sure if we can go that way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 331 at r1](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZ5VFliPmN4EJmDMUHU:-KZa6NSr0SFakl0IdR0u:b-a9s0kh) ([raw file](https://github.com/vaadin/framework/blob/6bf330ec0e829be54cb7079a8f23cbd64550e0ac/server/src/main/java/com/vaadin/data/BeanBinder.java#L331)):*

Weak maps cause problems when clustering Vaadin applications. Thus, I'm not sure if we can go that way.
</blockquote></details>

Yes, that's what we have discussed: no WeakHashMap .

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 263 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfI1qKzV5qDLlvEhr8:-KZfI1qLiDDh3IpY6v3Y:b-w7pyy4) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L263)):*
> ```Java
> 
>         @Override
>         public Binding<BEAN, TARGET> bind(ValueProvider<BEAN, TARGET> getter,
> ```

Why is it overridden ?

Looks like it doesn't do anything except calling a super method.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 329 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZauWd4ztAWoGnl2-wz:-KZauWd5AS3tgCQsT5zq:b-3zts5i) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L329)):*
> ```Java
>     private final Class<? extends BEAN> beanType;
>     private final Set<String> boundProperties = new HashSet<>();
>     private final Map<HasValue<?>, BeanBindingImpl<BEAN, ?, ?>> incompleteMemberFieldBindings = new IdentityHashMap<>();
> ```

Minor: 
I would move initialization into CTOR instead of inline at the definition.
`beanType` is initialized in the CTOR. So for consistency it's good to have all initializaiton in one place.

But doesn't really matter.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 360 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZauABdKkZPoXVDRj3T:-KZauABdKkZPoXVDRj3U:b-1eon37) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L360)):*
> ```Java
>      * @return the new binding builder
>      */
>     public <FIELDVALUE> BeanBindingBuilder<BEAN, FIELDVALUE> forMemberField(
> ```

Javadoc should say something about accompanied `forField` method.
`@see` should be added.

The same should go to the `forField`.
Some description which method to use is required.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 414 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfIRJK_12pzXF9e7le:-KZfIRJLBLl-Cw9cmkrH:b-igtfzc) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L414)):*
> ```Java
>             incompleteMemberFieldBindings.put(field, newBinding);
>         } else {
>             getIncompleteBindings().put(field, newBinding);
> ```

I would like to avoid having the exposed `getIncompleteBindings()` method at all.
There shouldn't be such method at all or it should be `private`.

The only reason why it's exposed is this one place.
But the method `createBinding` overrides the super method which actually does what you need out of the box.

The only problem is that the binding impl type is different.
But this can be solved via polymorphism : introduce something like `doCreateBinding` method which creates a binding by parameters.
In the super class it will create a `BindingBuilder` , here it will create a `BeanBindingImpl`.
(original implementation of the `createBinding`method).

And then use this method in the original implementation of the `createBinding`.
Here you will need to call only `super.createBinding` in the "else" block.

And there will be no need to expose the `getIncompleteBindings()` at all. This is internal detail of the implementation and shouldn't be exposed.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 558 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfLhTDIrqyfBz1v0vx:-KZfLhTEgvaph0gpRym3:bdld5jz) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/Binder.java#L558)):*
> ```Java
> 
>             bound = true;
>             getBinder().incompleteBindings.remove(getField());
> ```

If you have a `getIncompleteBindings` method then please use it everywhere instead of direct field access.
Especially if it's `protected`.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1604 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfM7pgeyuxBVDQpV2I:-KZfM7pgeyuxBVDQpV2J:b-ikg25o) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/Binder.java#L1604)):*
> ```Java
>      * @return a map from fields to their incomplete binding builders
>      */
>     protected Map<HasValue<?>, BindingBuilder<BEAN, ?>> getIncompleteBindings() {
> ```

See comments in the child class about it.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1732 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfMAemGs1CJyidpiR8:-KZfMAenpe27avzOiz9C:bnu0xl7) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/Binder.java#L1732)):*
> ```Java
>     }
> 
>     protected void checkBindingsCompleted(String message) {
> ```

This comment actually is not in the correct place. But I just can't an appropriate place for it in the `BeanBinder`.

I see an asymmetry in "fields bindings" and "regular bindings" .
There are two cases which should be considered "equally" :
- developer uses only manual bindings for everything
- developer uses "bindInstanceFields` (and may be custom bindings for some fields).

This method makes a check for the second case only. (And here it's OK since there are no any checks need to be done in this class).

But the first case should be handled in the `BeanBinder` in case of incorrect method usage. Developer is able to use `forMemberField` accidentally instead of `forField`.  And there is no any check that `incompleteMemberFieldBindings` is empty in the `BeanBinder` in similar situations where `incompleteBindings` should be empty.

So most likely this method should be overridden in the subcalss with `super` call.

And there should be a test for that.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 90 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfOHqrpMu5MDd53IZJ:-KZfOHqrpMu5MDd53IZK:bfcwbm7) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L90)):*
> ```Java
> 
>     @Test(expected = IllegalStateException.class)
>     public void bindInstanceFields_does_not_automatically_bind_incomplete_forField_bindings() {
> ```

Please see above about symmetric case for `BeanBinder`.

There should be a check that `incompleteMemberFieldBindings` is empty.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

ahie:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 263 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfI1qKzV5qDLlvEhr8:-K_TTx8bhjTELEY8IihK:bss4a2b) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L263)):*

Why is it overridden ?

Looks like it doesn't do anything except calling a super method.
</blockquote></details>

Removed.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 360 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZauABdKkZPoXVDRj3T:-K_TTp30wWc1JX7ShHS4:b-jaop8v) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L360)):*

Javadoc should say something about accompanied `forField` method.
`@see` should be added.

The same should go to the `forField`.
Some description which method to use is required.
</blockquote></details>

Clarified `forMemberField`.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 414 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfIRJK_12pzXF9e7le:-K_TM6YLK3mH_hEfhJbZ:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/BeanBinder.java#L414)):*

I would like to avoid having the exposed `getIncompleteBindings()` method at all.
There shouldn't be such method at all or it should be `private`.

The only reason why it's exposed is this one place.
But the method `createBinding` overrides the super method which actually does what you need out of the box.

The only problem is that the binding impl type is different.
But this can be solved via polymorphism : introduce something like `doCreateBinding` method which creates a binding by parameters.
In the super class it will create a `BindingBuilder` , here it will create a `BeanBindingImpl`.
(original implementation of the `createBinding`method).

And then use this method in the original implementation of the `createBinding`.
Here you will need to call only `super.createBinding` in the "else" block.

And there will be no need to expose the `getIncompleteBindings()` at all. This is internal detail of the implementation and shouldn't be exposed.
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 558 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfLhTDIrqyfBz1v0vx:-K_TM7gmTQbjkBimuweN:b-ekn649) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/Binder.java#L558)):*

If you have a `getIncompleteBindings` method then please use it everywhere instead of direct field access.
Especially if it's `protected`.
</blockquote></details>

`getIncompleteBindings` is now removed.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1732 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfMAemGs1CJyidpiR8:-K_TMFOya4zBVTdCLNiD:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/main/java/com/vaadin/data/Binder.java#L1732)):*

This comment actually is not in the correct place. But I just can't an appropriate place for it in the `BeanBinder`.

I see an asymmetry in "fields bindings" and "regular bindings" .
There are two cases which should be considered "equally" :
- developer uses only manual bindings for everything
- developer uses "bindInstanceFields` (and may be custom bindings for some fields).

This method makes a check for the second case only. (And here it's OK since there are no any checks need to be done in this class).

But the first case should be handled in the `BeanBinder` in case of incorrect method usage. Developer is able to use `forMemberField` accidentally instead of `forField`.  And there is no any check that `incompleteMemberFieldBindings` is empty in the `BeanBinder` in similar situations where `incompleteBindings` should be empty.

So most likely this method should be overridden in the subcalss with `super` call.

And there should be a test for that.
</blockquote></details>

Done.

---

*[server/src/test/java/com/vaadin/data/BeanBinderTest.java, line 90 at r3](https://reviewable.io:443/reviews/vaadin/framework/8012#-KZfOHqrpMu5MDd53IZJ:-K_TMQ7rP_wc-oPUvn86:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/090cda91d86c6182ea76fe63a4fb5d666f80ab03/server/src/test/java/com/vaadin/data/BeanBinderTest.java#L90)):*

Please see above about symmetric case for `BeanBinder`.

There should be a check that `incompleteMemberFieldBindings` is empty.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Reviewed 3 of 3 files at r4.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 599 at r4](https://reviewable.io:443/reviews/vaadin/framework/8012#-K_TleDVqeryloj24kw8:-K_TleDVqeryloj24kw9:bxhhjo9) ([raw file](https://github.com/vaadin/framework/blob/79d67e3737f48db70f9f5f276fc0195d7919fdae/server/src/main/java/com/vaadin/data/BeanBinder.java#L599)):*
> ```Java
> 
>     @Override
>     protected void checkBindingsCompleted(String message) {
> ```

I think the contract for this method could be better:

the error message is created by the code which performs the check.
The only reason for this is the correct method name to report.

But as a developer I would like to know which kind of bindings are incomplete (member bindings or "custom" bindings).
So I would suggest: 
- provide a method name causing the check as an argument instead of complete error message (rename the parameter)
- construct the message right here with the information about bindings that are incomplete
- use the parameter value in the message to report the method name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

ahie:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 599 at r4](https://reviewable.io:443/reviews/vaadin/framework/8012#-K_TleDVqeryloj24kw8:-K_TyisoWeEpVPrARlH0:b-wcs1i9) ([raw file](https://github.com/vaadin/framework/blob/79d67e3737f48db70f9f5f276fc0195d7919fdae/server/src/main/java/com/vaadin/data/BeanBinder.java#L599)):*

I think the contract for this method could be better:

the error message is created by the code which performs the check.
The only reason for this is the correct method name to report.

But as a developer I would like to know which kind of bindings are incomplete (member bindings or "custom" bindings).
So I would suggest: 
- provide a method name causing the check as an argument instead of complete error message (rename the parameter)
- construct the message right here with the information about bindings that are incomplete
- use the parameter value in the message to report the method name.
</blockquote></details>

The messages now contain info whether the bindings have been created with forField or forMemberField.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

denis-anisimov:
Should be unblocked by @hesara and @legioth 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012#-:-K_TzROamjnhYuQxuCIp:b-q4qb9a)*

Legioth:

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 602 at r5](https://reviewable.io:443/reviews/vaadin/framework/8012#-K_U54m1APfjACAIzcSh:-K_U54m1APfjACAIzcSi:b-awrsyf) ([raw file](https://github.com/vaadin/framework/blob/73b09283541b5d2e72e5608b44e10349edea7e02/server/src/main/java/com/vaadin/data/BeanBinder.java#L602)):*
> ```Java
>         if (!incompleteMemberFieldBindings.isEmpty()) {
>             throw new IllegalStateException(
>                     "All bindings created with forMemberField must be completed before calling "
> ```

I think this should explicitly say that `bindInstanceFields` should be called so that a user that gets this exception can know what they're expected to do to fix this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

ahie:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/BeanBinder.java, line 602 at r5](https://reviewable.io:443/reviews/vaadin/framework/8012#-K_U54m1APfjACAIzcSh:-K_UGTJYeExFhFMYNR8e:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/73b09283541b5d2e72e5608b44e10349edea7e02/server/src/main/java/com/vaadin/data/BeanBinder.java#L602)):*

I think this should explicitly say that `bindInstanceFields` should be called so that a user that gets this exception can know what they're expected to do to fix this.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

Legioth:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8012)*

PR #8035, opened 2016-12-19
Fix the upload test for IE11 (at least).

denis-anisimov:
The upload test should use an Upload in non-immediate mode because the
file name is written in-line into the Upload text input and the button
is used to upload a file without showing a file chooser dialog. The test
for upload in immediate mode works for Chrome (no file chooser window
issue) but it doesn't work for IE11 (at least, may be the same issue for
FF).
So the test UI is just updated to use non-immediate mode.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8035)


hesara:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8035#-:-KZLw8gtgPlK23PTmPYf:bnfp4nl)*

PR #8054, opened 2016-12-19
Improved English in line 17

macjuan:
Changed 

"There is only one Binder instance per form and it is used for all fields in that form" into
"There is only one Binder instance for each form and it is used for all fields in that form"

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8054)


Legioth:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8054)*

PR #8072, opened 2016-12-20
Use correct indexes in multiselect checkboxes after removing rows

### LINKED ISSUE #8011: Deleting an item in a grid has a sync issue on Chrome even after reset in 7.7.5
I manually delete the selected rows in the grid using:

```
for(Object selectedItem: grid.getSelectedRows())
    grid.getContainerDataSource().removeItem(selectedItem);

grid.getSelectionModel().reset();
```

According to the Vaadin documentation you need to call reset() on the selectionModel so that they are in sync, and that's great.

This all works great but only if I scroll down the grid with the mouse in the browser, so that the whole grid moves. If I don't scroll the grid in my browser then the mouse selects the wrong row by the exact number of rows that were previously selected before that row. So if I select row 2 and then delete it, then select row 3 it will in fact select row 4. If however I scroll down so that the grid view is refreshed then everything works exactly as expected and the correct row is selected.

Artur-:
Fixes #8011

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8072)

### DIFF
--- a/client/src/main/java/com/vaadin/client/widget/grid/selection/MultiSelectionRenderer.java
+++ b/client/src/main/java/com/vaadin/client/widget/grid/selection/MultiSelectionRenderer.java
@@ -26,6 +26,7 @@ import com.google.gwt.dom.client.BrowserEvents;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.dom.client.NativeEvent;
 import com.google.gwt.dom.client.TableElement;
+import com.google.gwt.dom.client.TableRowElement;
 import com.google.gwt.dom.client.TableSectionElement;
 import com.google.gwt.event.dom.client.ClickEvent;
 import com.google.gwt.event.dom.client.ClickHandler;
@@ -45,6 +46,7 @@ import com.vaadin.client.widget.grid.RendererCellReference;
 import com.vaadin.client.widget.grid.events.GridEnabledEvent;
 import com.vaadin.client.widget.grid.events.GridEnabledHandler;
 import com.vaadin.client.widget.grid.selection.SelectionModel.Multi.Batched;
+import com.vaadin.client.widgets.Escalator.AbstractRowContainer;
 import com.vaadin.client.widgets.Grid;
 
 /**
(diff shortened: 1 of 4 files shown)


denis-anisimov:
Reviewed 4 of 4 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8072#-KZ_MLUiygbrW55r1f3k:-KZ_MLUiygbrW55r1f3l:b-sehspz) (no related file):*
It's possible now to get `Escalator` instance from a `Grid`.
May be it's even good.

But may be there can be a reason do not expose it ( I don't know. This patch doesn't show me details).

What I see from the patch : it would be enough to expose an `AsbstractRawContainer` only.
May be it has sense to make `getEscalator()` as a protected method back and introduce one more method to return an `AsbstractRawContainer` from the `Grid`.

May be not. Up to you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8072)*

Artur-:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8072#-KZ_MLUiygbrW55r1f3k:-KZ_NVdMBLwfvhUDf9ZC:bdnkovy) (no related file):*

It's possible now to get `Escalator` instance from a `Grid`.
May be it's even good.

But may be there can be a reason do not expose it ( I don't know. This patch doesn't show me details).

What I see from the patch : it would be enough to expose an `AsbstractRawContainer` only.
May be it has sense to make `getEscalator()` as a protected method back and introduce one more method to return an `AsbstractRawContainer` from the `Grid`.

May be not. Up to you.
</blockquote></details>

I have never understood why Escalator is hidden from renderers, selection models etc as the API is needed every now and then... Maybe somebody else can tell if there is a reason

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8072)*

denis-anisimov:
:lgtm:

---

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8072#-KZ_MLUiygbrW55r1f3k:-KZ_NwwxkV_z-sNb4Cw5:b-ojldzu) (no related file):*

I have never understood why Escalator is hidden from renderers, selection models etc as the API is needed every now and then... Maybe somebody else can tell if there is a reason
</blockquote></details>

May be just because by default it's good to keep visibility as less as possible (without a good reason) and there was not enough thinking about its usage in renderers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8072#-:-KZ_OKvnNgmfx87DD3TV:bnfp4nl)*

PR #8079, opened 2016-12-22
Migrate 7.7.6 fixes to the master

denis-anisimov:
Fixes vaadin/framework8-issues#553

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8079)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/dd/DragHandle.java
+++ b/client/src/main/java/com/vaadin/client/ui/dd/DragHandle.java
@@ -87,6 +87,7 @@ public class DragHandle {
      *            later) both to provide the "-dragged" style and to make sure
      *            that the drag handle can be properly styled (it's otherwise
      *            invisible)
+     * @since 7.7.5
      */
     public DragHandle(String baseName) {
         this(baseName, null);
@@ -185,6 +186,7 @@ public class DragHandle {
      *
      * @param dragHandleCallback
      *            the callback object to use (can be null)
+     * @since 7.7.5
      */
     public void setCallback(DragHandleCallback dragHandleCallback) {
         userCallback = dragHandleCallback;
--- a/client/src/main/java/com/vaadin/client/ui/textfield/AbstractTextFieldConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/textfield/AbstractTextFieldConnector.java
@@ -135,6 +135,11 @@ public abstract class AbstractTextFieldConnector extends AbstractFieldConnector
         sendValueChange();
     }
 
+    /**
+     * {@inheritDoc}
+     * 
+     * @since 8.0.0
+     */
     @Override
     public boolean isWorkPending() {
         return getValueChangeHandler().isScheduled();
--- a/client/src/main/java/com/vaadin/client/ui/textfield/AbstractTextFieldConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/textfield/AbstractTextFieldConnector.java
@@ -16,6 +16,7 @@
 package com.vaadin.client.ui.textfield;
 
 import com.google.gwt.user.client.ui.Widget;
+import com.vaadin.client.DeferredWorker;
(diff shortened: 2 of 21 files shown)


elmot:
:lgtm:

---

Reviewed 11 of 15 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8079#-:-KZav_X59Rlh1Apl4fAJ:bnfp4nl)*

hesara:
Review status: 20 of 21 files reviewed at latest revision, 1 unresolved discussion.

---

*[uitest/src/test/java/com/vaadin/tests/components/textfield/TextFieldsValueChangeModeTest.java, line 99 at r5](https://reviewable.io:443/reviews/vaadin/framework/8079#-KZedsk_2pEAIJHFdzNT:-KZedsk_2pEAIJHFdzNU:b-h8l6da) ([raw file](https://github.com/vaadin/framework/blob/8eb5a369859c51a385249dfdafcaf0398c429dca/uitest/src/test/java/com/vaadin/tests/components/textfield/TextFieldsValueChangeModeTest.java#L99)):*
> ```Java
>         openTestURL();
>         WebElement eagerTextField = findElement(By.id(id));
>         eagerTextField.sendKeys("foo");
> ```

I guess the idea was that this was testing that there is no event between the f, o and o, and b, a and a but only when the timer expires. Why was this changed? Was the test unstable? Or did the implementation change so that the old assumption is no longer valid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8079)*

denis-anisimov:
Review status: 20 of 21 files reviewed at latest revision, 1 unresolved discussion.

---

*[uitest/src/test/java/com/vaadin/tests/components/textfield/TextFieldsValueChangeModeTest.java, line 99 at r5](https://reviewable.io:443/reviews/vaadin/framework/8079#-KZedsk_2pEAIJHFdzNT:-KZeg6Su7_AvR92pqpCt:b-qx4j7y) ([raw file](https://github.com/vaadin/framework/blob/8eb5a369859c51a385249dfdafcaf0398c429dca/uitest/src/test/java/com/vaadin/tests/components/textfield/TextFieldsValueChangeModeTest.java#L99)):*

I guess the idea was that this was testing that there is no event between the f, o and o, and b, a and a but only when the timer expires. Why was this changed? Was the test unstable? Or did the implementation change so that the old assumption is no longer valid?
</blockquote></details>

May be you are right.

But this cannot be tested in this way with the migrated patch.
What the patch does: it provides an automatic Testbench "waiting for vaadin".
So when there is a timeout set for textfield/textarea then the connector will return `true` as a result of `isWorkPending()`. So Testbench will wait until the value is sent back to the server.
So it will wait until timeout expired.
It means that there WILL BE an event about entered character (e.g. 'f').
And this test has no any sense. Since it just fails (that what's happen). 

So there should be only one command to do some action atomically before "wait for vaadin" start to do its work.

In the previous version of test there were no any activities on the Vaadin side because of absent patch. So "wait for vaadin" just did nothing ( almost) : no any waiting and subsequent `sendKeys` methods executed fast one after another .
This is not true anymore. (next `sendKeys` method will wait for the previous one and this will be exactly a timeout in 1 second)

In reality what I've got from talk with Artur is : `sendKeys("abc")` should do the same thing as subsequent `sendKeys("a")`, `sendKeys("b")`, `sendKeys("c")` calls.
And if it's true then the test is exactly the same since there is no check in between the calls of `sendKeys()` here in the test (that there is no unexpected log message). 

I hope my description is not too complex.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8079)*

pleku:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8079)*

PR #8119, opened 2016-12-30
Fix DataCommunicator - clear KeyMapper when DataProvider is replaced

### LINKED ISSUE #8100: Investigate broken WH
Something weird going on in WH (v7) Fill hours tab. When I update any row, the row selection doesn’t work anymore after that util reload.

WHv7 is apparently using 8.0.0.alpha10 nowadays

elmot:
Fix DataCommunicator - clear KeyMapper when DataProvider is replaced

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8119)

### DIFF
--- a/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
+++ b/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
@@ -460,6 +460,7 @@ public class DataCommunicator<T, F> extends AbstractExtension {
         Objects.requireNonNull(dataProvider, "data provider cannot be null");
         this.dataProvider = dataProvider;
         detachDataProviderListener();
+        getKeyMapper().removeAll();
         /*
          * This introduces behavior which influence on the client-server
          * communication: now the very first response to the client will always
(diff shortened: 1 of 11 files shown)


denis-anisimov:
Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[server/src/main/java/com/vaadin/ui/AbstractListing.java, line 384 at r1](https://reviewable.io:443/reviews/vaadin/framework/8119#-K_F4JO7hU_awGIsjOQR:-K_F4JO7hU_awGIsjOQS:b4yv6lv) ([raw file](https://github.com/vaadin/framework/blob/a25b690f1d188753fa0f202d862fe9f117e02b7e/server/src/main/java/com/vaadin/ui/AbstractListing.java#L384)):*
> ```Java
> 
>         List<T> readItems = readItems(design, context);
>         if (!readItems.isEmpty() && this instanceof Listing) {
> ```

I would rather keep it as is and override `doReeadDesign` method in sublcasses so that it does "select" action once items are read (calling super of course ).

`readItems` actually shouldn't care about "selection" (it's not obvious at least from the name and javadoc). So it would be better to extract this functionality from this method.

OK, it would be good but it will require redesign because selection should be passed somehow outside of `readItems` method.
That's the reason why it hasn't been done in that way from the very beginning.

Nevermind.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8119)*

elmot:
Is going to be fixed in another pull+request

PR #8139, opened 2017-01-04
Added expand helper method to ordered layouts

mstahv:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8139)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/HorizontalLayout.java
+++ b/server/src/main/java/com/vaadin/ui/HorizontalLayout.java
@@ -60,4 +60,32 @@ public class HorizontalLayout extends AbstractOrderedLayout {
         return (HorizontalLayoutState) super.getState(markAsDirty);
     }
 
+    /**
+     * Expands given the components to consume all available space. The method
+     * basically uses {@link #setExpandRatio(Component, float)} method and
+     * configures both given components and this component so that all available
+     * space is distributed to given components. The method also adds given
+     * components to layout if they are not added yet.
+     *
+     * @param componentsToExpand
+     *            the component(s) which should be expanded
+     */
+    public void expand(Component... componentsToExpand) {
+        configureParentForExapntion();
+        for (Component component : componentsToExpand) {
+            if (component.getParent() != this) {
+                addComponent(component);
+            }
+            setExpandRatio(component, 1);
+            component.setWidth(100, Unit.PERCENTAGE);
+        }
+    }
+
+    private void configureParentForExapntion() {
+        if (getWidth() < 0) {
+            // Make full width if no other size is set
+            setWidth(100, Unit.PERCENTAGE);
+        }
+    }
+
 }
--- a/server/src/main/java/com/vaadin/ui/HorizontalLayout.java
+++ b/server/src/main/java/com/vaadin/ui/HorizontalLayout.java
@@ -71,7 +71,7 @@ public class HorizontalLayout extends AbstractOrderedLayout {
      *            the component(s) which should be expanded
      */
     public void expand(Component... componentsToExpand) {
-        configureParentForExapntion();
+        configureParentForExpansion();
         for (Component component : componentsToExpand) {
             if (component.getParent() != this) {
                 addComponent(component);
@@ -81,7 +81,7 @@ public class HorizontalLayout extends AbstractOrderedLayout {
         }
     }
 
-    private void configureParentForExapntion() {
+    private void configureParentForExpansion() {
         if (getWidth() < 0) {
             // Make full width if no other size is set
             setWidth(100, Unit.PERCENTAGE);
--- a/server/src/main/java/com/vaadin/ui/VerticalLayout.java
+++ b/server/src/main/java/com/vaadin/ui/VerticalLayout.java
@@ -62,4 +62,33 @@ public class VerticalLayout extends AbstractOrderedLayout {
     protected VerticalLayoutState getState(boolean markAsDirty) {
         return (VerticalLayoutState) super.getState(markAsDirty);
     }
+
+    /**
+     * Expands given the components to consume all available space. The method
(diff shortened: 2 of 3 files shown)


mstahv:
This is how this change simplifies code

## From:

        VerticalLayout root = new VerticalLayout(toolbar,);
        root.addComponent(toolbar);
        root.addComponent(content);
        root.setSizeFull();
        content.setSizeFull();
        root.setExpandRatio(content, 1);

## To:

        VerticalLayout root = new VerticalLayout();
        root.addComponent(toolbar);
        root.expand(content);

... in many places in your Vaadin code. Also the new API makes it much easier for new users to use the expansion features.

I'd still consider also to configure all well known parent component if the parent layout already happens to be added some other component, but this is how it has worked in Viritin pretty well already.

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8139#-K_dbr6LgaA5Un_qriUK:-K_dbr6MTi0y-aJL3tmT:bit90vb) (no related file):*
I think it's quite hard to understand what the `expand` method does since it both adds a component and configures it, especially if you ever mix those two modes of operation in one method call:
```
VerticalLayout layout = new VerticalLayout(c1);
layout.expand(c2, c1);
```

I'd suggest splitting the functionality into two separate methods to make things clearer:

1. `expand` assumes the component is already added and throws an `IllegalArgumentException` if it's not (in the same way as `setExpandRatio` already does)
1. `addExpanded` as a shorthand for `addComponent`+ `expand`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8139)*

mstahv:
In which case it doesn't do what you wanted to do?

Legioth:
It does too much, which makes it confusing. When reading code and seeing `layout.expand(c1, c2)`, you cannot know whether `c1` will end up before or after `c2` since that depends on whether those components have already been added to the layout or not.

That's why I suggest splitting it up to two methods: If you see `layout.expand(c1, c2)`, then you would know that it doesn't affect the ordering at all since both components must already be children of `layout`. If you instead see `layout.addExpanded(c1, c2)`, you would know that `c1` will be before `c2` since they are both (re)added to the end of the layout.

What's your objection against splitting up the functionality into two separate methods that can be understood without knowledge of the previous state of the layout?

Legioth:
Ping @mstahv. Do you have any objections against my suggestion of splitting the logic into two separate methods? I could implement the changes if you don't have time.

mstahv:
I don't really agree with your arguments that the suggested method does too much or is too complex. If you have already added the components they remain in the parent in the places where they were added. I don't think nobody expects anything else. Also one should not that most often this is used with just one parameter.

In Viritin the current method has already been proven to work well. With one method there would be less API and documentation to maintain. Fight for simplicity.

The only issue I see is Viritin 2.0. In Viritin the method also returns "this" (fluent API). Thus, I'd actually suggest we'd change the mehtod name to expandComponent. What do you think?

pleku:
The team has discussed this and decided that

```java
public void addComponentsAndExpand(Component... components) { }
```
Is the API name and the implementation shall:
- `addComponents(component`, THUS if any of the components have been added, they are moved, which is consistent to how `addComponent` / `addComponents` works.
- expand components (`setExpandRatio(1)` and `setWidth("100%")` or `setHeight("100%")`)
- set layout w/h size 100% respectively

The name is perceived as _correct_ ™ and coherent with what is currently in the framework. It has some chance of being discovered because of the `addComponent`-prefix in the method name, for power usage writing `layout.expand` should autocomplete to this method (tested Eclipse). Mileage might vary

Now who updates this PR, @mstahv or @Legioth ?

mstahv:
The method is long as hell, but on the otherhand it tells quite well what it does. The bad thing in that "specification" is that might need split adding components to three different calls in some cases:

    addComponents(foo,bar,car);
    expand(bar);

becomes:

    addComponents(foo);
    addComponentsAndExpand(bar);
    addComponents(car);

But any kind of related method here will be a great aid for new users, so let's just get something in before 8 is out (although this kind of method can be added later as well, but for marketing purposes this would be really nice addition to 8.0).

mstahv:
The pull request has tag "Waiting for author", which I guess is me. What should I do?

denis-anisimov:
I'm not sure what Pekka meant but I think you should update the PR according to the team decision (written above by Pekka).

mstahv:
I read that the "team had discussed", but that doesn't help me to get convinced or motivated about the suggested changes (although they were quite minimal to my current suggestion). That's why all this kind of "discussions" should be public, so I or somebody else could buy your reasoning. I tried continue the discussion here, but without answer (I know, Pekka is sick). 

I hope some one from the team that discussed about the topic could answer if you thought about the use case I added above. It has been a common pattern form me previously (with V7 + Viritin).

pleku:
We, _the team_ can do the updates for this patch if necessary.

Again, for the reasoning why the `addComponentsAndExpand` was considered to be the best alternative available:
- it does what it says
- the naming is __coherent__ with the other API methods in the FW
- the functionality is __coherent__ with other API methods in the FW
  - any already added component is moved, same as with doing `addComponent` 
- it discoverable

We did see the merits in the proposed `expand` method, but after weighing the different options, we concluded that this is the way it should be done in the framework. If I recall, we chose "write a bit more for being more explicit for readers" over to "write less once and make it harder for readers". And we don't want every other similar method in the same class to have different semantics (like moving components or not).

I do agree that we should make the discussion more public in the future, we need to write down the _minutes_ or such also to the ticket. Sorry for not doing that for this.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8139#-:-Kbj8hekGQIQX20eJxvp:b9q2x0w)*

Legioth:
Implementation according to the "approved" design in #8480.

Legioth:
Closing this since #8480 has been merged.

PR #8153, opened 2017-01-05
Remove tracking of unregistered connectors

### LINKED ISSUE #7748: Memory leak in ConnectorTracker when push is enabled and no transmission from the client
**Originally by _arek_da_**

___
In such scenario the syncIdToUnregisteredConnectorIds TreeMap gets overloaded causing eventually OutOfMemory error.
Please note that ConnectorTracker.cleanConcurrentlyRemovedConnectorIds is called upon transmission from the client.
In my case the application does huge amount of pushes while some clients use it as a display board (without any client interaction for a long time)

A fragment of heap trace:
759 715 240 (96,4%) [768] 68 com/fxe/gui/MainUI 0x78091e4d0
|- 758 225 000 (96,21%) [88] 7 com/vaadin/ui/ConnectorTracker 0x780af6898
   |- 756 101 384 (95,94%) [80] 1 java/util/TreeMap 0x780b06690
      |- 756 101 304 (95,94%) [56] 4 java/util/TreeMap$Entry 0x78ac358c8

___
 Imported from https://dev.vaadin.com/ issue #20048

### LINKED ISSUE #8111: Remove tracking of unregistered connectors
e2e2b2c1e8acc2163d117dd8193b7ecc445f4ff8 fixes an issue in `Window` where `Window` did not call `markAsDirty` when closing the window, if a close listener threw an exception. This lead to the client side having the `Window` still visible, while the server had removed it. Clicking again on the close button caused only a warning to be logged on the server side because an RPC (`close`) was received for an already removed connector (component).

To avoid having similar issues for other broken component connectors, c598a852c2565b67226533622b5f75a3b588acf3 changed the handling of RPCs for unknown connectors to do a full resynchronization in addition to logging a warning. This is mainly a workaround for components broken in the same way as `Window` previously was.

53282726c5769bf763beb5d8576c71e0e7b5bef3 introduced a more complex concept of tracking which connectors the client thinks exist when sending a message to the server so that the server can silently ignore RPCs for connectors which have been recently been removed on the server, so recently that the client has not received this info before sending the RPC in question. RPCs for connectors which the client already knows that have been removed still cause a full resynchronization as this indicates a problem in the client side implementation of the component. The remove connectors are stored in a map which is cleaned whenever a message from the client is received as the messages contain a synchronization id telling the server what removals the client has received info about.

5b18c5469e40529fad0d3d014ab4237a56616d81 introduced an `acknowledge` message from the client to the server for pushed updates which include removed connectors, to allow the server to clean up the map where it stores which connectors the client knows have been removed. Failing to send a client -> server request would cause the map where removed connectors are stored to keep growing in size in monitoring style applications, where there is no other client -> server communication.

Now there is still the problem that if the client is disconnected and you update the UI on the server, it will keep tracking which connectors have been removed. As the client is no longer there, there is nobody who will send an `acknowledge` message and map will continue to grow in size until the session expires or the server runs out of memory.

To solve this problem, I suggest we revert all the above patches except for the fix for `Window` and go back to logging a message (with level FINE) when an RPC is received for a connector which is not attached. 

In most cases, this means that the server removed the component and the client did not get the info about it yet. This problem will resolve itself as the client either got the info pushed to it right after sending the RPC, or will get the info in the response for the request containing the RPC.

On the client side, we should add a warning if a connector tries to send an RPC when the client side already knows that the connector has been detached and not send the RPC. This will make it easier to spot broken component implementations.

We could additionally make `setParent` call `markAsDirty` on the old parent connector, which would ensure that the parent is updated whenever a child is removed (which was the original problem in `Window`)

ahie:
Closes #8111, fixes #7748

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8153)


denis-anisimov:
Needs rebase. There is a failing test.

denis-anisimov:
Reviewed 15 of 15 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka20v9LbaTL7RY9tTLD:b5tqixf) (no related file):*
A several questions:
- I didn't get from the patch what causes the memory leak. As I understand the logic which should have removed the connectors from a map wasn't woking in some cases (when ?)
- Push test is completely removed. Shouldn't there be a replacement for it (testing a happy path  may be) ?
- Is it possible to make a test for the patch ?

For the latter question: it's possible to test that reference to an object is GCed. The question is : can we make a test assuming that we can check the absence of an object (it has been GCed)  ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka262YPiZagZpcvqFu6:b-4d8xy7) (no related file):*
>  it's possible to test that reference to an object is GCed.

See `CurrentInstanceTest.waitUntilGarbageCollected(WeakReference<?>)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

Artur-:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka27fkK-kcqLIZUqKhh:bfjdank) (no related file):*
> I didn't get from the patch what causes the memory leak. As I understand the logic which should have removed the connectors from a map wasn't woking in some cases (when ?)

A connector is added to `ConnectorTracker.syncIdToUnregisteredConnectorIds`  on component detach. It's cleaned when the browser does a request. If the browser does not do a request, it's never cleaned

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka2E8Z2P0P-Bm-kF0i8:biydl8m) (no related file):*

> I didn't get from the patch what causes the memory leak. As I understand the logic which should have removed the connectors from a map wasn't woking in some cases (when ?)

A connector is added to `ConnectorTracker.syncIdToUnregisteredConnectorIds`  on component detach. It's cleaned when the browser does a request. If the browser does not do a request, it's never cleaned
</blockquote></details>

A `ConnectorTracker` is 1:1 to `UI`, right ?
Each time when UI get a browser request the map is cleaned.
So the issue happens only when there are no anymore requests to the `UI`. But it that case the `UI` will be closed at some point. Can the clean up be done there instead (or why it doesn't work if it's already done) ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

Artur-:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka2FE95L7801yg5A5m6:b-ydkmw0) (no related file):*

A `ConnectorTracker` is 1:1 to `UI`, right ?
Each time when UI get a browser request the map is cleaned.
So the issue happens only when there are no anymore requests to the `UI`. But it that case the `UI` will be closed at some point. Can the clean up be done there instead (or why it doesn't work if it's already done) ?
</blockquote></details>

Do you have any good reason for keeping the map?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka2GcyCzl_AK48N66O9:b-5y2y7p) (no related file):*

Do you have any good reason for keeping the map?
</blockquote></details>

Not really........ But.
The code looks fragile for me at the moment: you have to remember to implement the `setParent()` method properly.
The method can be completely overridden with some custom logic. But you have to remember to do the clean up.
It would be better if this happens automatically.....
That's the only reason.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

ahie:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka2IQ-cxphW0nDSdLeb:b-xljj65) (no related file):*

Not really........ But.
The code looks fragile for me at the moment: you have to remember to implement the `setParent()` method properly.
The method can be completely overridden with some custom logic. But you have to remember to do the clean up.
It would be better if this happens automatically.....
That's the only reason.
</blockquote></details>

The push test was added to test the functionality of the acknowledge RPC (https://github.com/vaadin/framework/commit/5b18c5469e40529fad0d3d014ab4237a56616d81) which has been removed in this patch. Since the map causing the memory leak is completely removed, there are no references to test with `waitUntilGarbageCollected` or similar. The removal of the push test could be reverted, it will now pass without the acknowledge messages, but I don't see any added value in doing so.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka2nvXDO-do3KgoVU46:b-gyp4i4) (no related file):*

The push test was added to test the functionality of the acknowledge RPC (https://github.com/vaadin/framework/commit/5b18c5469e40529fad0d3d014ab4237a56616d81) which has been removed in this patch. Since the map causing the memory leak is completely removed, there are no references to test with `waitUntilGarbageCollected` or similar. The removal of the push test could be reverted, it will now pass without the acknowledge messages, but I don't see any added value in doing so.
</blockquote></details>

OK. Got it, thanks.

What about the test for the patch ( that there are no memory leaks) ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka6BLlIg-CO_3HGSVmI:bi9vcip) (no related file):*

OK. Got it, thanks.

What about the test for the patch ( that there are no memory leaks) ?
</blockquote></details>

The thing that could still be tested in the same way as `waitUntilGarbageCollected` would be that a component that has been detached is actually garbage collected.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka6BlEjLW5LUq0lghkJ:b-2cxzdi) (no related file):*

The thing that could still be tested in the same way as `waitUntilGarbageCollected` would be that a component that has been detached is actually garbage collected.
</blockquote></details>

That's what I meant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

ahie:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka6gbVK4JdQASvMXxTq:bomnrsz) (no related file):*

That's what I meant.
</blockquote></details>

This patch doesn't effect the GC of components themselves, only a string identifier is stored in the map that has been removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-Ka6jTMVPMw9Lt3wv5wK:b-o4fou0) (no related file):*

This patch doesn't effect the GC of components themselves, only a string identifier is stored in the map that has been removed.
</blockquote></details>

Right.........
It doesn't use a map anymore where the memory leak could appear.
So a memory leek is caused by the map where only integer numbers and strings are stored.....
Cool.


But.

There are changes in the patch that replaces the logic which has been implemented with the map.
This should be tested.
I'm not sure about testing it in UI (the consequence of calling markAsDirty() for an old parent) but at least there should be a unit test that checks that `markAsDirty()` is called when `setParent()`is called.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

ahie:
Review status: 15 of 17 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-KaBROSICqtQZupSEbqP:b-896fix) (no related file):*

Right.........
It doesn't use a map anymore where the memory leak could appear.
So a memory leek is caused by the map where only integer numbers and strings are stored.....
Cool.


But.

There are changes in the patch that replaces the logic which has been implemented with the map.
This should be tested.
I'm not sure about testing it in UI (the consequence of calling markAsDirty() for an old parent) but at least there should be a unit test that checks that `markAsDirty()` is called when `setParent()`is called.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[server/src/test/java/com/vaadin/tests/server/abstractextension/AbstractExtensionSetParentTest.java, line 21 at r2](https://reviewable.io:443/reviews/vaadin/framework/8153#-KaBRnWDabv7slwGv1p2:-KaBRnWEIhyeeEOAybbJ:b-3iuqlp) ([raw file](https://github.com/vaadin/framework/blob/989a8bc808dcc059caad8abf60bdcf2fce7fac4e/server/src/test/java/com/vaadin/tests/server/abstractextension/AbstractExtensionSetParentTest.java#L21)):*
> ```Java
>         extension.setParent(connector);
>         extension.setParent(null);
>         Mockito.verify(connector, Mockito.times(1)).markAsDirty();
> ```

JFYI : `Mockito.times(1)` shouldn't be needed. The `verify()` method checks this by default.
No need to correct. Just a note.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

denis-anisimov:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8153#-Ka20v9LbaTL7RY9tTLC:-KaBSA18Dlc8ACxRq5JH:bqxegw4) (no related file):*

Done.
</blockquote></details>

@artur- , please unblock.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8153)*

PR #8223, opened 2017-01-11
ensure tests will run on non-english operating systems

juergenkoras:
maven surefire creates a new JVM to execute the tests in. This JVM needs to have english language, as there are tests using bean-validation messages and check the output.
e.g. expected:<[size must be between 3 and 16]> but was:<[muss zwischen 3 und 16 liegen]>

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8223)

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -225,7 +225,7 @@
                         </manifest>
                     </archive>
                 </configuration>
-                <!-- Package src/test into a jar so that compatbility-server 
+                <!-- Package src/test into a jar so that compatbility-server
                     can use the same test classes, e.g. test beans -->
                 <executions>
                     <execution>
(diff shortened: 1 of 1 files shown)


hesara:
Ideally the locale should be set explicitly for every test that depends on it so that the test can also be executed individually without being dependent on the current locale. This change, while fixing the symptoms when testing the whole project, would hide the underlying problems, so we'll need to discuss within the core team on whether we want to merge this or not.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8223#-:-KaGT1fz8L5PCd4MubQ6:bq7d8c7)*

hesara:
Created #8226 to fix the underlying test problems

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8223#-:-KaGcdDBEYYn0zurabeb:b9zu6f6)*

hesara:
Thank you for the problem report and taking some time to find a solution. However, after some discussion, we decided not to merge this as it hides the underlying problem, and makes tests behave differently when run individually or as a part of the Maven build. Instead, I hope #8226 addresses your immediate problems.

Please let us know if you run into such issues again.

juergenkoras:
OK, thank you

PR #8241, opened 2017-01-13
Introduce Binder.setReadOnly().

### LINKED ISSUE #8232: Binder.setReadOnly
There should be a `Binder.setReadOnly(boolean)` method similarly as in v7 there is `FieldGroup.setReadOnly(boolean)`, which sets all the fields in the _form_ as read-only.

denis-anisimov:
Fixes #8232

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8241)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -1909,6 +1909,23 @@ public class Binder<BEAN> implements Serializable {
     }
 
     /**
+     * Sets the read only state to the given value for all currently bound
+     * fields.
+     * <p>
+     * This is just a shorthand for calling setReadOnly for all currently bound
+     * fields. It means that fields bound after this method call won't be set
+     * read-only.
+     *
+     * @param fieldsReadOnly
+     *            true to set the fields to read only, false to set them to read
+     *            write
+     */
+    public void setReadOnly(boolean fieldsReadOnly) {
+        getBindings().stream().map(BindingImpl::getField)
+                .forEach(field -> field.setReadOnly(fieldsReadOnly));
+    }
+
+    /**
      * Returns the event router for this binder.
      *
      * @return the event router, not null
--- a/server/src/test/java/com/vaadin/data/BinderTest.java
+++ b/server/src/test/java/com/vaadin/data/BinderTest.java
@@ -431,4 +431,67 @@ public class BinderTest extends BinderTestBase<Binder<Person>, Person> {
     public void noArgsConstructor_stringBind_throws() {
         binder.bind(new TextField(), "firstName");
     }
+
+    @Test
+    public void setReadOnly_unboundBinder() {
+        binder.forField(nameField).bind(Person::getFirstName,
+                Person::setFirstName);
+
+        binder.forField(ageField);
+
+        binder.setReadOnly(true);
+
+        Assert.assertTrue(nameField.isReadOnly());
+        Assert.assertFalse(ageField.isReadOnly());
+
+        binder.setReadOnly(false);
+
+        Assert.assertFalse(nameField.isReadOnly());
+        Assert.assertFalse(ageField.isReadOnly());
+    }
+
+    @Test
+    public void setReadOnly_boundBinder() {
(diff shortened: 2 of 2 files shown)


ahie:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1918 at r1](https://reviewable.io:443/reviews/vaadin/framework/8241#-KaNLmYrYhtw5WGNl_D6:-KaNLmYsHZS8eu7AdHKR:b-4a3u5u) ([raw file](https://github.com/vaadin/framework/blob/c8c376a57b2a515217f42299ba15d8e4d18e0757/server/src/main/java/com/vaadin/data/Binder.java#L1918)):*
> ```Java
>      *            write
>      */
>     public void setReadOnly(boolean fieldsReadOnly) {
> ```

One concern with this is that if you add a binding after calling this, the added field won't necessarily be set to the same state. I see two options: either keep the logic as is, or store this setting to the binder (also add isReadOnly?) and set readOnly from the state when bind completes a binding.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8241)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1918 at r1](https://reviewable.io:443/reviews/vaadin/framework/8241#-KaNLmYrYhtw5WGNl_D6:-KaNPOd-GmE2q8Zwi8_r:bduckw6) ([raw file](https://github.com/vaadin/framework/blob/c8c376a57b2a515217f42299ba15d8e4d18e0757/server/src/main/java/com/vaadin/data/Binder.java#L1918)):*

One concern with this is that if you add a binding after calling this, the added field won't necessarily be set to the same state. I see two options: either keep the logic as is, or store this setting to the binder (also add isReadOnly?) and set readOnly from the state when bind completes a binding.
</blockquote></details>

You are totally correct.

And I have no idea how it should be in the result.
This should have been described in the issue.
But it doesn't.

The current behavior is consistent with `FieldGroup` behavior which has the same issue.......

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8241)*

ahie:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1918 at r1](https://reviewable.io:443/reviews/vaadin/framework/8241#-KaNLmYrYhtw5WGNl_D6:-KaaOijJOvh9j2egHdOC:b-ygaez5) ([raw file](https://github.com/vaadin/framework/blob/c8c376a57b2a515217f42299ba15d8e4d18e0757/server/src/main/java/com/vaadin/data/Binder.java#L1918)):*

You are totally correct.

And I have no idea how it should be in the result.
This should have been described in the issue.
But it doesn't.

The current behavior is consistent with `FieldGroup` behavior which has the same issue.......
</blockquote></details>

Keeping it as is seems like the best option, because otherwise bind would change the state of the field as a side effect. The javadoc for this method could be improved to explicitly state that this is practically a shorthand for calling setReadOnly for all currently bound fields.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8241)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1918 at r1](https://reviewable.io:443/reviews/vaadin/framework/8241#-KaNLmYrYhtw5WGNl_D6:-KaaZuAXyvTb9ofqoQ7E:b-pbdqz3) ([raw file](https://github.com/vaadin/framework/blob/c8c376a57b2a515217f42299ba15d8e4d18e0757/server/src/main/java/com/vaadin/data/Binder.java#L1918)):*

Keeping it as is seems like the best option, because otherwise bind would change the state of the field as a side effect. The javadoc for this method could be improved to explicitly state that this is practically a shorthand for calling setReadOnly for all currently bound fields.
</blockquote></details>

Agreed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8241)*

ahie:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8241)*

PR #8249, opened 2017-01-16
Add separate modules to test Bean Validation lib necessity.

### LINKED ISSUE #8141: Vaadin 8 seems to require bean validation library, but this is not declared in documentation
With Vaadin 8.0.0.beta1 and BeanBinder you'll get an exception if bean validation library is not present. In previous versions Vaadin versions this was not required.

Either the functionality should be similar as before (just gracefully continue with limited featureset) or documentation and archetypes should be fixed.

See https://github.com/mstahv/v-leaflet/pull/144#issuecomment-270355264

denis-anisimov:
Fixes #8141

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8249)

### DIFF
--- a/server/src/main/java/com/vaadin/data/util/BeanUtil.java
+++ b/server/src/main/java/com/vaadin/data/util/BeanUtil.java
@@ -20,6 +20,7 @@ import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.io.Serializable;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
@@ -229,14 +230,21 @@ public final class BeanUtil implements Serializable {
 
         private static boolean isAvailable() {
             try {
-                Class.forName("javax.validation.Validation");
+                Class<?> clazz = Class.forName("javax.validation.Validation");
+                Method method = clazz.getMethod("buildDefaultValidatorFactory");
+                method.invoke(null);
                 return true;
-            } catch (ClassNotFoundException e) {
+            } catch (ClassNotFoundException | NoSuchMethodException
+                    | InvocationTargetException e) {
                 Logger.getLogger(BeanValidator.class.getName())
                         .fine("A JSR-303 bean validation implementation not found on the classpath. "
                                 + BeanValidator.class.getSimpleName()
                                 + " cannot be used.");
                 return false;
+            } catch (IllegalAccessException | IllegalArgumentException e) {
+                throw new RuntimeException(
+                        "Unable to invoke javax.validation.Validation.buildDefaultValidatorFactory()",
+                        e);
             }
         }
 
--- a/test/bean-api-validation/pom.xml
+++ b/test/bean-api-validation/pom.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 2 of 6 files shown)


ahie:
Reviewed 4 of 6 files at r1.
Review status: 4 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/util/BeanUtil.java, line 246 at r1](https://reviewable.io:443/reviews/vaadin/framework/8249#-KabSol9f7FWnVraOAvg:-KabSolA9evNjhoD6kmP:bt27vn0) ([raw file](https://github.com/vaadin/framework/blob/29b605aefc9fac33f48b394a8748de8d70069b9f/server/src/main/java/com/vaadin/data/util/BeanUtil.java#L246)):*
> ```Java
>             } catch (IllegalAccessException | IllegalArgumentException e) {
>                 throw new RuntimeException(
>                         "Unable to invoke javax.validation.Validation.buildDefaultValidatorFactory()",
> ```

Could be more descriptive since this seems like a bug that will be difficult to resolve, especially since this method is only evaluated once when the class is being initialized.

---

*[test/bean-impl-validation/src/test/java/com/vaadin/data/BeanValidationTest.java, line 46 at r1](https://reviewable.io:443/reviews/vaadin/framework/8249#-KabKHycxOew4UIXFuvK:-KabKHycxOew4UIXFuvL:b-bdlew8) ([raw file](https://github.com/vaadin/framework/blob/29b605aefc9fac33f48b394a8748de8d70069b9f/test/bean-impl-validation/src/test/java/com/vaadin/data/BeanValidationTest.java#L46)):*
> Without

With?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8249)*

denis-anisimov:
Review status: 4 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/util/BeanUtil.java, line 246 at r1](https://reviewable.io:443/reviews/vaadin/framework/8249#-KabSol9f7FWnVraOAvg:-KabVl8Mr_Oi6x0_2O1Q:b8l30ur) ([raw file](https://github.com/vaadin/framework/blob/29b605aefc9fac33f48b394a8748de8d70069b9f/server/src/main/java/com/vaadin/data/util/BeanUtil.java#L246)):*

Could be more descriptive since this seems like a bug that will be difficult to resolve, especially since this method is only evaluated once when the class is being initialized.
</blockquote></details>

This is something that can't happen at all unless Validation API has changed.
One exception means that `buildDefaultValidatorFactory` method is not public.
Another means that it requires arguments.

The method is used in the BeanValidator class where checks are done via the compiler.
Someone can of course forget to update this code once API method has changed since it's in the different place and is not checked via compiler.
But I have no any idea how to write more descriptive message here (exception tells details).

Anyway : this can't happen in runtime if Validation API is not changed (our it will fail along with the compiled code). But this will fail in our tests immediately once the API changes. So the message could internal only.

---

*[test/bean-impl-validation/src/test/java/com/vaadin/data/BeanValidationTest.java, line 46 at r1](https://reviewable.io:443/reviews/vaadin/framework/8249#-KabKHycxOew4UIXFuvK:-KabXd3upD8hLQZGPREb:bf8ybpp) ([raw file](https://github.com/vaadin/framework/blob/29b605aefc9fac33f48b394a8748de8d70069b9f/test/bean-impl-validation/src/test/java/com/vaadin/data/BeanValidationTest.java#L46)):*

> Without

With?
</blockquote></details>

Why with ?

This tests is intended to check that is works if Validation API available but there in NO IMPL available . So it's without validation impl. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8249)*

ahie:
:lgtm:

---

Reviewed 2 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8249#-:-Kabi0n8YAzA5FX6FwCX:bnfp4nl)*

PR #8273, opened 2017-01-18
Introduce Binder.addValueChangeListener

denis-anisimov:
The method Binder::addValueChangeListener allows to listen changes
value change events in any bound field.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8273)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -39,6 +39,7 @@ import java.util.stream.Stream;
 import com.googlecode.gentyref.GenericTypeReflector;
 import com.vaadin.annotations.PropertyId;
 import com.vaadin.data.HasValue.ValueChangeEvent;
+import com.vaadin.data.HasValue.ValueChangeListener;
 import com.vaadin.data.converter.StringToIntegerConverter;
 import com.vaadin.data.validator.BeanValidator;
 import com.vaadin.event.EventRouter;
@@ -928,6 +929,7 @@ public class Binder<BEAN> implements Serializable {
                     binderValidationResults);
             getBinder().getValidationStatusHandler().statusChange(status);
             getBinder().fireStatusChangeEvent(status.hasErrors());
+            getBinder().fireValueChangeEvent(event);
         }
 
         /**
@@ -1735,6 +1737,29 @@ public class Binder<BEAN> implements Serializable {
     }
 
     /**
+     * Adds field value change listener to all the fields in the binder.
+     * <p>
+     * Added listener is notified every time whenever any bound field value is
+     * changed. The same functionality can be achieved by adding a
+     * {@link ValueChangeListener} to all fields in the {@link Binder}.
+     * <p>
+     * The listener is added to all fields regardless of whether the method is
+     * invoked before or after field is bound.
+     * 
+     * @see ValueChangeEvent
+     * @see ValueChangeListener
+     *
+     * @param listener
+     *            a field value change listener
+     * @return a registration for the listener
+     */
+    public Registration addValueChangeListener(
+            ValueChangeListener<?> listener) {
+        return getEventRouter().addListener(ValueChangeEvent.class, listener,
+                ValueChangeListener.class.getDeclaredMethods()[0]);
+    }
+
+    /**
      * Creates a new binding with the given field.
      *
      * @param <FIELDVALUE>
@@ -2239,4 +2264,7 @@ public class Binder<BEAN> implements Serializable {
         return fieldName.toLowerCase(Locale.ENGLISH).replace("_", "");
     }
 
+    private <V> void fireValueChangeEvent(ValueChangeEvent<V> event) {
+        getEventRouter().fireEvent(event);
+    }
 }
--- a/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java
+++ b/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2000-2016 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.data;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.vaadin.data.Binder.BindingBuilder;
+import com.vaadin.data.HasValue.ValueChangeEvent;
+import com.vaadin.data.converter.StringToIntegerConverter;
+import com.vaadin.tests.data.bean.Person;
+import com.vaadin.ui.TextField;
+
+/**
+ * @author Vaadin Ltd
+ *
+ */
+public class BinderValueChangeTest
+        extends BinderTestBase<Binder<Person>, Person> {
+
+    private AtomicReference<ValueChangeEvent<?>> event;
+
+    private static class TestTextField extends TextField {
+        @Override
+        protected boolean setValue(String value, boolean userOriginated) {
+            return super.setValue(value, userOriginated);
+        }
+    }
+
+    @Before
+    public void setUp() {
+        binder = new Binder<>();
+        item = new Person();
+        event = new AtomicReference<>();
+    }
+
+    @Test
+    public void unboundField_noEvents() {
+        binder.addValueChangeListener(this::statusChanged);
+
+        BindingBuilder<Person, String> binding = binder.forField(nameField);
+
+        nameField.setValue("");
+        Assert.assertNull(event.get());
+
+        binding.bind(Person::getFirstName, Person::setFirstName);
+        Assert.assertNull(event.get());
+    }
+
+    @Test
+    public void setBean_unbound_noEvents() {
+        binder.addValueChangeListener(this::statusChanged);
+
+        Assert.assertNull(event.get());
+
+        binder.setBean(item);
+
+        Assert.assertNull(event.get());
+    }
+
+    @Test
+    public void readBean_unbound_noEvents() {
+        binder.addValueChangeListener(this::statusChanged);
+
+        Assert.assertNull(event.get());
+
+        binder.readBean(item);
+
+        Assert.assertNull(event.get());
+    }
+
+    @Test
+    public void setValue_unbound_singleEventOnSetValue() {
+        binder.forField(nameField).bind(Person::getFirstName,
+                Person::setFirstName);
+        binder.forField(ageField)
+                .withConverter(new StringToIntegerConverter(""))
+                .bind(Person::getAge, Person::setAge);
+
+        binder.addValueChangeListener(this::statusChanged);
+
+        Assert.assertNull(event.get());
+        nameField.setValue("foo");
+        verifyEvent(nameField);
+    }
+
+    @Test
+    public void setValue_bound_singleEventOnSetValue() {
+        binder.forField(nameField).bind(Person::getFirstName,
+                Person::setFirstName);
+        binder.forField(ageField)
+                .withConverter(new StringToIntegerConverter(""))
+                .bind(Person::getAge, Person::setAge);
+        binder.setBean(item);
+
+        binder.addValueChangeListener(this::statusChanged);
+
+        Assert.assertNull(event.get());
+        nameField.setValue("foo");
+        verifyEvent(nameField);
+    }
+
+    @Test
+    public void userOriginatedUpdate_unbound_singleEventOnSetValue() {
+        TestTextField field = new TestTextField();
+
+        binder.forField(field).bind(Person::getFirstName, Person::setFirstName);
+        binder.forField(ageField)
+                .withConverter(new StringToIntegerConverter(""))
+                .bind(Person::getAge, Person::setAge);
+
+        binder.addValueChangeListener(this::statusChanged);
+
+        Assert.assertNull(event.get());
+        field.setValue("foo", true);
+        verifyEvent(field, true);
+    }
+
+    @Test
+    public void addListenerFirst_bound_singleEventOnSetValue() {
+        binder.addValueChangeListener(this::statusChanged);
+
+        binder.forField(nameField).bind(Person::getFirstName,
+                Person::setFirstName);
+        binder.forField(ageField)
+                .withConverter(new StringToIntegerConverter(""))
+                .bind(Person::getAge, Person::setAge);
+        binder.setBean(item);
+
+        Assert.assertNull(event.get());
+        ageField.setValue(String.valueOf(1));
+        verifyEvent(ageField);
+    }
+
+    private void verifyEvent(HasValue<?> field) {
+        verifyEvent(field, false);
+    }
+
+    private void verifyEvent(HasValue<?> field, boolean isUserOriginated) {
+        ValueChangeEvent<?> changeEvent = event.get();
+        Assert.assertNotNull(changeEvent);
+        Assert.assertEquals(field, changeEvent.getSource());
+        Assert.assertEquals(field, changeEvent.getComponent());
+        Assert.assertEquals(isUserOriginated, changeEvent.isUserOriginated());
+    }
+
+    private void statusChanged(ValueChangeEvent<?> evt) {
+        Assert.assertNull(event.get());
+        event.set(evt);
+    }
+}


ahie:
Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java, line 35 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauoY3OKlGCC0Jxr7Qt:-KauoY3OKlGCC0Jxr7Qu:b-i36kjw) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java#L35)):*
> extends EventObject

Could this be changed to extend ValueChangeEvent?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

ahie:
Review status: 0 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1983 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauyJmeP1vzBROZi-E5:-KauyJmfZdQ4l2Vxq2Lc:b-j4xtsd) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1983)):*
> ```Java
>     private void fireValidationStatusChange(
>             BinderValidationStatus<BEAN> status) {
>         if (!status.equals(currentValidationStatus)) {
> ```

This will in some cases evaluate to true, even though the validation status hasn't changed. For example in the following case:

- set a valid bean and call validate: currentValidationStatus will contain a list of `BindingValidationStatus`es for all bindings as valid
- change a field's value to another valid value: this method will be called with a `status`, whose list only contains the `BindingValidationStatus` for a single binding.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1985 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kav1pENMh3loDYHK_v7:-Kav1pENMh3loDYHK_v8:b-6uul64) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1985)):*
> ```Java
>         if (!status.equals(currentValidationStatus)) {
>             currentValidationStatus = status;
>             getValidationStatusHandler().statusChange(status);
> ```

On the other hand, if this were to be called with the full list after a single field's value has changed it would notify every single binding's status handler, if using the current default binder status handler.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

denis-anisimov:
Review status: 0 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1983 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauyJmeP1vzBROZi-E5:-KawR6yYhSfsUVNkYWGM:bavua81) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1983)):*

This will in some cases evaluate to true, even though the validation status hasn't changed. For example in the following case:

- set a valid bean and call validate: currentValidationStatus will contain a list of `BindingValidationStatus`es for all bindings as valid
- change a field's value to another valid value: this method will be called with a `status`, whose list only contains the `BindingValidationStatus` for a single binding.
</blockquote></details>

Good catch, but......

There are two cases:
- binder is bound 
- binder is unbound.

If binder is bound then this check works correctly (am I wrong ?) because both `validate()` and `handleFieldValueChange()` (which is the action on value change) walks through all bindings to do validation.

So only the second case introduces some inconsistency. But this inconsistency is already there:  `getValidationStatusHandler().statusChange()` is called with the status which doesn't contain the _whole_ binder status. It contains status which is "calculated" for the given field only. And from this point of view the status has really changed.
So either we should change the logic which fires the event (it should track all the changes in all bindings and compose an event based on previous state) or this logic is consistent with the current approach.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1985 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kav1pENMh3loDYHK_v7:-KawTFeORaeknixY5Hp1:bvmbydd) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1985)):*

On the other hand, if this were to be called with the full list after a single field's value has changed it would notify every single binding's status handler, if using the current default binder status handler.
</blockquote></details>

See above: it's just a consequence of the current way of firing the status change event.

---

*[server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java, line 35 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauoY3OKlGCC0Jxr7Qt:-KawUuMlxHpR3tZrPto6:bccoz9m) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java#L35)):*

> extends EventObject

Could this be changed to extend ValueChangeEvent?
</blockquote></details>

Good question.
I don't know.....

ValueChangeEvent contains nowadays an old value and a new value. 
Is it worth to provide them in this event ?
May be yes, may be now.
Looks like it requires discussion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

caalador:
Review status: 0 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java, line 35 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauoY3OKlGCC0Jxr7Qt:-KbFA3SjhHnMburfOmsS:b-jj2s1r) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java#L35)):*

Good question.
I don't know.....

ValueChangeEvent contains nowadays an old value and a new value. 
Is it worth to provide them in this event ?
May be yes, may be now.
Looks like it requires discussion.

</blockquote></details>

I could expect that anyone wanting to listen to changes in all bound fields would probably like to have the original ValueChangeEvent from the field also.
In such a case it would be good to extend from ValueChangeEvent so all the same information would be available as if you were listening to a fields ValueChangeEvent.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Reviewed 3 of 9 files at r1.
Review status: 3 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1983 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauyJmeP1vzBROZi-E5:-KbnUwOCqvHZgz9KOxQC:b-dvsfsr) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1983)):*

Good catch, but......

There are two cases:
- binder is bound 
- binder is unbound.

If binder is bound then this check works correctly (am I wrong ?) because both `validate()` and `handleFieldValueChange()` (which is the action on value change) walks through all bindings to do validation.

So only the second case introduces some inconsistency. But this inconsistency is already there:  `getValidationStatusHandler().statusChange()` is called with the status which doesn't contain the _whole_ binder status. It contains status which is "calculated" for the given field only. And from this point of view the status has really changed.
So either we should change the logic which fires the event (it should track all the changes in all bindings and compose an event based on previous state) or this logic is consistent with the current approach.
</blockquote></details>

I'm a bit concerned at introducing a new feature and "optimizing" another in the same PR. I was not perhaps clear enough in my comment on the issue "This feature is for the latter" meaning that should just focus on adding the feature, and optimize separately.

Specially since the latter has caused this to be stuck. We should look at the use cases separately for figuring out how it should work, and I'd prefer if it would not be in this same patch, but another ticket.

---

*[server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java, line 35 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauoY3OKlGCC0Jxr7Qt:-KbnQeNEYih9BP9h44TE:b-20ytb9) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java#L35)):*

Actually I see now quite strong reason why it should not extend `ValueChangeEvent`:
the latter class is parameterized with field value `V`. It means that `FieldValueChangeEvent` should contain two parameters `<BEAN,V>` and the second one will be useless since you never know which type `V` is when the event is fired via the `Binder`.
So `FieldValueChangeEvent` will never have information about the second parameter.

`ValueChangeEvent<V>` is strongly connected to `HasValue<V>` and it's not really useful without type information.
</blockquote></details>

To me at first, the prefix Field* is confusing, as it somehow makes this a duplicate of `ValueChangeEvent` (it is always from a Field but just doesn't have the prefix), but it is not.
Probably it is not an `BinderValueChangeEvent` either. 

How about just ignoring the `ValueChangeEvent<V>` type here ? If you are specifically interested in some certain event's type,  you should use a value change listener in the specific field. To my understanding, this might be more like doing some universal actions. Probably should take a look at the one suggested use case for this.

At least one user has noted that just wants a shorthand for not having to add ValueChangeListener to all fields.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

denis-anisimov:
Reviewed 1 of 9 files at r1.
Review status: 4 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1983 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauyJmeP1vzBROZi-E5:-KbnxQWPTugIr7TG5WOi:blkuofo) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1983)):*

I'm a bit concerned at introducing a new feature and "optimizing" another in the same PR. I was not perhaps clear enough in my comment on the issue "This feature is for the latter" meaning that should just focus on adding the feature, and optimize separately.

Specially since the latter has caused this to be stuck. We should look at the use cases separately for figuring out how it should work, and I'd prefer if it would not be in this same patch, but another ticket.
</blockquote></details>

OK , I will change the PR to introduce value change listener only.

---

*[server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java, line 35 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauoY3OKlGCC0Jxr7Qt:-Kbnvma_5KY5ihkBzeNo:b-1df82u) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/FieldValueChangeEvent.java#L35)):*

To me at first, the prefix Field* is confusing, as it somehow makes this a duplicate of `ValueChangeEvent` (it is always from a Field but just doesn't have the prefix), but it is not.
Probably it is not an `BinderValueChangeEvent` either. 

How about just ignoring the `ValueChangeEvent<V>` type here ? If you are specifically interested in some certain event's type,  you should use a value change listener in the specific field. To my understanding, this might be more like doing some universal actions. Probably should take a look at the one suggested use case for this.

At least one user has noted that just wants a shorthand for not having to add ValueChangeListener to all fields.


</blockquote></details>

The problem here is that `ValueChangeEvent` is based on a `Component` as a source.
But event should contain `Binder`. It should be considered as a source.
And it's really useful. But it won't be possible to get a `Binder` reference from the event (or component).

So it should be another event.
Thought I can derive it from ValueChangeEvent.......

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Reviewed 5 of 12 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/data/BinderChangeEvent.java, line 36 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp3KJvfMv5CRIaIGXl:-Kbp3KJwP9n5tmG7Agss:b-w4g2jc) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/main/java/com/vaadin/data/BinderChangeEvent.java#L36)):*
> ```Java
>  * @see BinderChangeListener
>  */
> public class BinderChangeEvent<BEAN, V> extends ValueChangeEvent<V> {
> ```

IMO it should be `BinderValueChangeEvent`

---

*[server/src/main/java/com/vaadin/data/BinderChangeListener.java, line 33 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp3eDt6rRwP2ID3hrB:-Kbp3eDt6rRwP2ID3hrC:b-i1wvj0) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/main/java/com/vaadin/data/BinderChangeListener.java#L33)):*
> ```Java
>  */
> @FunctionalInterface
> public interface BinderChangeListener<BEAN> {
> ```

Same here, `BinderValueChangeListener` 

---

*[server/src/test/java/com/vaadin/data/BinderValueChangeTest.java, line 127 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp484LvgyKiX59D9fW:-Kbp484LvgyKiX59D9fX:b-4voprm) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java#L127)):*
> ```Java
> 
>     private void verifyEvent(HasValue<?> field) {
>         BinderChangeEvent<Person, ?> statusChangeEvent = event.get();
> ```

kind of misleading name, statusChangeEvent when it should be valueChangeEvent

---

*[server/src/test/java/com/vaadin/data/BinderValueChangeTest.java, line 132 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp4L87aImUQkCEm-bY:-Kbp4L88NgQ3yL5K0n2D:b-lxjy2z) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java#L132)):*
> ```Java
>         Assert.assertEquals(field, statusChangeEvent.getSource());
>         Assert.assertEquals(field, statusChangeEvent.getComponent());
>         Assert.assertFalse(statusChangeEvent.isUserOriginated());
> ```

NIT: now not testing that user originated value changes are passed on to this event.

Could mock a rpc call, like in e.g. CheckBoxTest

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Reviewed 1 of 12 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/data/BinderChangeEvent.java, line 38 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-KbsU4643LjjlsBvBxmW:-KbsU4643LjjlsBvBxmX:brribzx) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/main/java/com/vaadin/data/BinderChangeEvent.java#L38)):*
> ```Java
> public class BinderChangeEvent<BEAN, V> extends ValueChangeEvent<V> {
> 
>     private final Binder<BEAN> binder;
> ```

Do you have a use case where it is actually needed to have the binder reference ?

As discussed in flowdock, let's just use the existing `ValueChangeListener` and `ValueChangeEvent`, the developer can figure out a way to get the binder if they ever run into the case where they need it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

denis-anisimov:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/data/BinderChangeEvent.java, line 36 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp3KJvfMv5CRIaIGXl:-KbsWxgaQHM9K0ybDv6X:bwu3oow) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/main/java/com/vaadin/data/BinderChangeEvent.java#L36)):*

IMO it should be `BinderValueChangeEvent`
</blockquote></details>

Whatever you say.
I almost broke my mind trying to find appropriate name.

---

*[server/src/main/java/com/vaadin/data/BinderChangeEvent.java, line 38 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-KbsU4643LjjlsBvBxmW:-KbsXPv8snpKYPVajImp:b-4i8rg8) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/main/java/com/vaadin/data/BinderChangeEvent.java#L38)):*
From the original description of the ticket: 

>Event should have binder reference, and the event source is the field/select/whatever changed. Should verify any book examples and have tests.

I can do anything. Let's just agree exactly what should be done.

---

*[server/src/main/java/com/vaadin/data/BinderChangeListener.java, line 33 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp3eDt6rRwP2ID3hrB:-KbsXkD9KPhMpb9gdooq:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/main/java/com/vaadin/data/BinderChangeListener.java#L33)):*

Same here, `BinderValueChangeListener` 
</blockquote></details>

Done.

---

*[server/src/test/java/com/vaadin/data/BinderValueChangeTest.java, line 127 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp484LvgyKiX59D9fW:-KbsXwVEK8H8QBglDzmH:brcbfs7) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java#L127)):*

kind of misleading name, statusChangeEvent when it should be valueChangeEvent
</blockquote></details>

copy/paste

Thanks.

---

*[server/src/test/java/com/vaadin/data/BinderValueChangeTest.java, line 132 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp4L87aImUQkCEm-bY:-KbsYBVjFNOJ_F9rVcKu:b-4c6nn9) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java#L132)):*

NIT: now not testing that user originated value changes are passed on to this event.

Could mock a rpc call, like in e.g. CheckBoxTest
</blockquote></details>

Will write a test for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Reviewed 1 of 12 files at r2, 4 of 4 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1740 at r3](https://reviewable.io:443/reviews/vaadin/framework/8273#-KbtehftIS7aCx4mjx8a:-KbtehftIS7aCx4mjx8b:b-50eg0m) ([raw file](https://github.com/vaadin/framework/blob/a9c8be3446f46acb91fc8bbd758a3409872a8509/server/src/main/java/com/vaadin/data/Binder.java#L1740)):*
> ```Java
> 
>     /**
>      * Adds field value change listener to the binder.
> ```

Now I think it should read,
"..listener to all the fields in the binder."

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1741 at r3](https://reviewable.io:443/reviews/vaadin/framework/8273#-KbteoUE-a8VkRA0ZWxR:-KbteoUE-a8VkRA0ZWxS:b2yssws) ([raw file](https://github.com/vaadin/framework/blob/a9c8be3446f46acb91fc8bbd758a3409872a8509/server/src/main/java/com/vaadin/data/Binder.java#L1741)):*
> ```Java
>     /**
>      * Adds field value change listener to the binder.
>      * <p>
> ```

Could also mention that it is automatically added for any new fields too 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Just to mention, this doesn't add anything to the documentation, and I'm not sure what the use case would be that could be presented there.

---

Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273#-:-KbtgSQMW7ny5-8esFGy:bbyrvqw)*

denis-anisimov:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1983 at r1](https://reviewable.io:443/reviews/vaadin/framework/8273#-KauyJmeP1vzBROZi-E5:-KbtkkOA9wAAJCwZR7sI:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/333a08f5ac341a78e944a1d21758b261912ecb32/server/src/main/java/com/vaadin/data/Binder.java#L1983)):*

OK , I will change the PR to introduce value change listener only.
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1740 at r3](https://reviewable.io:443/reviews/vaadin/framework/8273#-KbtehftIS7aCx4mjx8a:-KbthbH_rjrI-cqImn2U:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/a9c8be3446f46acb91fc8bbd758a3409872a8509/server/src/main/java/com/vaadin/data/Binder.java#L1740)):*

Now I think it should read,
"..listener to all the fields in the binder."
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1741 at r3](https://reviewable.io:443/reviews/vaadin/framework/8273#-KbteoUE-a8VkRA0ZWxR:-KbtiJuKNxr6jS8ACXW5:b-tb66qi) ([raw file](https://github.com/vaadin/framework/blob/a9c8be3446f46acb91fc8bbd758a3409872a8509/server/src/main/java/com/vaadin/data/Binder.java#L1741)):*

Could also mention that it is automatically added for any new fields too 
</blockquote></details>

Not sure about wording, but it's done somehow.

---

*[server/src/test/java/com/vaadin/data/BinderValueChangeTest.java, line 132 at r2](https://reviewable.io:443/reviews/vaadin/framework/8273#-Kbp4L87aImUQkCEm-bY:-Kbtkx6Exvt__GWxlTWN:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/82d7cac6ae50f1bf75745b0e51ba6369c14961f6/server/src/test/java/com/vaadin/data/BinderValueChangeTest.java#L132)):*

Will write a test for that.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273)*

pleku:
Also another reminder on fixing the commit message before merging

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8273#-:-KbtnZ7av7IUZWTOH3CO:b1wq6fs)*

Issue #8287, opened 2017-01-20
Vaadin 8 Binder - Binding a null bean should clear the bound field values


caalador:
When using binder there is now no simple way of actually clearing all fields of values when we remove a bound bean.

Setting a null bean should clear all the bound fields like setting a null Datasource in Vaadin7 FieldGroup clears all fields.

Also for non-immediate beans we should have a clear method that clears all fields if no bean bound

PR #8288, opened 2017-01-20
Binding a null bean should clear the bound field values

caalador:
fixes issue #8287

Added functionality to make it possible to clear bound field values manually
or when setting a null bean to binder.

Binder will now clear bound fields for setBean(null) or if using readBean() functionality
when calling the clear() method. Binder.clear() will throw a IllegalStateException if
we have a bound bean.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8288)


Legioth:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavCR-P6hKuQ4Ky03-f:-KavCR-P6hKuQ4Ky03-g:bkguv2x) (no related file):*
What's the reasoning behind making `clear()` throw if a bean has been set? The other alternative would be to clear all fields and let that propagate null or empty values to the setters in the bean.

I'm also pondering whether it would make sense to make `readBean(null)` also trigger a clear? In that case, we could also consider not introducing `clear()` as a public method at all if we feel that it might be confusing in cases where a bean has been set.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavCR-P6hKuQ4Ky03-f:-KavDQt1RQ2gpRTpAQjf:b-24rej5) (no related file):*

What's the reasoning behind making `clear()` throw if a bean has been set? The other alternative would be to clear all fields and let that propagate null or empty values to the setters in the bean.

I'm also pondering whether it would make sense to make `readBean(null)` also trigger a clear? In that case, we could also consider not introducing `clear()` as a public method at all if we feel that it might be confusing in cases where a bean has been set.
</blockquote></details>

The reasoning behind not letting clear go through when a bean is set is that it leaves the fields in an inconsistent state when seen against the bean that is bound as it won't update the bean values that happens when editing fields with a bound bean. 

readBean(null) would make things easier on the application development side as then you could just have 
```
	public void setPerson(Person person) {
		binder.readBean(person);
	}
```
instead of 
```
	public void setPerson(Person person) {
		if(person != null)
			binder.readBean(person);
		else
			binder.clear();
	}
```

I  would fully support this change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavCR-P6hKuQ4Ky03-f:-KavGFBO5VjdZ1XFSHco:bf247dc) (no related file):*
> The reasoning behind not letting clear go through when a bean is set is that it leaves the fields in an inconsistent state when seen against the bean that is bound as it won't update the bean values that happens when editing fields with a bound bean.

What if it would be implemented so that the bean is also updated?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavCR-P6hKuQ4Ky03-f:-KavHj46t9kwNT28okv-:b-5ws0e6) (no related file):*

> The reasoning behind not letting clear go through when a bean is set is that it leaves the fields in an inconsistent state when seen against the bean that is bound as it won't update the bean values that happens when editing fields with a bound bean.

What if it would be implemented so that the bean is also updated?
</blockquote></details>

Then I would pose the question of would anyone want to clear their bean of field data?

I feel that it would be an unexpected thing to happen.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavCR-P6hKuQ4Ky03-f:-KavSTrZSG4ktfP0sxRI:b-belzex) (no related file):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

Then I would pose the question of would anyone want to clear their bean of field data?

I feel that it would be an unexpected thing to happen.
</blockquote></details>

I would then suggest that we don't make `clear()` public at all at this point since any alternative would in some cases be confusing, and the method is not really needed if you can do either `setBean(null)` or ` readBean(null)` depending on how you have been using `Binder`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

ahie:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1565 at r1](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavbZE_a2ucDR2T_9xa:-KavbZE_a2ucDR2T_9xb:b-e4zva5) ([raw file](https://github.com/vaadin/framework/blob/8c120c5cc157f5bee45e0c02f804e73f92f846f8/server/src/main/java/com/vaadin/data/Binder.java#L1565)):*
> ```Java
>         for (Binding binding : bindings) {
>             HasValue field = binding.getField();
>             if (field.isReadOnly()) {
> ```

I believe HasValue being read only should only affect attempted changes from the client side, thus this check shouldn't be necessary. At least `AbstractField` allows the server side to still set values even though it has been set to read only.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1565 at r1](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavbZE_a2ucDR2T_9xa:-Kavd5J3bWJV_7_nO7JN:b-1cn777) ([raw file](https://github.com/vaadin/framework/blob/8c120c5cc157f5bee45e0c02f804e73f92f846f8/server/src/main/java/com/vaadin/data/Binder.java#L1565)):*

I believe HasValue being read only should only affect attempted changes from the client side, thus this check shouldn't be necessary. At least `AbstractField` allows the server side to still set values even though it has been set to read only.
</blockquote></details>

That is true. The `userOriginated` didn't for some reason register and I just noted that `isReadOnly() -> return`

I'll fix this part and add a test or two

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1565 at r1](https://reviewable.io:443/reviews/vaadin/framework/8288#-KavbZE_a2ucDR2T_9xa:-Kb9ZzK4jTdk_k-uBx8-:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/8c120c5cc157f5bee45e0c02f804e73f92f846f8/server/src/main/java/com/vaadin/data/Binder.java#L1565)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

That is true. The `userOriginated` didn't for some reason register and I just noted that `isReadOnly() -> return`

I'll fix this part and add a test or two
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

ahie:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1327 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-Kb9zj249hBnysdDMwOT:-Kb9zj25bRwsyu4WlZiw:bnqfu6w) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1327)):*
> or {@code null} to remove a currently bound bean and clear bound fields

Should probably also document this behavior in removeBean, which is a shorthand for setBean(null).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1327 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-Kb9zj249hBnysdDMwOT:-KbA1LqUx2JxoFVdWcyW:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1327)):*

> or {@code null} to remove a currently bound bean and clear bound fields

Should probably also document this behavior in removeBean, which is a shorthand for setBean(null).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

denis-anisimov:
Reviewed 1 of 3 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbA5NMcD2ffC_G83Gzp:b3b4pu5) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
> ```Java
>      */
>     public void readBean(BEAN bean) {
>         if(bean == null) {
> ```

So it doesn't do anything below when the bean instance is `null`.
Isn't it confusing ?
What's the usecase for this  ?

The method doesn't do anything except calling `clearFields()`. Why it can't be just a public method instead of providing some "undocumented" functionality for `readBean()` ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

Legioth:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbA9q5L13zI0T1rwHkj:b-fuih1y) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*

So it doesn't do anything below when the bean instance is `null`.
Isn't it confusing ?
What's the usecase for this  ?

The method doesn't do anything except calling `clearFields()`. Why it can't be just a public method instead of providing some "undocumented" functionality for `readBean()` ?
</blockquote></details>

The original reason was that making `clearFields()` public would be confusing for someone who as been using `setBean` since it would then also clear out the values in that bean instance.

I'm however starting to feel that it might actually make sense to have it work in that way.

Doing a clear for `readBean(null)` is still also useful since you'd otherwise in many cases have to do the `null` check in application code an then either call `readBean` or `clearFields`.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r3](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA9eUBsOqymDc5AKUT:-KbA9eUBsOqymDc5AKUU:b-ck8xdg) ([raw file](https://github.com/vaadin/framework/blob/cb1614af88712462be3393b73fee6e7a5b9fa665/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
> ```Java
>      */
>     public void readBean(BEAN bean) {
>         if(bean == null) {
> ```

The code formatting is by the way broken on this line (no space between `if` and `(`). Please make sure you have your IDE set up to automatically format whenever you save.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbA8ZKSKzMqG9yFy7Bs:b9par5) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
We went through this with Leif that we don't want a public clear method

> I would then suggest that we don't make clear() public at all at this point since any alternative would in some cases be confusing, and the method is not really needed if you can do either setBean(null) or readBean(null) depending on how you have been using Binder.

The case is also that if you now `readBean(null)` you clear the fields, whereas in the existing implementation the call `readBean(null)` will throw a `NullPointerException` as it does a `Objects.requireNonNull(...)`

The NPE is problematic in that on the application side you would have to do lots of checks and call different methods depending on for instance if anything is selected in a grid. Being able to just call `readBean(NullableItem)` makes things much clearer than having to explicitly call a clear() method.

Also it is documented in the @param

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbAP7PH3eZo2ARKd25O:b-ljzbnb) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

We went through this with Leif that we don't want a public clear method

> I would then suggest that we don't make clear() public at all at this point since any alternative would in some cases be confusing, and the method is not really needed if you can do either setBean(null) or readBean(null) depending on how you have been using Binder.

The case is also that if you now `readBean(null)` you clear the fields, whereas in the existing implementation the call `readBean(null)` will throw a `NullPointerException` as it does a `Objects.requireNonNull(...)`

The NPE is problematic in that on the application side you would have to do lots of checks and call different methods depending on for instance if anything is selected in a grid. Being able to just call `readBean(NullableItem)` makes things much clearer than having to explicitly call a clear() method.

Also it is documented in the @param


</blockquote></details>

`null` is not an issue.
I'm thinking about the logic which is skipped by `return` in the check.

It's a question , not a complain: are you sure that neither of subsequent calls have sense in case bean is null ?
(like `setHasChange(false)` , no status change events and so forth).
So we are doing all those things if the bean is not null and we don't do them is it's set to `null`. I'm not sure that they have no sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbAQ7PVw5cwHIGYLPfe:bkgf0ln) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*

`null` is not an issue.
I'm thinking about the logic which is skipped by `return` in the check.

It's a question , not a complain: are you sure that neither of subsequent calls have sense in case bean is null ?
(like `setHasChange(false)` , no status change events and so forth).
So we are doing all those things if the bean is not null and we don't do them is it's set to `null`. I'm not sure that they have no sense.
</blockquote></details>

The subsequent code can be wrapped in an else if you don't like the `return;`
the current implementation is:
```
    public void readBean(BEAN bean) {
        Objects.requireNonNull(bean, "bean cannot be null");
```

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r3](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA9eUBsOqymDc5AKUT:-KbAN16fRRM7TBWBMUgr:bi3b7tb) ([raw file](https://github.com/vaadin/framework/blob/cb1614af88712462be3393b73fee6e7a5b9fa665/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*

The code formatting is by the way broken on this line (no space between `if` and `(`). Please make sure you have your IDE set up to automatically format whenever you save.
</blockquote></details>

In Idea, I do not save. 
It's just that the Eclipse rules don't 100% align in idea so I never format the whole file, but I wouldn't want to go back to eclipse either... 
We should just have a maven formatting target.

Formatting fixed for this change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

denis-anisimov:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbAR1UQgc29qCKKU8sD:bmffzl5) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

The subsequent code can be wrapped in an else if you don't like the `return;`
the current implementation is:
```
    public void readBean(BEAN bean) {
        Objects.requireNonNull(bean, "bean cannot be null");
```
</blockquote></details>

And again : no issues with `return` itself.
The question is : should we return right away without notifying listeners about status change and state updates (like "hasChanges" state).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbE-M7RHBpiG_0hUR9u:bq32gjg) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
> The question is : should we return right away without notifying listeners about status change and state updates (like "hasChanges" state).

Now I'm on the same page. After mulling this through over the night I would say that no I would probably not want or expect a status change event. Same as it is right now for `removeBean()` and `setBean(null)`.

The events are right now run for:
```
// if there has been field value change listeners that trigger
// validation, need to make sure the validation errors are cleared
```

I would put this on under question as in cases where I bind a bean I know has validation errors and expect the StatusChangeEvent to have hasValidationErrors to be true I right now actually always get false on set as it's not the actual validation status.
The case here is for instance to have a save button enable/disable on validation status that is checked on StatusChangeEvent, which doesn't work now as the status if always 'no validation errors'.

Clearing the validation errors for the fields is ok, but the validation status should be by the bean.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r3](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA9eUBsOqymDc5AKUT:-KbE39Y8gTojmSS1Vp6_:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/cb1614af88712462be3393b73fee6e7a5b9fa665/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

In Idea, I do not save. 
It's just that the Eclipse rules don't 100% align in idea so I never format the whole file, but I wouldn't want to go back to eclipse either... 
We should just have a maven formatting target.

Formatting fixed for this change.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

Legioth:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbE7l8hbzzD4JRopQ2_:b-vvybdz) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

> The question is : should we return right away without notifying listeners about status change and state updates (like "hasChanges" state).

Now I'm on the same page. After mulling this through over the night I would say that no I would probably not want or expect a status change event. Same as it is right now for `removeBean()` and `setBean(null)`.

The events are right now run for:
```
// if there has been field value change listeners that trigger
// validation, need to make sure the validation errors are cleared
```

I would put this on under question as in cases where I bind a bean I know has validation errors and expect the StatusChangeEvent to have hasValidationErrors to be true I right now actually always get false on set as it's not the actual validation status.
The case here is for instance to have a save button enable/disable on validation status that is checked on StatusChangeEvent, which doesn't work now as the status if always 'no validation errors'.

Clearing the validation errors for the fields is ok, but the validation status should be by the bean.



</blockquote></details>

I think `hasChanges` should change to `false` since the values in all fields will be exactly in sync with the values in the "bean". This should in turn trigger a status change event at least in the case where `hasChanges` was previously `true`.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r3](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA9eUBsOqymDc5AKUT:-KbE7gIjEk_qjqE13Pg0:b54436y) ([raw file](https://github.com/vaadin/framework/blob/cb1614af88712462be3393b73fee6e7a5b9fa665/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Still no space between `if` and `(` on this particular line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KbEB1p04_GrRVmpgsLP:b9f262k) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*

I think `hasChanges` should change to `false` since the values in all fields will be exactly in sync with the values in the "bean". This should in turn trigger a status change event at least in the case where `hasChanges` was previously `true`.
</blockquote></details>

Agreed and added. 
Now each clearFields() call will reset hasChanges to false.

Still creating tests.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r3](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA9eUBsOqymDc5AKUT:-KbEBM_2S85gW36DWLOG:bwlw90z) ([raw file](https://github.com/vaadin/framework/blob/cb1614af88712462be3393b73fee6e7a5b9fa665/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*

Still no space between `if` and `(` on this particular line.
</blockquote></details>

That's due to the fact that I haven't pushed the change as we still have the whole above discussion in limbo.

I guess I'll need a 'how we use Reviewable and push commits' manual and then share that to Kirill and all new persons.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:
After coming over the question of is it ok to clear all fields when user has a bound bean through `setBean` and then calls `readBean(null)` to clear all fields which will then propagate to the bound bean, we decided to put this change on hold for now.

pleku:
Reviewed 2 of 2 files at r4.

---

*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 340 at r4](https://reviewable.io:443/reviews/vaadin/framework/8288#-Kbtq7-HS7MnRhCxtsyo:-Kbtq7-HS7MnRhCxtsyp:btrv3hu) ([raw file](https://github.com/vaadin/framework/blob/6e4a1e2f83647cd8b4ab2ed37fd2a76215bacbc2/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L340)):*
> ```Java
> 
>             @Override
>             public Set<T> getEmptyValue() {
> ```

Good catch, this probably should be a default method in `MultiSelect` interface with overridden javadocs that mention that it returns an empty set if nothing selected. Then you can remove the implementation from `AbstractMultiSelect too.

Also use Collections.emptySet() instead of initializing a new set.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

caalador:
Review status: 2 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 340 at r4](https://reviewable.io:443/reviews/vaadin/framework/8288#-Kbtq7-HS7MnRhCxtsyo:-KbxmLFJyPIq7CWtjNsi:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/6e4a1e2f83647cd8b4ab2ed37fd2a76215bacbc2/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L340)):*

Good catch, this probably should be a default method in `MultiSelect` interface with overridden javadocs that mention that it returns an empty set if nothing selected. Then you can remove the implementation from `AbstractMultiSelect too.

Also use Collections.emptySet() instead of initializing a new set.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

Legioth:
Shouldn't `Add getEmptyValue as default to MultiSelect` be a separate PR since it also makes lots of sense regardless of this features that is on hold?

---

Review status: 2 of 4 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288#-:-KbxmfoOtcJcg2cAPGUZ:b589i14)*

caalador:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1375 at r2](https://reviewable.io:443/reviews/vaadin/framework/8288#-KbA5NMb5qdjt_LElIok:-KcWWStl2XvBI8-bM_Xm:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/93c21468caee96fa2793a4003d846b4c17761d97/server/src/main/java/com/vaadin/data/Binder.java#L1375)):*
<details><summary><i>Previously, caalador wrote…</i></summary><blockquote>

Agreed and added. 
Now each clearFields() call will reset hasChanges to false.

Still creating tests.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

pleku:
Reviewed 2 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8288)*

sknitelius:
Sorry to dig out this old pull request, but a Binder.clear() / Binder.resetFields() function would be far more intuitive then Binder.setBean(null) / Binder.readBean(null);

PR #8291, opened 2017-01-20
Remove custom preloading support and load scripts using async=false

### LINKED ISSUE #5339: Use async = false instead of prefetching scripts in all modern browsers
**Originally by @Legioth**

___
#13956 fixes script prefetching by using async = false instead of loading as `text/cache` in IE11. This approach is generally recommendable for other supporting browsers as well but was not change as part of that patch to reduce the regression risk.

We should consider changing the default to use feature detection and use async = false for all supporting browsers. This should also get rid of anything resembling #11275.

___
 Imported from https://dev.vaadin.com/ issue #14066

Artur-:
When using async='false' for scripts created by scripts
the execution order is guaranteed to be the same as the order the
script tags are created

Fixes #5339, #3631

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8291)

### DIFF
--- a/client/src/main/java/com/vaadin/client/DependencyLoader.java
+++ b/client/src/main/java/com/vaadin/client/DependencyLoader.java
@@ -113,12 +113,6 @@ public class DependencyLoader {
         ResourceLoadListener resourceLoadListener = new ResourceLoadListener() {
             @Override
             public void onLoad(ResourceLoadEvent event) {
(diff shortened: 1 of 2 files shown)


denis-anisimov:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8291)*

Issue #8257, opened 2017-01-17
AbsoluteLayout in AbsoluteLayout defined width positioning after component size change


caalador:
Vaadin 7.7.6 & Vaadin 8.0.0.beta1

Adding a 250px wide AbsoluteLayout with content into absolute layout with the css position 'right: 0; top: 10px;' positions the component correctly.

But setting the size of the contained AbsoluteLayout to 100% will make the component disappear as it gets a width of 0 for some reason.

Test example:
`
        final AbsoluteLayout layout = new AbsoluteLayout();

        AbsoluteLayout layoutExpading = new AbsoluteLayout();
        layoutExpading.setWidth("250px");
        layoutExpading.addComponent(new Panel(new CssLayout(new Label("Start Width 250px"))));

        Button changeSize = new Button("Change Size",
                e -> {
                    if (layoutExpading.getWidthUnits().equals(Unit.PERCENTAGE)) {
                        layoutExpading.setWidth("250px");
                    } else {
                        layoutExpading.setWidth("100%");
                    }
                }
        );
        layout.addComponent(layoutExpading, "right:0;top:10px;");
        layout.addComponent(changeSize, "left: 10x; top: 50px;");

        setContent(layout);
`

Issue #8256, opened 2017-01-17
AbsoluteLayout defined width component positioning after component size change


caalador:
Vaadin 7.7.6 & Vaadin 8.0.0.beta1

Adding a 250px wide component into absolute layout with the css  'right: 0; top: 10px;' positions the component correctly.

But when resizing the component to be 100% wide the component should stretch from side to sice, but this doesn't happen and the component actually becomes smaller.

`
        final AbsoluteLayout layout = new AbsoluteLayout();

        final Panel full = new Panel(new CssLayout(new Label("Start Width 250px")));
        full.setWidth("250px");

        Button changeSize = new Button("Change Size",
                e -> {
                    if (full.getWidthUnits().equals(Unit.PERCENTAGE)) {
                        full.setWidth("250px");
                    } else {
                        full.setWidth("100%");
                    }
                }
        );
        layout.addComponent(full, "right:0;top:10px;");
        layout.addComponent(changeSize, "left: 10x; top: 50px;");

        setContent(layout);
`

Issue #8255, opened 2017-01-17
AbsoluteLayout fullWidth component positioning after component size change


caalador:
Vaadin 7.7.6  & Vaadin 8.0.0.beta1

Adding a 100% wide component into absolute layout with the css position 'right: 0; top: 10px;' positions the component correctly.

But when resizing the component to be absolute width for instance 250px the expected result is that the component would be positioned on the right side of the view. 

Now the component stays positioned to the left side of the view.

Test example: (Panel used to make the size easier to see as it has borders)
`
        final AbsoluteLayout layout = new AbsoluteLayout();

        final Panel full = new Panel(new CssLayout(new Label("Start Width 100%")));
        full.setWidth("100%");

        Button changeSize = new Button("Change Size",
                e -> {
                    if (full.getWidthUnits().equals(Unit.PERCENTAGE)) {
                        full.setWidth("250px");
                    } else {
                        full.setWidth("100%");
                    }
                }
        );
        layout.addComponent(full, "right:0;top:10px;");
        layout.addComponent(changeSize, "left: 10x; top: 50px;");

        setContent(layout);
`

caalador:
This has most likely to do with the fix for #4778

PR #8304, opened 2017-01-23
Fix absolute layout component sizing on size change in Vaadin7

caalador:
Fixes problems for #8255, #8256, #8257 for Vaadin7

Changing from relative to defined size leaves the component positioned wrong,
also changing from defined size to relative size leaves the component in the
wrong position with the wrong size.

Change makes the wrapper rerun setChildWidgetPosition() that adds/removes
necessary css definitions to/from the cssPosition String.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8304)


denis-anisimov:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8304)*

PR #8286, opened 2017-01-19
Serve VAADIN files also from META-INF/resources

Artur-:
Fixes #8206

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8286)

PR #8320, opened 2017-01-24
Serve VAADIN files also from META-INF/resources (#8286)

### LINKED ISSUE #8206: VAADIN doesn't work in web fragments
For larger web application it's handy to split it into several web fragments. Static resources in web fragments are put into the META-INF/resources folder instead of the root folder. However, there is a bug in VaadinServlet.isAllowedVAADINResourceUrl which makes this impossible:

> Blocked attempt to access a JAR entry not starting with /VAADIN/: jar:file:/blablabla/com.knime.enterprise.war/WEB-INF/lib/webportal-4.4.0.jar!/META-INF/resources/VAADIN/login.html 

The affected method should also allow /META-INF/resources/VAADIN in case it exists.

Artur-:
Fixes #8206 

From 7.7

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8320)

### DIFF
--- a/server/src/main/java/com/vaadin/server/VaadinServlet.java
+++ b/server/src/main/java/com/vaadin/server/VaadinServlet.java
@@ -1140,6 +1140,7 @@ public class VaadinServlet extends HttpServlet implements Constants {
     @Deprecated
     protected boolean isAllowedVAADINResourceUrl(HttpServletRequest request,
             URL resourceUrl) {
+        String resourcePath = resourceUrl.getPath();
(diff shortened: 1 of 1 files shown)


denis-anisimov:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8320)*

knime:
Thanks for fixing this. When will the fix be available in a release?

PR #8332, opened 2017-01-25
Extract Bean validation logic from Binder to BeanValidationBinder

### LINKED ISSUE #8253: Introduce BeanValidationBinder that always uses bean validation
The current design of `Binder` is slightly confusing and fragile since it might or might not add JSR 303 validators for its bindings.

We could make `Binder` never add any validators automatically, and then introduce `BeanValidationBinder` that always adds JSR 303 validators and throws a helpful exception from any constructor if there's no bean validation API or implementation on the classpath.

This would help avoid surprises both in the case when bean validation is expected and when it's not expected.

denis-anisimov:
Fixes #8253

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8332)

### DIFF
--- a/server/src/main/java/com/vaadin/data/BeanBinderPropertySet.java
+++ b/server/src/main/java/com/vaadin/data/BeanBinderPropertySet.java
@@ -30,9 +30,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.vaadin.data.Binder.BindingBuilder;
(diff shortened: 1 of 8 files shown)


elmot:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8332#-:-KbL1xMXwTFasIbAj7sQ:bnfp4nl)*

PR #8327, opened 2017-01-25
Fixes NPE when grid is sorted by column with null values

elmot:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8327)

PR #8344, opened 2017-01-26
Make Grid.addColumn set a comparator from valueProvider where possible

### LINKED ISSUE #8128: Grid column using automatic string conversion should be sorted according to the original value
Now it's sorted by the string value since the value is already converted to a string by the time it reaches the in-memory sorting logic. Should probably be fixed by using a custom renderer that accepts anything and encodes it as a string instead of the current approach.

ahie:
Renderers no longer determine the in memory comparator of grid columns.
Additionally, columns are now typed to the return type of ValueProvider.apply,
instead of the presentation type of its Renderer.

An overload for addColumn has been added for the case where valueProvider
does not return a comparable value.

Closes #8128 and #8327

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8344)


elmot:
:lgtm:

---

Reviewed 15 of 15 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344#-:-KbQDzYRUj7gtX1LrwRq:bnfp4nl)*

denis-anisimov:
Reviewed 13 of 15 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1819 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQEOIK6rwUtW1fKzys:-KbQEOIK6rwUtW1fKzyt:b-bqjf5l) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1819)):*
> ? super V

What does it mean in practice ?
Can't image a situation where it will work......
Should it be just `Comparable<V>` ?

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1872 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFsSLE-PTZxi0fKYE:-KbQFsSLE-PTZxi0fKYF:bu9hlzt) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1872)):*
> ```Java
>             ValueProvider<T, ? extends V> valueProvider,
>             AbstractRenderer<? super T, ? super V> renderer,
>             SerializableComparator<T> comparator) {
> ```

Is it possible to add an overloaded method without the third argument ? (to avoid passing `null` every time)

---

*[server/src/test/java/com/vaadin/tests/server/component/grid/GridTest.java, line 37 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFByRZv6UC3ANlMcU:-KbQFByRZv6UC3ANlMcV:b-ajdbv9) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/test/java/com/vaadin/tests/server/component/grid/GridTest.java#L37)):*
> String.valueOf

Why ?
Shouldn't it work out of the box ?

---

*[uitest/src/main/java/com/vaadin/tests/components/grid/GridAddColumn.java, line 18 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFQCrLCsHNsVW8D_P:-KbQFQCrLCsHNsVW8D_Q:bfunm95) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/uitest/src/main/java/com/vaadin/tests/components/grid/GridAddColumn.java#L18)):*
> String.valueOf(

?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

ahie:

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1819 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQEOIK6rwUtW1fKzys:-KbTnxBNY8FNV8g9Ac06:byyud7p) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1819)):*

> ? super V

What does it mean in practice ?
Can't image a situation where it will work......
Should it be just `Comparable<V>` ?
</blockquote></details>

It is necessary, see for example the type parameters of `Comparator.comparing`.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1872 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFsSLE-PTZxi0fKYE:-KbTnZa4-O2NrRo2wl1W:bkv95i4) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1872)):*

Is it possible to add an overloaded method without the third argument ? (to avoid passing `null` every time)
</blockquote></details>

Due to type erasure you can't enforce a valueProvider that returns a comparable and have an overload that doesn't.

---

*[server/src/test/java/com/vaadin/tests/server/component/grid/GridTest.java, line 37 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFByRZv6UC3ANlMcU:-KbToUx9DfDHK9-wBL0e:b-b0qfny) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/test/java/com/vaadin/tests/server/component/grid/GridTest.java#L37)):*

> String.valueOf

Why ?
Shouldn't it work out of the box ?
</blockquote></details>

It's not comparable. Other option is to use the overload with 3 params.

---

*[uitest/src/main/java/com/vaadin/tests/components/grid/GridAddColumn.java, line 18 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFQCrLCsHNsVW8D_P:-KbTogrI7Ur6sjTKdlvJ:b-macvsj) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/uitest/src/main/java/com/vaadin/tests/components/grid/GridAddColumn.java#L18)):*

> String.valueOf(

?
</blockquote></details>

Same as `GridTest`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

denis-anisimov:

---

*[server/src/test/java/com/vaadin/tests/server/component/grid/GridTest.java, line 37 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbQFByRZv6UC3ANlMcU:-KbTqRYuqxoacyM_MrgE:b-1998a2) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/test/java/com/vaadin/tests/server/component/grid/GridTest.java#L37)):*

It's not comparable. Other option is to use the overload with 3 params.
</blockquote></details>

Should it be comparable ? I thought the idea was to have here some type which doesn't work out of the box.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

Legioth:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTyojHDlchB-4SRaKI:-KbTyojHDlchB-4SRaKJ:bycrjmg) (no related file):*
I'm concerned that this optimizes for the in-memory case while making things overly complicated for backend cases where it doesn't matter whether the column type is comparable.

Prior to this patch, I could do like this with a back end data provider:
```
grid.addColumn(Person::getDepartment).setSortProperty("department");
```

With this patch, I'd instead have to do this:
```
grid.addColumn(Person::getDepartment, new ObjectRenderer(), null).setSortProperty("department");
```

---

*[client/src/main/java/com/vaadin/client/renderers/ObjectRenderer.java, line 26 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTswx7EmqIGIrkqxgD:-KbTswx8x0t4wFH1oY1N:bb7y8h9) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/client/src/main/java/com/vaadin/client/renderers/ObjectRenderer.java#L26)):*
> ```Java
>  * @author Vaadin Ltd
>  */
> public class ObjectRenderer implements Renderer<String> {
> ```

This class shouldn't be necessary. `ObjectRendererConnector.getRenderer()` can instead return the identical `com.vaadin.client.renderers.TextRenderer`.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1000 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTtNOHjEwWr3GDh6Uq:-KbTtNOHjEwWr3GDh6Ur:b5qgs16) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1000)):*
> ```Java
>          * {@link Column#setSortProperty(String...)}.
>          * <p>
>          * Setting a comparator enables client-side sorting by default. To
> ```

What does "by default" refer to. What is the case where the default isn't applied?

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1001 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTt9oddQhMPeoR_Vjt:-KbTt9oddQhMPeoR_Vju:b-iuoz1q) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1001)):*
> ```Java
>          * <p>
>          * Setting a comparator enables client-side sorting by default. To
>          * configure client-side sorting call {@link #setSortable(boolean)}.
> ```

This implies that I'd always also need to call `setSortable` because I'd still somehow need to configure it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

Legioth:

---

*[client/src/main/java/com/vaadin/client/renderers/ObjectRenderer.java, line 21 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbU-1JUFLwM5n5dmJOv:-KbU-1JUFLwM5n5dmJOw:b-skt4a6) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/client/src/main/java/com/vaadin/client/renderers/ObjectRenderer.java#L21)):*
> ```Java
> 
> /**
>  * Renderer that renders the encoded string representations of an object.
> ```

representation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

ahie:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTyojHDlchB-4SRaKI:-KbiMwkJ2WX-lmrt7eQh:b-kma4l) (no related file):*

I'm concerned that this optimizes for the in-memory case while making things overly complicated for backend cases where it doesn't matter whether the column type is comparable.

Prior to this patch, I could do like this with a back end data provider:
```
grid.addColumn(Person::getDepartment).setSortProperty("department");
```

With this patch, I'd instead have to do this:
```
grid.addColumn(Person::getDepartment, new ObjectRenderer(), null).setSortProperty("department");
```

</blockquote></details>

True, but in the case where the object doesn't have a sensible toString implementation neither would work out of the box, and the developer would end up having to do something like
```
grid.addColumn(person -> person.getDepartment().getName()).setSortProperty("department");
```
in either case.

---

*[client/src/main/java/com/vaadin/client/renderers/ObjectRenderer.java, line 26 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTswx7EmqIGIrkqxgD:-KbiL0TdqN0A9oBmrohy:b7heuz) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/client/src/main/java/com/vaadin/client/renderers/ObjectRenderer.java#L26)):*

This class shouldn't be necessary. `ObjectRendererConnector.getRenderer()` can instead return the identical `com.vaadin.client.renderers.TextRenderer`.
</blockquote></details>

True, but originally decided to keep them separate. Now removed.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1000 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTtNOHjEwWr3GDh6Uq:-KbiJRCjJnvYv1aK5TrR:bss4a2b) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1000)):*

What does "by default" refer to. What is the case where the default isn't applied?
</blockquote></details>

Removed.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1001 at r2](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTt9oddQhMPeoR_Vjt:-KbiJS1teBNKcYS61_dz:bss4a2b) ([raw file](https://github.com/vaadin/framework/blob/4f416b0db429d067bb39725128eaded78317bb70/server/src/main/java/com/vaadin/ui/Grid.java#L1001)):*

This implies that I'd always also need to call `setSortable` because I'd still somehow need to configure it.
</blockquote></details>

Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

Legioth:
Review status: 12 of 14 files reviewed at latest revision, 4 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbTyojHDlchB-4SRaKI:-KbinmXcE5UDF0NsqXGu:b-fq7r8f) (no related file):*

True, but in the case where the object doesn't have a sensible toString implementation neither would work out of the box, and the developer would end up having to do something like
```
grid.addColumn(person -> person.getDepartment().getName()).setSortProperty("department");
```
in either case.
</blockquote></details>

Having a sensible `toString` implementation is good in many cases and is a reasonable requirement for being able to benefit from a shorthand.

Implementing `Comparable` is on the other hand not a reasonable requirement, especially not in the case when a back end data provider is used and the `compareTo` implementation will not even be used for anything.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 1000 at r3](https://reviewable.io:443/reviews/vaadin/framework/8344#-KbioXn5MtBT3CoJ5fhm:-KbioXn5MtBT3CoJ5fhn:b-stpz27) ([raw file](https://github.com/vaadin/framework/blob/3b4c4950fa2287b7b2a72df2ac86bf6bf928d89e/server/src/main/java/com/vaadin/ui/Grid.java#L1000)):*
> ```Java
>          * {@link Column#setSortProperty(String...)}.
>          * <p>
>          * Setting a comparator enables client-side sorting.
> ```

...when an in-memory data provider is used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8344)*

Legioth:
Implemented with a different specification in #8458.

Issue #8369, opened 2017-01-30
Make ValueChangeListener constructor parameters parameterized


Legioth:
#8351 added convenience constructors to various components. The problem is that the `ValueChangeListener` parameters are defined without any type parameter, which means that you cannot do something like this:
```
new TextField(event -> {
  System.out.println(event.getValue().toLowerCase());
)};
```
since `event.getValue()` will be defined to return `Object` instead of the expected `String`.

PR #8388, opened 2017-01-31
Patameterize the ValueChangeListeners for rebuilt fields (#8369)

caalador:
Listeners are now defined with the correct type parameter for
each field type so the event.getValue() is what one would expect.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8388)


denis-anisimov:
Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8388)*

PR #8453, opened 2017-02-06
Use V type of a ValueProvider<T, V> for the Column<T, V> class.

### LINKED ISSUE #8371: Make Grid column types based on the value provider instead of the renderer
You can create a grid column with a renderer that accepts a super type of the value provider type, so that e.g. a `ValueProvider<Person, Integer>` can be used with a `NumberRenderer` that is typed to `Number`. The problem is that `addColumn` will in that case return a `Column<Person, Number>` instead of the more specific `Column<Person, Integer>`.

denis-anisimov:
Fixes #8371.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8453)


denis-anisimov:
Covered by #8458

PR #8456, opened 2017-02-06
Fix the absolute layout component resize positioning tests.

caalador:
Fix the problem that older browsers do not remove the 'left' css target
but actually calculate the width instead of setting it as 'AUTO'

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8456)


pleku:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8456)*

PR #8458, opened 2017-02-06
Make one-arg addColumn use natural sort for mutually comparable values

### LINKED ISSUE #8128: Grid column using automatic string conversion should be sorted according to the original value

### LINKED ISSUE #8371: Make Grid column types based on the value provider instead of the renderer

Legioth:
Fixes #8371
Fixes #8128

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8458)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -805,8 +805,8 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
          * @param renderer
          *            the type of value, not <code>null</code>
          */
-        protected Column(ValueProvider<T, ? extends V> valueProvider,
-                Renderer<V> renderer) {
+        protected Column(ValueProvider<T, V> valueProvider,
+                Renderer<? super V> renderer) {
             Objects.requireNonNull(valueProvider,
                     "Value provider can't be null");
             Objects.requireNonNull(renderer, "Renderer can't be null");
@@ -824,12 +824,11 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
             // removed
             addExtension(renderer);
 
-            Class<V> valueType = renderer.getPresentationType();
+            Class<? super V> valueType = renderer.getPresentationType();
 
             if (Comparable.class.isAssignableFrom(valueType)) {
                 comparator = (a, b) -> compareComparables(
                         valueProvider.apply(a), valueProvider.apply(b));
-                state.sortable = true;
             } else if (Number.class.isAssignableFrom(valueType)) {
                 /*
                  * Value type will be Number whenever using NumberRenderer.
@@ -839,10 +838,37 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
                 comparator = (a, b) -> compareNumbers(
                         (Number) valueProvider.apply(a),
                         (Number) valueProvider.apply(b));
-                state.sortable = true;
             } else {
-                state.sortable = false;
+                comparator = (a, b) -> compareMaybeComparables(
+                        valueProvider.apply(a), valueProvider.apply(b));
             }
+        }
+
+        private static int compareMaybeComparables(Object a, Object b) {
+            if (hasCommonComparableBaseType(a, b)) {
+                return compareComparables(a, b);
+            } else {
+                return compareComparables(a.toString(), b.toString());
+            }
+        }
+
+        private static boolean hasCommonComparableBaseType(Object a, Object b) {
+            if (a instanceof Comparable<?> && b instanceof Comparable<?>) {
+                Class<?> aClass = a.getClass();
+                Class<?> bClass = b.getClass();
+
+                if (aClass == bClass) {
+                    return true;
+                }
+
+                Class<?> baseType = ReflectTools.findCommonBaseType(aClass,
+                        bClass);
+                if (!baseType.isAssignableFrom(Comparable.class)) {
+                    return false;
+                }
+            }
+
+            return false;
         }
 
         @SuppressWarnings("unchecked")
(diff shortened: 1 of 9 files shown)


hesara:
Reviewed 9 of 9 files at r1.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 809 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIsNIj-V_WlN9AbLbW:-KcIsNIj-V_WlN9AbLbX:b-488zvo) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/server/src/main/java/com/vaadin/ui/Grid.java#L809)):*
> ```Java
>          */
>         protected Column(ValueProvider<T, V> valueProvider,
>                 Renderer<? super V> renderer) {
> ```

Does this typing change impact old/migrated user code? If so (and not only 8.0 new code), should it be mentioned somewhere?

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 864 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIrnvBkA50YqP6HvAw:-KcIrnvBkA50YqP6HvAx:bn82pue) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/server/src/main/java/com/vaadin/ui/Grid.java#L864)):*
> ```Java
>                 }
> 
>                 Class<?> baseType = ReflectTools.findCommonBaseType(aClass,
> ```

(this only handles comparable base classes, not e.g. the case where both implement an interface that is comparable)

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/ColumnState.java, line 24 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIs7Xyobh80YHOKrn3:-KcIs7Xyobh80YHOKrn4:b-jhaui0) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/shared/src/main/java/com/vaadin/shared/ui/grid/ColumnState.java#L24)):*
> ```Java
>     public String caption;
>     public String internalId;
>     public boolean sortable = true;
> ```

should this be mentioned in release notes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8458)*

Legioth:

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 809 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIsNIj-V_WlN9AbLbW:-KcJT0n2yBXshbdH0yGx:b-n2bul8) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/server/src/main/java/com/vaadin/ui/Grid.java#L809)):*

Does this typing change impact old/migrated user code? If so (and not only 8.0 new code), should it be mentioned somewhere?
</blockquote></details>

I don't see how this particular adjustment would have any significant impact since `Column` in FW 7 doesn't have any type parameters at all. 

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 864 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIrnvBkA50YqP6HvAw:-KcJTmU7j_kCZNXuiHqx:b1vx490) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/server/src/main/java/com/vaadin/ui/Grid.java#L864)):*

(this only handles comparable base classes, not e.g. the case where both implement an interface that is comparable)
</blockquote></details>

Let's fix that case when there's a concrete example of how it happens in real life.

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/ColumnState.java, line 24 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIs7Xyobh80YHOKrn3:-KcJWjeKnzeS0dxrBULe:b-wkzbfx) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/shared/src/main/java/com/vaadin/shared/ui/grid/ColumnState.java#L24)):*

should this be mentioned in release notes?
</blockquote></details>

Seems like all other Grid mentions in the release notes are on a more high level than this kind of detail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8458)*

hesara:

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/ColumnState.java, line 24 at r1](https://reviewable.io:443/reviews/vaadin/framework/8458#-KcIs7Xyobh80YHOKrn3:-KcMWucpYFgFYivSqxjS:b-txd6jf) ([raw file](https://github.com/vaadin/framework/blob/6716baf655be6ccbdc7f903ddfbf7207aeb0e104/shared/src/main/java/com/vaadin/shared/ui/grid/ColumnState.java#L24)):*

Seems like all other Grid mentions in the release notes are on a more high level than this kind of detail.
</blockquote></details>

I'd still consider adding this as the defaults change, but not mandatory

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8458)*

PR #8460, opened 2017-02-06
Provide a way to configure bean binder to auto set required fields

### LINKED ISSUE #8382: Fields mapped to a bean property with @NotNull or @NotEmpty should be visually marked as required
@Legioth commented on [Wed Sep 07 2016](https://github.com/vaadin/framework8-issues/issues/198)

Still needs to consider the exact details since e.g.  @NotNull would accept `""`, even though that is usually not considered acceptable for a required field.

This also relates to https://dev.vaadin.com/ticket/14746

---

@pleku commented on [Tue Oct 11 2016](https://github.com/vaadin/framework8-issues/issues/198#issuecomment-252859996)

SHOULD HAVE, but not necessary for FW8.

---

@Legioth commented on [Wed Nov 02 2016](https://github.com/vaadin/framework8-issues/issues/198#issuecomment-257796652)

#419 detaches the validation logic from the displaying of the standard required indicator. This means that we would only need some way of identifying which validation annotations are such that `field.setRequiredIndicatorVisible(true)` should be called when binding.

This could be done e.g. by configuring `Binder` with a list of annotation types or a predicate that checks each constraint to see if its semantics can be seen as "required." This would include at least `@NotNull` and `@Size(min = 1)` from the standard annotations in JSR-303, and maybe also supplemented with the commonly used `@NotEmpty` constraint defined by Hibernate Validator (though it should be implemented in a way that doesn't require Hibernate Validator classes to be on the classpath).

One open question is whether the configuration would be per `Binder` instance, or if it would be more useful to actually define this in a static fashion so that the same settings would be automatically used in all binders of the applications.

denis-anisimov:
@NotNull, @Size(min>1), @NotEmpty annotations are handled via default
configuration which marks fields as required.

Fixes #8382

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8460)

### DIFF
--- a/server/src/main/java/com/vaadin/data/BeanValidationBinder.java
+++ b/server/src/main/java/com/vaadin/data/BeanValidationBinder.java
@@ -15,6 +15,10 @@
  */
 package com.vaadin.data;
 
+import javax.validation.metadata.BeanDescriptor;
+import javax.validation.metadata.ConstraintDescriptor;
+import javax.validation.metadata.PropertyDescriptor;
+
 import com.vaadin.data.util.BeanUtil;
 import com.vaadin.data.validator.BeanValidator;
 
@@ -28,6 +32,8 @@ import com.vaadin.data.validator.BeanValidator;
 public class BeanValidationBinder<BEAN> extends Binder<BEAN> {
 
     private final Class<BEAN> beanType;
+
+    private RequiredFieldConfigurator requiredConfigurator = RequiredFieldConfigurator.DEFAULT;
 
     /**
      * Creates a new binder that uses reflection based on the provided bean type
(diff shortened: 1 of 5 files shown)


Legioth:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/src/main/java/com/vaadin/data/BeanValidationBinder.java, line 108 at r1](https://reviewable.io:443/reviews/vaadin/framework/8460#-KcM6xDCYdc_Bdf9p664:-KcM6xDCYdc_Bdf9p665:b-1am492) ([raw file](https://github.com/vaadin/framework/blob/8fc41552ec7066d200076ce5d1401e3153d6abad/server/src/main/java/com/vaadin/data/BeanValidationBinder.java#L108)):*
> ```Java
>         PropertyDescriptor propertyDescriptor = descriptor
>                 .getConstraintsForProperty(definition.getName());
>         if (propertyDescriptor == null || requiredConfigurator == null) {
> ```

It seems to me like getConstraintsForClass and getConstraintsForProperty might be implemented to use relatively expensive reflection calls. In that case, it could be good to return immediately if the configurator is null.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8460#-:-KcM7cpNMjbbZnlK84AH:bnfp4nl)*

denis-anisimov:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/src/main/java/com/vaadin/data/BeanValidationBinder.java, line 108 at r1](https://reviewable.io:443/reviews/vaadin/framework/8460#-KcM6xDCYdc_Bdf9p664:-KcMNes6CTLSjcIgreH2:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/8fc41552ec7066d200076ce5d1401e3153d6abad/server/src/main/java/com/vaadin/data/BeanValidationBinder.java#L108)):*

It seems to me like getConstraintsForClass and getConstraintsForProperty might be implemented to use relatively expensive reflection calls. In that case, it could be good to return immediately if the configurator is null.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8460)*

Legioth:
Reviewed 4 of 5 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8460)*

PR #8495, opened 2017-02-07
Wrong Action may be executed when for detached component (#5864)

mcollovati:
Due to action key generation on ActionManager a wrong action
may be executed if the component that fired the wanted action
is already detached.

This fix was originally pushed on gerrit and partially reviewed
https://dev.vaadin.com/review/#/c/13229/

I hope this could still be useful

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8495)


pleku:
Reviewed 4 of 4 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdZaYkhKgNvCpb8EIgk:-KdZaYkhKgNvCpb8EIgl:bwnw4i7) (no related file):*
:lgtm: , I have commented only on couple clarifying things. Trying to get someone else to take a look too, preferably familiar with ActionManager

---

*[server/src/main/java/com/vaadin/event/ActionManager.java, line 183 at r1](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdUeszbvOo7J9w87iB7:-KdUeszbvOo7J9w87iB8:b7aiqth) ([raw file](https://github.com/vaadin/framework/blob/d95f18b3b282ece51e977dc240d6f9b00979d5b6/server/src/main/java/com/vaadin/event/ActionManager.java#L183)):*
> final Action a

Could be renamed to _ action_  just to avoid use of single letter variables not describing what they really do

---

*[server/src/main/java/com/vaadin/event/ActionManager.java, line 262 at r1](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdUfUShtFZfv8TTFfRP:-KdUfUShtFZfv8TTFfRQ:bdfku6l) ([raw file](https://github.com/vaadin/framework/blob/d95f18b3b282ece51e977dc240d6f9b00979d5b6/server/src/main/java/com/vaadin/event/ActionManager.java#L262)):*
> Action.Handler h

To avoid single letter variables, could be renamed to _handler_

---

*[server/src/main/java/com/vaadin/event/ActionManager.java, line 268 at r1](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdUfYy6HzWGjIuG8LmS:-KdUfYy6HzWGjIuG8LmT:bdacnqg) ([raw file](https://github.com/vaadin/framework/blob/d95f18b3b282ece51e977dc240d6f9b00979d5b6/server/src/main/java/com/vaadin/event/ActionManager.java#L268)):*
> Action[] as

To avoid single letter variables, could be renamed to _actions_

---

*[server/src/main/java/com/vaadin/event/ActionManager.java, line 363 at r1](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdZa8D4kJ-cJOvrQeNq:-KdZa8D4kJ-cJOvrQeNr:bqd5c22) ([raw file](https://github.com/vaadin/framework/blob/d95f18b3b282ece51e977dc240d6f9b00979d5b6/server/src/main/java/com/vaadin/event/ActionManager.java#L363)):*
> ```Java
> 
>         private final Set<String> prefixes = new HashSet<String>();
>         private final static Pattern PREFIXED_KEY = Pattern.compile("^([^_]+)_(.*)$");
> ```

Could add a brief comment for this, an example on what type of pattern it expects

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8495)*

hesara:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdZaYkhKgNvCpb8EIgk:-Kd_HRzApuWPlDo3QBUT:bq3637b) (no related file):*

:lgtm: , I have commented only on couple clarifying things. Trying to get someone else to take a look too, preferably familiar with ActionManager
</blockquote></details>

My main concern with the patch is the additional complexity it introduces, and I wonder if the whole problem could be solved in a simpler way e.g. by just ensuring that the action key mappers for different connectors do not use keys that overlap with previously used ones from other ones? The keys are strings throughout the system.

Would it be enough have a more customized ActionKeyMapper that always prefixes connector related keys with the connector id? Note that ActionManager.getActionSet() is private, and could be modified to return a LinkedHashMap or a List of pairs that includes the connector id information. This would probably eliminate the need for wrapping and unwrapping, and ActionManager.handleActions() could then simply look up the keys in the action map as it does now.

I might be missing something essential as I didn't try to make these changes myself, though.

---

*[server/src/main/java/com/vaadin/event/ActionManager.java, line 383 at r1](https://reviewable.io:443/reviews/vaadin/framework/8495#-Kd_66Bgl8F3KF9vwq1x:-Kd_66BhCSJJLGOKfLpv:b-7m5o7u) ([raw file](https://github.com/vaadin/framework/blob/d95f18b3b282ece51e977dc240d6f9b00979d5b6/server/src/main/java/com/vaadin/event/ActionManager.java#L383)):*
> ```Java
>                 Matcher matcher = PREFIXED_KEY.matcher(key);
>                 if (matcher.matches() && prefixes.contains(matcher.group(1))) {
>                     return super.get(matcher.group(2));
> ```

If keeping this approach, could this branch and the fallback below (when any of the conditions is not fulfilled) still lead to conflicts in some cases? Should this method always return null(?) if matcher.matches() but the prefix is incorrect?

---

*[uitest/src/com/vaadin/tests/actions/ActionsOnDetachedComponentsTest.java, line 72 at r1](https://reviewable.io:443/reviews/vaadin/framework/8495#-Kd_0XLrWUSMRguNwCSO:-Kd_0XLshvyXSE3LiA9m:bi2daj4) ([raw file](https://github.com/vaadin/framework/blob/d95f18b3b282ece51e977dc240d6f9b00979d5b6/uitest/src/com/vaadin/tests/actions/ActionsOnDetachedComponentsTest.java#L72)):*
> ```Java
>     public void actionOnDetachedComponentShouldNotBeHandled() throws InterruptedException {
>         TableElement table = $(TableElement.class).first();
>         table.getRow(0).contextClick();
> ```

there might be some issues with the current Firefox driver and selenium version and the various click events, we need to run the test in our testing infra to check

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8495)*

mcollovati:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdZaYkhKgNvCpb8EIgk:-Ke-q5xIn9n6BlWHArBn:b-tjmhjs) (no related file):*

My main concern with the patch is the additional complexity it introduces, and I wonder if the whole problem could be solved in a simpler way e.g. by just ensuring that the action key mappers for different connectors do not use keys that overlap with previously used ones from other ones? The keys are strings throughout the system.

Would it be enough have a more customized ActionKeyMapper that always prefixes connector related keys with the connector id? Note that ActionManager.getActionSet() is private, and could be modified to return a LinkedHashMap or a List of pairs that includes the connector id information. This would probably eliminate the need for wrapping and unwrapping, and ActionManager.handleActions() could then simply look up the keys in the action map as it does now.

I might be missing something essential as I didn't try to make these changes myself, though.
</blockquote></details>

I agree with you, this solution introduces complexity.
I'll try to work on a solution based on  your suggestions about using a different return type for ActionManager.getActionSet().
I think it would be better return a list of pairs instead of a map to avoid potential problems with equals/hashcode implementations.
So, if I understand correctly, ActionKeyMapper could take the pair as type parameter and generate the prefixed key without wrapping and unwrapping the action.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8495)*

mcollovati:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8495#-KdZaYkhKgNvCpb8EIgk:-KeTYtiWI8TeD6ML9mdP:b-qjyc72) (no related file):*
<details><summary><i>Previously, mcollovati (Marco Collovati) wrote…</i></summary><blockquote>

I agree with you, this solution introduces complexity.
I'll try to work on a solution based on  your suggestions about using a different return type for ActionManager.getActionSet().
I think it would be better return a list of pairs instead of a map to avoid potential problems with equals/hashcode implementations.
So, if I understand correctly, ActionKeyMapper could take the pair as type parameter and generate the prefixed key without wrapping and unwrapping the action.
</blockquote></details>

I took a better look into this issue and found that there could be conflicts
also on `ActionManager` own actions in some cases (ie an action that removes itself or
changes the order of other actions).

A way to resolve conflicts could be to have all actions
managed by an `ActionManager` identified by some kind of stable identifier
that is immutable between subsequent request.

An option could be to maintain a Map<Action, String> member (or WeakHashMap),
where String is the action identifier that is looked up/filled/cleaned in `getActionSet` method; 
the method will then return the Action/prefix pairs that would then
be used by `ActionKeyMapper`.

The downside of this approach is that the `ActionManager` should keep
a reference to all Action instances, self ones and from other action `Handler`s.

Do you think this could lead to other problems? 
Is it worth to try this kind of implementation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8495)*

hesara:
I'm taking a look if I can make a simpler version of this fix.

hesara:
I tried a much simpler approach in #9806 - @mcollovati do you have any comments about that approach?

mcollovati:
@hesara Your approach is much better and cleaner :+1: 
Just a question: may `ActionKeyMapper.lastKey` be declared as volatile or it doesn't matter?

hesara:
To be safe across sessions, addressed the issue with Atomicinteger, which also does not need an explicit "volatile" declaration.

Closing this PR as superceded by #9806.

PR #8523, opened 2017-02-09
Delegate Checkbox caption handling to parent

mazerty:
I don't know why it had been set to false... Is there any particular reason ?

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8523)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/checkbox/CheckBoxConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/checkbox/CheckBoxConnector.java
@@ -48,7 +48,7 @@ public class CheckBoxConnector extends AbstractFieldConnector
 
     @Override
     public boolean delegateCaptionHandling() {
-        return false;
+        return true;
     }
 
     @Override


hesara:
The rendering of the caption on the right of the check box is intentional and customary, and letting the generic layout logic handle captions would break that and usually place the captions above the check box. Thus, closing this as won't fix.

Note that with CSS, it is still possible (although somewhat complicated) to affect the layout location of the caption of a CheckBox.

mazerty:
The problem is with FormLayout, then it doesn't align with the other labels.
People have made the FormCheckbox plugin to compensate, still I think the parent should be one who handle the Checkbox differently, rather than the opposite, you see what I'm saying ?

Legioth:
The problem here is that the "caption" of some components, most noticeably `Button` and `CheckBox` is overloaded to be understood as the component's own internal label instead of the label that the parent layout will show for it.

This is in retrospect a bad design decision – those are really two completely separate concepts. Sometimes it's better to have the caption managed by the layout (e.g. to make labels align in `FormLayout`) and in other cases it would be really strange (e.g. most cases with a `VerticalLayout` or `HorizontalLayout`).

Simply changing the value of that one boolean value would have severe implications on backwards compatibility. What could be considered is instead to either make the boolean setting configurable or introducing two separate values with some kind of fallback logic so that already implemented UIs wouldn't suddenly change. Either approach would remove the need for workarounds like `FormCheckbox`.

mazerty:
Thank you for elaborating, I see it goes far beyond the quick fix I offered ^^
I'm afraid I'm not familiar enough with custom components yet to undertake such a rework :(
Is it possible that it would be included in a future release ?

PR #8557, opened 2017-02-14
Use TestBench version 5.0.0.beta5

hesara:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8557)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -17,7 +17,7 @@
 
     <properties>
         <vaadin.spring.version>2.0-SNAPSHOT</vaadin.spring.version>
-        <vaadin.testbench.version>5.0.0.alpha2</vaadin.testbench.version>
+        <vaadin.testbench.version>5.0.0.beta5</vaadin.testbench.version>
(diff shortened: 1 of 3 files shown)


tsuoanttila:
Reviewed 3 of 3 files at r1.

---

*[testbench-api/pom.xml, line 18 at r1](https://reviewable.io:443/reviews/vaadin/framework/8557#-Kcvvt8ntrNFrX1ukZBZ:-Kcvvt8ntrNFrX1ukZB_:bv60ary) ([raw file](https://github.com/vaadin/framework/blob/3eb39270eada791e2a5b8262016fbca1eb99cf23/testbench-api/pom.xml#L18)):*
> ```XML
> 
>     <properties>
>         <testbench.core.version>5.0.0.beta5</testbench.core.version>
> ```

Is this version number used in something else than the version definition? I think it should be inherited from the vaadin-root pom dependency management too.

We can do any clean up that comes from this in a separate patch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8557)*

PR #8568, opened 2017-02-14
Update copyright year to 2017

elmot:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8568)


hesara:
I have more and more the impression that this has to be done with the script scripts/updateCopyrightYear.sh to make sure it can be done easily and exactly consistently in all the pending review branches etc. Also tb-api needs to the same changes.

hesara:
Will use the script to do this for all relevant branches.

Issue #8220, opened 2017-01-11
Support for sub-navigators (navigator hierarchy)


heruan:
Since https://github.com/vaadin/framework/issues/2930 was automatically closed I'd like to reopen it with a simple use case, common to many SPA frameworks (see for example [Aurelia](http://aurelia.io)'s [Router](http://aurelia.io/hub.html#/doc/api/aurelia/router/1.1.0/overview)).

Say I have views `Foo` and `Bar` and a menu to navigate them: I instruct the `Navigator` to attach them to a certain `ComponentContainer`, say "Container A". Then, `Bar` view has some content and a sub-menu which should open sub-views in a sub-container, say "Container B".

![vaadin navigator hierarchy](https://cloud.githubusercontent.com/assets/851745/21851592/09371c00-d810-11e6-8d64-08651d6ae761.png)

This needs a second `Navigator`, child of the root one, such that a fragment like `bar/qux` will instruct the root `Navigator` to attach view `Bar` to "Container A" and then will trigger the child `Navigator` of `Bar` to attach view `Qux` to "Container B".

hesara:
This actually doesn't necessarily need a second Navigator, but just a tiny bit of Navigator/ViewProvider/ViewDisplay magic.

A while back I wrote a very very quick prototype that does something similar to demonstrate related ideas - see https://gist.github.com/hesara/9d58c9030ee83a0c376867907bb7d420 .

heruan:
Thank you @hesara for the example. I understand a single sublevel could be implemented with some magic, but this would mean all the hierarchy must be known at the root level, while having sub-navigator may open a wide range of scenarios such as composable hierarchies (e.g. modular applications).

The idea is that a child navigator doesn't need to know at which level of the hierarchy it's attached, it just managed the fragment part it receives and resolves the view with its provider and attaches to its container.

TatuLund:
This feature has been implemented in Vaadin 10+ Router concept. Doing the same in Vaadin 8 is not feasible without introducing similar incompatibility step, hence Navigator will not be rewritten.

PR #8663, opened 2017-02-22
Allow powerful Navigator subclassing

heruan:
`protected` methods in `Navigator` permit to create new `Navigator`s to add new capabilities to Vaadin, but `Navigator` fields are still `private` and this limit, for example, the overriding of `Navigator.init`. This PR makes `protected` the `Navigator` fields.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8663)

### DIFF
--- a/server/src/main/java/com/vaadin/navigator/Navigator.java
+++ b/server/src/main/java/com/vaadin/navigator/Navigator.java
@@ -364,14 +364,14 @@ public class Navigator implements Serializable {
         }
     }
 
-    private UI ui;
-    private NavigationStateManager stateManager;
-    private ViewDisplay display;
-    private View currentView = null;
-    private List<ViewChangeListener> listeners = new LinkedList<>();
-    private List<ViewProvider> providers = new LinkedList<>();
-    private String currentNavigationState = null;
-    private ViewProvider errorProvider;
+    protected UI ui;
+    protected NavigationStateManager stateManager;
+    protected ViewDisplay display;
+    protected View currentView = null;
+    protected List<ViewChangeListener> listeners = new LinkedList<>();
+    protected List<ViewProvider> providers = new LinkedList<>();
+    protected String currentNavigationState = null;
+    protected ViewProvider errorProvider;
 
     /**
      * Creates a navigator that is tracking the active view using URI fragments
@@ -997,7 +997,7 @@ public class Navigator implements Serializable {
      *            state string
      * @return suitable provider
      */
-    private ViewProvider getViewProvider(String state) {
+    protected ViewProvider getViewProvider(String state) {
         String longestViewName = null;
         ViewProvider longestViewNameProvider = null;
         for (ViewProvider provider : providers) {
--- a/server/src/main/java/com/vaadin/navigator/Navigator.java
+++ b/server/src/main/java/com/vaadin/navigator/Navigator.java
@@ -364,14 +364,27 @@ public class Navigator implements Serializable {
         }
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


pleku:
Reviewed 1 of 1 files at r1.

---

*[server/src/main/java/com/vaadin/navigator/Navigator.java, line 374 at r1](https://reviewable.io:443/reviews/vaadin/framework/8663#-Kddr9NUaiPByN0kTtAN:-Kddr9NUaiPByN0kTtAO:bs3oywi) ([raw file](https://github.com/vaadin/framework/blob/8b8c0d75035c5cc30273187ce4326ea3259e5d0b/server/src/main/java/com/vaadin/navigator/Navigator.java#L374)):*
> ```Java
>     protected List<ViewProvider> providers = new LinkedList<>();
>     protected String currentNavigationState = null;
>     protected ViewProvider errorProvider;
> ```

Making these protected makes it public API, meaning there should be also javadocs... (I know that is not the case everywhere in the FW, but we do try to..)

---

*[server/src/main/java/com/vaadin/navigator/Navigator.java, line 504 at r1](https://reviewable.io:443/reviews/vaadin/framework/8663#-KddqgSRVxumiNA9T1dS:-KddqgSS_FsmdZmyCQUQ:bgj5nqw) ([raw file](https://github.com/vaadin/framework/blob/8b8c0d75035c5cc30273187ce4326ea3259e5d0b/server/src/main/java/com/vaadin/navigator/Navigator.java#L504)):*
> ```Java
>      */
>     protected void init(UI ui, NavigationStateManager stateManager,
>             ViewDisplay display) {
> ```

Can you please elaborate why `Navigator.init` requires overriding so that you cannot call `super.init(...)` in your subclass? It just sets the instances for `UI`, `NavigationStateManager` and `ViewDisplay`.

Just want to understand so that can later on avoid the same limitation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8663)*

heruan:
Thank you @pleku for the interest on this. I'm working on a hierarchical `ViewNavigator` which will be able to manage sub-navigation on a specific view (see #8220).

This navigator will extend `Navigator`, but it shouldn't be bound to the whole UI and will have its own state manager and display; problem is UI, state manager and display are all `private` and coupled in the `init` method (i.e. I need to call `super.init` to set the state manager and the display, but that would also set the UI and the navigator instance on it).

pleku:

---

*[server/src/main/java/com/vaadin/navigator/Navigator.java, line 374 at r1](https://reviewable.io:443/reviews/vaadin/framework/8663#-KdytOkNDNbLTu8VEZBX:-KdytOkNDNbLTu8VEZBY:b-yo2h5w) ([raw file](https://github.com/vaadin/framework/blob/8b8c0d75035c5cc30273187ce4326ea3259e5d0b/server/src/main/java/com/vaadin/navigator/Navigator.java#L374)):*
> ```Java
>     protected List<ViewProvider> providers = new LinkedList<>();
>     protected String currentNavigationState = null;
>     protected ViewProvider errorProvider;
> ```

Changing these member fields to `protected` would require them to have javadocs as they are then part of the public API.

I know that is not the case everywhere in the FW, but trying to do this at least with any new stuff.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8663)*

pleku:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8663)*

PR #8677, opened 2017-02-24
Easy DevelopmentServerLauncher run with maven

elmot:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8677)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,18 @@ The following preferences need to be set to keep the project consistent. You nee
 
 ## Running a Development Server
 
-1. Open *Maven Settings* dialog by clicking a button in *Maven Projects* window 
-1. Expand *Maven* in left-hand settings tree
-1. Select *Runner* and set *VM options* to <code>-ea</code>
-1. Click *OK*
-1. In *Maven Projects* window select and run *vaadin-uitest* -> *Plugins* -> *exec* -> *exec:java* target
-1. Open URL *http://localhost:8888/run/&lt;testUI&gt;*
+1. Open *Run* menu  and click *Edit Configurations*
+1. Click green ***+*** sign at top left corner, select *Maven* from popup 
+1. In the run configuration page, set any name for the configuration, select *vaadin-uitest* project folder as *Working directory*
+1. Type <code>exec:exec@run-development-server</code> into *Command line* and save the configuration
+1. Run the configuration and open URL *http://localhost:8888/run/&lt;testUI&gt;*
+
+## Running a Development Server in a debug mode
+
+1. Repeat steps 1..3 from above 
+1. Click green ***+*** sign at top left corner, select *Maven* from popup 
+1. In the run configuration page, set any name for the configuration, select *vaadin-uitest* project folder as *Working directory*
+1. Type <code>exec:exec@debug-development-server</code> into *Command line* and save the configuration
+1. In the same dialog, create new "Remote" debug configuration, using *localhost* and *Port 5005* 
+1. Start both configurations and open URL *http://localhost:8888/run/&lt;testUI&gt;*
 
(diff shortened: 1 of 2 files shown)


hesara:
Reviewed 1 of 2 files at r1.

---

*[uitest/pom.xml, line 359 at r1](https://reviewable.io:443/reviews/vaadin/framework/8677#-Kdyn8yI6ZqgtES6tUbk:-Kdyn8yI6ZqgtES6tUbl:bp7uz1t) ([raw file](https://github.com/vaadin/framework/blob/25558896b791373c8001235ce262af974ede1292/uitest/pom.xml#L359)):*
> ```XML
>                 </configuration>
>             </plugin>
>             <plugin>
> ```

I have two concerns with this:
1) that we might want to add some other exec:java goal later (for whatever purpose) at which point it would be necessary to limit which one is executed in each case, but that can be done with profiles, making the above instructions just a little bit more complicated. 
2) exec:java runs the process in the same VM, which might potentially cause some complications with classpath conflicts - we just had such in the Maven plug-in with the migration tool, and I think it is not at all unlikely that the classpaths of DevelopmentServerLauncher and Maven would conflict. The goal exec:exec would be safer as it runs the program in a separate JVM, and it would probably also allow configuring -ea here rather than in the build conf for Maven itself.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8677)*

elmot:

---

*[uitest/pom.xml, line 359 at r1](https://reviewable.io:443/reviews/vaadin/framework/8677#-Kdyn8yI6ZqgtES6tUbk:-Kdyq6D8GnLiR3yYlDrl:bmr7vae) ([raw file](https://github.com/vaadin/framework/blob/25558896b791373c8001235ce262af974ede1292/uitest/pom.xml#L359)):*

I have two concerns with this:
1) that we might want to add some other exec:java goal later (for whatever purpose) at which point it would be necessary to limit which one is executed in each case, but that can be done with profiles, making the above instructions just a little bit more complicated. 
2) exec:java runs the process in the same VM, which might potentially cause some complications with classpath conflicts - we just had such in the Maven plug-in with the migration tool, and I think it is not at all unlikely that the classpaths of DevelopmentServerLauncher and Maven would conflict. The goal exec:exec would be safer as it runs the program in a separate JVM, and it would probably also allow configuring -ea here rather than in the build conf for Maven itself.
</blockquote></details>

1. There is one more possibility -  starting from maven 3.3.1, one could call explicitly specified execution id, in form of `mvn exec:java@run-development-server`. See [MNG-5768](https://issues.apache.org/jira/browse/MNG-5768), so I don't think potential conflict is an obstacle.
2. Good point, I will update this, and also I will add another `execution` to run the dev server under debugger.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8677)*

elmot:

---

*[uitest/pom.xml, line 359 at r1](https://reviewable.io:443/reviews/vaadin/framework/8677#-Kdyn8yI6ZqgtES6tUbk:-KdzNtmc-YcKiiee9YzL:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/25558896b791373c8001235ce262af974ede1292/uitest/pom.xml#L359)):*

1. There is one more possibility -  starting from maven 3.3.1, one could call explicitly specified execution id, in form of `mvn exec:java@run-development-server`. See [MNG-5768](https://issues.apache.org/jira/browse/MNG-5768), so I don't think potential conflict is an obstacle.
2. Good point, I will update this, and also I will add another `execution` to run the dev server under debugger.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8677)*

hesara:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8677#-:-Ke2gU33sPP2pePBENcN:bnfp4nl)*

PR #8691, opened 2017-02-25
Fix partial state change handling of JavaScript components

kisstkondoros:
In ace0e324b69753431dcde9949eaa9b0e3e648db9 (Use diffstate for JS connectors) 
partial state change handling for JavaScript components was introduced, but the
setNativeState method in JavaScriptConnectorHelper was not adjusted.
By removing the cleanup code from the updateNativeState method it can 
be ensured that the non-changed properties are still present on the state object.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8691)


kisstkondoros:
Null values might be problematic 😞

Legioth:
Thank you for you fix! I've added an automatic test and retargeted your PR for the `master` branch in #8695. We'll pick the fix from `master` to `8.0` once it has been merged there.

PR #8695, opened 2017-02-27
Fix js state partial updates

Legioth:
#7853 updated JavaScript connectors to only send shared state changes instead of sending the full state for every change. That patch failed to remove some client-side logic that assumed that each state change always contains all properties.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8695)


pleku:
Reviewed 1 of 1 files at r1, 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8695)*

PR #8764, opened 2017-03-07
Add license headers to all Java files

asashour:
Add license headers to all Java files

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8764)


hesara:
I'll put this change on hold pending #8774 and related changes to where we automatically check copyright headers. We might need to address also other branches at the same time.

elmot:
Review status: 0 of 1462 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8764#-KfAweShfI2Rz2vcBkPu:-KfAweSqH191TrBiEC3q:bq23eel) (no related file):*
After some discussion we concluded we do not care about license headers in test sources, because we do not ship those sources in any binary distributions. In the same time, accepting of this change bring us and our contributors a lot of rebases and unnecessary work to keep in sync different parts of the framework, without giving real value to our community.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8764)*

asashour:
I would argue about not having headers on the test sources, since other open source projects mandate them.

Anyhow, it is the team decision.

PR #8777, opened 2017-03-08
Include charset in text/html responses

### LINKED ISSUE #8775: HTTP response header is missing charset from Content-Type
Tested in 7.7.7 and 8.0.1

The HTTP response returned by Vaadin Framework contains the Content-Type header, whose content is `text/html`. However, there should also be the character set defined, so the Content-Type value should be `text/html;charset=utf-8`. 

The `<meta>` tag in the response content (HTML headers) has the correct charset defined in Content-Type:
```html
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
```

Artur-:
Fixes #8775

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8777)

### DIFF
--- a/server/src/main/java/com/vaadin/server/BootstrapHandler.java
+++ b/server/src/main/java/com/vaadin/server/BootstrapHandler.java
@@ -352,7 +352,8 @@ public abstract class BootstrapHandler extends SynchronizedRequestHandler {
 
     private void writeBootstrapPage(VaadinResponse response, String html)
             throws IOException {
-        response.setContentType("text/html");
+        response.setContentType(
+                ApplicationConstants.CONTENT_TYPE_TEXT_HTML_UTF_8);
         BufferedWriter writer = new BufferedWriter(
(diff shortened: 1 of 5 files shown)


elmot:
:lgtm:

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8777#-:-Kei8IhzWejhwYBP-8No:bnfp4nl)*

elmot:
To be ported to 8 as well.

Artur-:
What am I supposed to do with this? Why is it "waiting for author"?

hesara:
Some other changes were also merged to 7.7 and need to be picked to master - mostly ones which should be easy to pick like this one. Thus, I don't think this needs to wait except for update to latest 7.7 (please enable "project maintainers can update branch" as not sure of integration order of different patches to 7.7).

PR #8786, opened 2017-03-08
Added helper method to get the amount of rows from the Grid

### LINKED ISSUE #8679: Grid should have a function to get the total number of rows
Grid (Vaadin 8) has no function to ask for the complete number of data rows.
We need such a function to visualize the row count of the Grid.
Optional Filters should be taken into account.
see https://vaadin.com/forum#!/thread/15252122

mstahv:
closes #8679

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8786)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -85,7 +85,6 @@ import com.vaadin.shared.ui.grid.GridConstants;
 import com.vaadin.shared.ui.grid.GridConstants.Section;
 import com.vaadin.shared.ui.grid.GridServerRpc;
 import com.vaadin.shared.ui.grid.GridState;
-import com.vaadin.shared.ui.grid.GridStaticCellType;
 import com.vaadin.shared.ui.grid.HeightMode;
 import com.vaadin.shared.ui.grid.ScrollDestination;
 import com.vaadin.shared.ui.grid.SectionState;
@@ -101,7 +100,6 @@ import com.vaadin.ui.components.grid.FooterRow;
 import com.vaadin.ui.components.grid.GridSelectionModel;
 import com.vaadin.ui.components.grid.Header;
 import com.vaadin.ui.components.grid.Header.Row;
-import com.vaadin.ui.components.grid.HeaderCell;
 import com.vaadin.ui.components.grid.HeaderRow;
 import com.vaadin.ui.components.grid.ItemClickListener;
 import com.vaadin.ui.components.grid.MultiSelectionModel;
@@ -3415,6 +3413,13 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
      */
     public void scrollToEnd() {
         getRpcProxy(GridClientRpc.class).scrollToEnd();
+    }
+
+    /**
+     * @return the number of rows in this Grid
+     */
+    public int size() {
+        return getDataProvider().size(new Query<>());
     }
 
     @Override
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -3428,6 +3428,7 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
      *
      * @return the size of grids data
      * @see DataProvider#size(Query)
+     * @since
      */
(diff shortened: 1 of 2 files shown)


hesara:
Reviewed 1 of 2 files at r1.

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 3419 at r1](https://reviewable.io:443/reviews/vaadin/framework/8786#-Kem1AdCmpjyfs5b-y7I:-Kem1AdCmpjyfs5b-y7J:by9m21j) ([raw file](https://github.com/vaadin/framework/blob/cd476ae8ebf8fbdd40f2701d3f887339a8c6e518/server/src/main/java/com/vaadin/ui/Grid.java#L3419)):*
> ```Java
> 
>     /**
>      * @return the amount of rows in this Grid
> ```

number of rows
(I would have fixed this myself, but don't seem to have push access to the PR branch)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8786)*

mstahv:

---

*[server/src/main/java/com/vaadin/ui/Grid.java, line 3419 at r1](https://reviewable.io:443/reviews/vaadin/framework/8786#-Kem1AdCmpjyfs5b-y7I:-Kem9I1HTBh4LPLEfttV:b-ha8u6c) ([raw file](https://github.com/vaadin/framework/blob/cd476ae8ebf8fbdd40f2701d3f887339a8c6e518/server/src/main/java/com/vaadin/ui/Grid.java#L3419)):*

number of rows
(I would have fixed this myself, but don't seem to have push access to the PR branch)
</blockquote></details>

Don't know why. I'm pretty sure the checkbox was on when I did the PR. I checked that on now (after the PR was done).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8786)*

pleku:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8786#-KemSBZyyjNpacT9N408:-KemSB_2_iAG0xEOcpTU:b-nvumu3) (no related file):*
I'm not even sure if we want to add this method, especially like this as it can have some performance indications and requires lot of explaining.

The simplified version could be to cache the latest reported size in data communicator, and then delegate the added `size()` method to fetch that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8786)*

mstahv:
The need to get the number of rows is so common that the method for sure is needed. I personally don't care at all how it is implemented. I know nothing about Grid's internals (and don't want to know at this point), so I just copied the implementation from the Grid class from an other place where the number of rows was needed. If the size query is slow in the backend, developers probably need to cache that somehow anyways. If this could be optimised at Grids level, it could be done later.

pleku:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8786#-KemSBZyyjNpacT9N408:-KemfMy-50Qc_r8O867f:b-3t7kdr) (no related file):*

I'm not even sure if we want to add this method, especially like this as it can have some performance indications and requires lot of explaining.

The simplified version could be to cache the latest reported size in data communicator, and then delegate the added `size()` method to fetch that.
</blockquote></details>

Yeah my concern is just that with the current implementation, it is not the number of rows in grid, but the number of rows in the used backend. The grid might actually have different amount of rows visible if the data set has changed and grid hasn't been updated. So for the use case, _visualize number of rows in grid_, this is not a valid solution.

I might push another patch for this that used as cached value from data provider, thus it would be lot simpler.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8786)*

mstahv:
See this as well (I don't know how Grid works, but you might want to use the cached value here as well): 
https://github.com/mstahv/framework/blob/e86c18bbe155d6bb7381feb6b0873441c44a8a7b/server/src/main/java/com/vaadin/ui/Grid.java#L3397

pleku:
Thanks, I'll fix that too

pleku:
Closing this PR because it has multiple issues, described in the ticket #8679, and this approach just won't work.

Issue #8656, opened 2017-02-22
Grid cell value using value-providers


heruan:
Given the awesome enhancements in FW8 `Grid`, I was surprised to see missing a quite common use case about adding columns:

```java
ComboBox<Group> groupBox = new ComboBox<>();
groupBox.setItemCaptionGenerator(Group::getName);
groupBox.setItems(groups);

Grid<User> grid = new Grid<>();
grid.addColumn(User::getGroup)
// .setValueGenerator(Group::getName) --> this is missing, value will be Group.toString()!
.setEditorComponent(groupBox, User::setGroup);
```

I really hope the next release will support this!

ahie:
Unless I'm misunderstanding what you want to achieve, you could do this with the following lambda:

```
grid.addColumn(user -> user.getGroup().getName());
```

heruan:
Thank you @ahie, I know that but then I cannot do this:

```java
.setEditorComponent(groupBox, User::setGroup);
```

since the type of column would be `String` and not `Group`.

ahie:
Oh yes of course, my bad. The simplest way around the lack of this feature would be overriding `toString`. If that is not possible, a custom `Renderer` implementation would also do the trick.

heruan:
I'm overriding `toString()` right now, but I'd really like to see this in the next release 😉

heruan:
I've implemented this `Renderer`

```java
public class ValueProviderRenderer<T> extends TextRenderer {

    private ValueProvider<T, String> stringProvider;

    public ValueProviderRenderer(ValueProvider<T, String> stringProvider) {
        this(stringProvider, "");
    }

    public ValueProviderRenderer(ValueProvider<T, String> stringProvider, String nullRepresentation) {
        super(nullRepresentation);
        this.stringProvider = stringProvider;
    }

    @Override
    @SuppressWarnings("unchecked")
    public JsonValue encode(Object value) {
        if (value == null) {
            return super.encode(null);
        } else {
            return Json.create(this.stringProvider.apply((T) value));
        }
    }

}
```

that can be used like this

```java
ComboBox<Group> groupBox = new ComboBox<>();
groupBox.setItemCaptionGenerator(Group::getName);
groupBox.setItems(groups);

Grid<User> grid = new Grid<>();
grid.addColumn(User::getGroup, new ValueProviderRenderer<>(Group::getName))
.setEditorComponent(groupBox, User::setGroup);
```

Legioth:
[...]

Issue #8625, opened 2017-02-20
Components in Grid


pleku:
It should be possible to have _any_ components in Grid's body's cells, just like with Table component.

This is an epic split into several tickets for the actual implementation.

heruan:
Is there a branch available to follow, contribute and test the development of this?

pleku:
@heruan the work for this is planned to start next week and you can follow e.g. the subtickets currently (#8622, #8623). Will include it in an 8.1.0. alpha release as soon as possible.

heruan:
Thank you @pleku for the info. I've just issued PR #8806 proposing a column feature which may work in synergy with this.

pleku:
All subtickets implemented and pending release in 8.1.0.alpha1

PR #8806, opened 2017-03-10
Add columns with a presentation value provider

heruan:
This allows decoupling the column *semantic* type from the *presentation* type. For example:

```java
public class Group {
    private String name;
    private Double privilegeLevel;
    /* getters and setters */
}

public class User {
    private String name;
    private Group group;
    /* getters and setters */
}

Grid<User> grid = new Grid<>();
grid.addColumn(User::getName);
grid.addColumn(User::getGroup, Group::getName);
```
the `addColumn(ValueProvider<T, V> valueProvider, ValueProvider<V, P> presentationProvider)` is a new method to add a column which type is `V`, but the type the renderer will encode is `P`. In the example, the last column will be a `Column<User, Group>` but it uses a `TextRenderer` to encode the `Group`'s name.

This avoids the need to have custom renderers for every column whose type is an object, but also can be used with an explicit renderer, for example:

```java
Grid<User> grid = new Grid<>();
grid.addColumn(User::getName);
grid.addColumn(User::getGroup, Group::getPrivilegeLevel, new ProgressBarRenderer());
```

This also permits using editors components like this (see https://github.com/vaadin/framework/issues/8656):

```java
ComboBox<Group> groups = new ComboBox<>();
groups.setItemCaptionGenerator(Group::getName);
groups.setDataProvider(/* groups */);

Grid<User> grid = new Grid<>();
grid.addColumn(User::getName)
        .setEditorComponent(new TextField(), User::setName);
grid.addColumn(User::getGroup, Group::getName)
        .setEditorComponent(groups, User::setGroup);
```

I think this also may be useful when #8625 completes, i.e. having a component as the presentation type:

```java
Grid<User> grid = new Grid<>();
grid.addColumn(User::getName, name -> new CustomLabel(name));
grid.addColumn(User::getGroup, Group::getName);
```

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8806)

### DIFF
--- a/server/src/main/java/com/vaadin/data/ValueConverter.java
+++ b/server/src/main/java/com/vaadin/data/ValueConverter.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2000-2016 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.data;
+
+import com.vaadin.server.SerializableFunction;
+
+/**
+ * A callback interface for converting values from a given source.
+ * <p>
+ * For example this interface can be implemented to simply extract a value with
+ * a getter, or to create a composite value based on the fields of the source
+ * object.
+ *
+ * @author Vaadin Ltd.
+ * @since 8.0
+ *
+ * @param <SOURCE>
+ *            the type of the object used to provide the value
+ * @param <TARGET>
+ *            the type of the provided value
+ */
+@FunctionalInterface
+public interface ValueConverter<SOURCE, TARGET>
+        extends SerializableFunction<SOURCE, TARGET> {
+
+    /**
+     * Returns a value converter that always returns its input argument.
+     *
+     * @param <T>
+     *            the type of the input and output objects to the function
+     * @return a function that always returns its input argument
+     */
+    public static <T> ValueConverter<T, T> identity() {
+        return t -> t;
+    }
+
+    /**
+     * Provides a value from the given source object.
+     *
+     * @param source
+     *            the source to retrieve the value from
+     * @return the value provided by the source
+     */
+    @Override
+    public TARGET apply(SOURCE source);
+}
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -816,7 +816,8 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
      */
     public static class Column<T, V> extends AbstractGridExtension<T> {
 
-        private final SerializableFunction<T, ? extends V> valueProvider;
+        private final SerializableFunction<T, V> valueProvider;
+        private final SerializableFunction<V, ?> presentationProvider;
 
         private SortOrderProvider sortOrderProvider = direction -> {
             String id = getId();
@@ -844,19 +845,42 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
          *            the function to get values from items, not
          *            <code>null</code>
          * @param renderer
-         *            the type of value, not <code>null</code>
+         *            the value renderer, not <code>null</code>
          */
         protected Column(ValueProvider<T, V> valueProvider,
                 Renderer<? super V> renderer) {
+            this(valueProvider, ValueProvider.identity(), renderer);
+        }
+
+        /**
+         * Constructs a new Column configuration with given renderer and value
+         * provider.
+         *
+         * @param valueProvider
+         *            the function to get values from items, not
+         *            <code>null</code>
+         * @param presentationProvider
+         *            the function to get presentations for this column
+         *            values, not <code>null</code>
+         * @param renderer
+         *            the presentation renderer, not <code>null</code>
+         * @param <P> the presentation type
+         */
+        protected <P> Column(ValueProvider<T, V> valueProvider,
+                ValueProvider<V, P> presentationProvider,
+                Renderer<? super P> renderer) {
             Objects.requireNonNull(valueProvider,
                     "Value provider can't be null");
+            Objects.requireNonNull(presentationProvider,
+                    "Presentation provider can't be null");
             Objects.requireNonNull(renderer, "Renderer can't be null");
 
             ColumnState state = getState();
 
             this.valueProvider = valueProvider;
-            state.renderer = renderer;
+            this.presentationProvider = presentationProvider;
 
+            state.renderer = renderer;
             state.caption = "";
 
             // Add the renderer as a child extension of this extension, thus
@@ -864,7 +888,7 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
             // removed
             addExtension(renderer);
 
-            Class<? super V> valueType = renderer.getPresentationType();
+            Class<? super P> valueType = renderer.getPresentationType();
 
             if (Comparable.class.isAssignableFrom(valueType)) {
                 comparator = (a, b) -> compareComparables(
@@ -948,6 +972,11 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
 
         @Override
         public void generateData(T data, JsonObject jsonObject) {
+            generatePresentationData(data, jsonObject, presentationProvider);
+        }
+
+        private <P> void generatePresentationData(T data, JsonObject jsonObject,
+                SerializableFunction<V, P> presentationProvider) {
             ColumnState state = getState(false);
 
             String communicationId = getConnectorId();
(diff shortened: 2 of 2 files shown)


Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8806#-Kf5iI0ozFhLNsAsR2NG:-Kf5iI0ozFhLNsAsR2NH:b-xx6fnu) (no related file):*
I'm slightly sceptical about this approach for a couple of reasons:
* There are already quite many different variants of `addColumn`. Introducing two more would further add to the complexity and confusion.
* Passing two different `ValueProvider` instances to the same method might be confusing.
* Making the second `ValueProvider` accept the return value of the first `ValueProvider` does, from a semantic point of view, mean that it's not a "value provider", but rather a "converter".

One alternative approach that I've been considering is to introduce a "converting renderer" that wraps some existing render but adds a conversion callback "in front of" it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8806)*

heruan:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8806#-Kf5iI0ozFhLNsAsR2NG:-Kf60aPcNcT-Lsm0GcCd:b5siug0) (no related file):*

I'm slightly sceptical about this approach for a couple of reasons:
* There are already quite many different variants of `addColumn`. Introducing two more would further add to the complexity and confusion.
* Passing two different `ValueProvider` instances to the same method might be confusing.
* Making the second `ValueProvider` accept the return value of the first `ValueProvider` does, from a semantic point of view, mean that it's not a "value provider", but rather a "converter".

One alternative approach that I've been considering is to introduce a "converting renderer" that wraps some existing render but adds a conversion callback "in front of" it.
</blockquote></details>

I understand your concerns, and the "wrapped renderer" was my first approach. Problem is the wrapped renderer fails somewhere causing the grid to display a gray box instead of rows and columns. I guess it's something related to the extension hierarchy of the renderer, I tried to override `setParent` to add the wrapped renderer to the hierarchy but it still doesn't work. Any ideas?

I like the idea of a "value converter", which may be a better "wrapped renderer" and achieve the same usability without touching the grid. I'll keep investigating on the wrapped renderer and commit a new approach to this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8806)*

heruan:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8806#-Kf5iI0ozFhLNsAsR2NG:-Kf6Hpl4fagw2cJv_zJF:bw0djfw) (no related file):*
<details><summary><i>Previously, heruan (Giovanni Lovato) wrote…</i></summary><blockquote>

I understand your concerns, and the "wrapped renderer" was my first approach. Problem is the wrapped renderer fails somewhere causing the grid to display a gray box instead of rows and columns. I guess it's something related to the extension hierarchy of the renderer, I tried to override `setParent` to add the wrapped renderer to the hierarchy but it still doesn't work. Any ideas?

I like the idea of a "value converter", which may be a better "wrapped renderer" and achieve the same usability without touching the grid. I'll keep investigating on the wrapped renderer and commit a new approach to this PR.
</blockquote></details>

Of course the simple reason why a wrapping renderer won't work is that it lacks its client-side counterpart. This would be a nice generic renderer:

```java
public class ValueConverter<T, V, C> extends AbstractRenderer<T, Object> {

    private ValueProvider<V, C> valueProvider;

    private Renderer<? super C> renderer;

    public ValueConverter(ValueProvider<V, C> valueProvider) {
        this(valueProvider, new TextRenderer());
    }

    public ValueConverter(ValueProvider<V, C> valueProvider, Renderer<? super C> renderer) {
        super(Object.class);
        this.valueProvider = valueProvider;
        this.renderer = renderer;
    }

    @Override
    @SuppressWarnings("unchecked")
    public JsonValue encode(Object value) {
        if (value == null) {
            return this.renderer.encode(null);
        } else {
            return this.renderer.encode(this.valueProvider.apply((V) value));
        }
    }
}
```

… if only it could inherit the wrapped renderer connector and widget.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8806)*

heruan:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8806#-Kf5iI0ozFhLNsAsR2NG:-Kf6R-R5lzDhDc4Nzw8k:bw6wsoq) (no related file):*
<details><summary><i>Previously, heruan (Giovanni Lovato) wrote…</i></summary><blockquote>

Of course the simple reason why a wrapping renderer won't work is that it lacks its client-side counterpart. This would be a nice generic renderer:

```java
public class ValueConverter<T, V, C> extends AbstractRenderer<T, Object> {

    private ValueProvider<V, C> valueProvider;

    private Renderer<? super C> renderer;

    public ValueConverter(ValueProvider<V, C> valueProvider) {
        this(valueProvider, new TextRenderer());
    }

    public ValueConverter(ValueProvider<V, C> valueProvider, Renderer<? super C> renderer) {
        super(Object.class);
        this.valueProvider = valueProvider;
        this.renderer = renderer;
    }

    @Override
    @SuppressWarnings("unchecked")
    public JsonValue encode(Object value) {
        if (value == null) {
            return this.renderer.encode(null);
        } else {
            return this.renderer.encode(this.valueProvider.apply((V) value));
        }
    }
}
```

… if only it could inherit the wrapped renderer connector and widget.
</blockquote></details>

I really think the grid should provide a method to add a column with a *provide-convert-render* value presentation chain. Wrapping the conversion phase into a render won't work and won't permit the use of existing renderers.

I've updated to PR trying to address your original concerns with the introduction of a `ValueConverter` functional interface, which will also avoid confusion on choosing the right method overloading.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8806)*

Legioth:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8806#-Kf5iI0ozFhLNsAsR2NG:-KfAfKJdGsK5Qce_7Lmn:bg1gbgs) (no related file):*
<details><summary><i>Previously, heruan (Giovanni Lovato) wrote…</i></summary><blockquote>

I really think the grid should provide a method to add a column with a *provide-convert-render* value presentation chain. Wrapping the conversion phase into a render won't work and won't permit the use of existing renderers.

I've updated to PR trying to address your original concerns with the introduction of a `ValueConverter` functional interface, which will also avoid confusion on choosing the right method overloading.
</blockquote></details>

A wrapping renderer would either require a client side counterpart or some custom server-side logic that uses the renderer instance in different ways depending on its type. Either alternative would have some slightly ugly aspects.

Defining a separate `ValueConverter` type should definitely help making the different overloads easier to use.

I'll leave final judgement about this PR to the team that actually maintains the Framework. They'll also have to determine how to deal with automatic testing and external documentation for this functionality. Ping @pleku.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8806)*

tsuoanttila:
I was looking at this change in addition to a recent change regarding changing the renderer of a column. The columns should somehow either store the presentation type, so that the renderer can still be changed, or the converter needs to be given again when ever you want to change the renderer. At least that's what it looks like for me. So the question in my head is, what is the actual reason for providing the concept of presentation type and its conversion?

If the reason is mostly the editor component, I'd say we should add an overload of the .setEditorComponent that is typed to edited value type, accepting a HasValue<E>, ValueProvider<E> and a Setter<T, E>. This way we could provide the editor component with custom type without making the concept of the Column any more complex.

How does this sound for you @heruan and @Legioth ?

heruan:
Thank you @tsuoanttila for the feedback. This idea originated from the use of the editor component, but then I thought that having the column typed on the semantic instead of the presentation would be more elegant. Consider this:

```java
Grid<User> grid = new Grid<>();
Column<User, Group> column1 = grid.addColumn(User::getGroup);
Column<User, String> column2 = grid.addColumn(user -> user.getGroup().getName());
```
Here we know that `column1` will render a `Group`, and as you anticipate we can later change how the `Group` will be rendered; in `column2` instead, we restrict the rendering to a string and we can't access other field other than `name`.

That's why I see a missing piece here, i.e. the bridge between the semantic (column) and the presentation (renderer) which in the current implementation are coincident, but in my opinion should be decoupled.

Wouldn't be nice to be able to do this?

```java
Column<User, Group> column1 = grid.addColumn(User::getGroup);
column1.setRenderer(Group::getName, new TextRenderer<>());
// ... later on
Function<Group, Component> generateButton = group ->
    new Button(group.getName(), click -> navigator.navigateTo("groups/" + group.getId());
column1.setRenderer(generateButton, new ComponentRenderer<>());


```

Legioth:
Now when we have `Column.setRenderer` for changing the renderer after the column has been created, it might make sense to provide an additional overload of that method while still not adding yet another overload of `addColumn`.

This approach would allow using the suggested `Function<Group, Component> generateButton` while still avoiding the confusion of an `addColumn` overload that accept multiple lambda parameters.

tsuoanttila:
Ok, I think I get your point on the semantics of the column. It might be more correct to consider the column type of Group rather than String in this case, also makes sorting implementations much nicer with in-memory data. So basically what we do need to provide is the .setRenderer with custom type PRESENTATION that takes the renderer and the converter. This is missing from the patch. Can you rebase it on top of master and implement this missing feature?

pinonpierre:
I have the same issue.

In my project, I need to format my value in html. After that, I have no way to sort the value if I use "addColumn(ValueProvider" because I don't want to sort on html code!

hesara:
Was this superceded by #9553?

pinonpierre:
@Legioth 
It's strange to answer my comment by editing my own comment?

See above "pinonpierre commented on 29 May • edited by Legioth"

tsuoanttila:
Obsolete since #9553 was merged. Thanks @heruan for the idea and most of the code!

PR #8821, opened 2017-03-13
Fix type warnings

asashour:
Fix type warnings, as reported in Eclipse.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8821)

### DIFF
--- a/client/src/main/java/com/vaadin/client/communication/JsonEncoder.java
+++ b/client/src/main/java/com/vaadin/client/communication/JsonEncoder.java
@@ -85,12 +85,12 @@ public class JsonEncoder {
         } else if (value instanceof Enum) {
             return encodeEnum((Enum<?>) value, connection);
         } else if (value instanceof Map) {
-            return encodeMap((Map) value, type, connection);
+            return encodeMap((Map<Object, Object>) value, type, connection);
         } else if (value instanceof Connector) {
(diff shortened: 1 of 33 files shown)


elmot:
Review status: 0 of 33 files reviewed at latest revision, 1 unresolved discussion.

---

*[compatibility-server/src/main/java/com/vaadin/v7/data/fieldgroup/FieldGroup.java, line 1279 at r1](https://reviewable.io:443/reviews/vaadin/framework/8821#-KfAf_bO0UPbgNHacn6C:-KfAf_bO0UPbgNHacn6D:b2dntq9) ([raw file](https://github.com/vaadin/framework/blob/56c5dd758b641a6e6cab2e1b6b59066589679780/compatibility-server/src/main/java/com/vaadin/v7/data/fieldgroup/FieldGroup.java#L1279)):*
> <?>
Thanks for your contributions, but unfortunately this change can not be accepted, because we are trying to keep compatibility package code as close as possible to 7.7 branch. Any code cleanup without actual bug fixing in those parts break this rule and, thus, makes code maintenance unnecessarily complicated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8821)*

asashour:
I see, thanks for your feedback.

PR #8826, opened 2017-03-14
Fix removing rows from the middle of Escalator

### LINKED ISSUE #8825: Escalator renders wrong rows when removing rows from middle
When the number of rows in escalator is more than what is currently rendered, and there are rows removed from the middle of the escalator so that that number of rows after removing will be less than what is currently shown, the rows are rendered wrong.

This is caused by removing some rows from the middle, then moving rest of the rows up to "fill the gap", but then only the data for the "new rows" from the end is updated.

pleku:
Fixes #8825

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8826)

### DIFF
--- a/client/src/main/java/com/vaadin/client/widgets/Escalator.java
+++ b/client/src/main/java/com/vaadin/client/widgets/Escalator.java
@@ -3130,14 +3130,8 @@ public class Escalator extends Widget
                         y += spacerContainer.getSpacerHeight(i);
                     }
 
-                    /*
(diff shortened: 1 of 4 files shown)


ahie:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8826#-:-KfBYRU8HwwlwT2xOMH7:bnfp4nl)*

PR #8844, opened 2017-03-14
AbstractDateField.setDateFormat() To handle 'z' (timezone) as in SimpleDateFormat

asashour:
`AbstractDateField`.`setDateFormat()` To handle `z` (timezone) as in `SimpleDateFormat`.

As asked in the [forum](https://vaadin.com/forum#!/thread/15337242).

This patch is to initiate further discussion.

1. How to get the `TimeZone`/`TimeZoneInfo` to use? Should it be sent from the server in `LocaleService`?
2. How to iterate over all system `TimeZone`s?
3. Should we check if `z` is inside a single quote, so that it is skipped? Note that in `EEEE`/`MMM` this check is not verified as well.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8844)


elmot:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8844#-KfFvdfgpbdBtA3FcKfS:-KfFvdfhIR0tOG0uWsl5:b14kwie) (no related file):*
Tests are missing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8844#-KfFvdfgpbdBtA3FcKfS:-KfFxhiAoGT-sbd_dPKN:b-8i5yi7) (no related file):*

Tests are missing
</blockquote></details>

Thanks for feedback.

Yes, tests are missing and the provided implementation uses `Europe/Berlin` by default, which is obviously wrong.

I would also like to know:

1. How to get the `TimeZone`/`TimeZoneInfo` to use? Should it be sent from the server in `LocaleService`?
2. How to iterate over all system `TimeZone`s?
3. Should we check if `z` is inside a single quote, so that it is skipped? Note that in `EEEE`/`MMM` this check is not verified as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:
Implementation is completely done, with test cases.

pleku:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8844#-KfFvdfgpbdBtA3FcKfS:-KfotN4G_dloEaKNH1Yi:b-idv2e0) (no related file):*

Thanks for feedback.

Yes, tests are missing and the provided implementation uses `Europe/Berlin` by default, which is obviously wrong.

I would also like to know:

1. How to get the `TimeZone`/`TimeZoneInfo` to use? Should it be sent from the server in `LocaleService`?
2. How to iterate over all system `TimeZone`s?
3. Should we check if `z` is inside a single quote, so that it is skipped? Note that in `EEEE`/`MMM` this check is not verified as well.
</blockquote></details>

Thanks for the contribution, couple comments:
- There is no need to currently apply the change to the compatibility package classes since that would make those out of sync with 7.7, if those were to be updated, then that should happen only after this change would has been picked to 7.7 branch. You can however keep them for there for now
- You can leave the since tags empty, since it is not guaranteed this would make to 8.0.4, we have a script for updating empty `@since` tags when doing a release

I'll try to find the time to look at the actual implementation this week or early next week.

---

*[uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java, line 17 at r1](https://reviewable.io:443/reviews/vaadin/framework/8844#-KfopYSfjN-0jWCwSU9Q:-KfopYSfjN-0jWCwSU9R:b-h24cfe) ([raw file](https://github.com/vaadin/framework/blob/c3dc6234bbbb6709b8b2e198c2d7c402d282c981/uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java#L17)):*
> AbstractReindeerTestUI

Please just use AbstractTestUI instead

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:
Hi all,

Appreciate if someone has time to review this.

tsuoanttila:
We're currently having a discussion on how to tackle some persisting issues in the DateFields. These changes might potentially make your change completely obsolete. Rest assured, if we do decide to make the changes we've been discussing, this use-case will be taken into account and made sure it will work. If we decide not to make those changes, we'll have another round of testing and review on your patch. We hope to come back to the topic of DateFields soon, and we'll be contacting you again at that point at least to see if our solution solves your case.

hesara:
There seem to be some compilation issues also with the latest version with changes from master merged.

As we haven't managed to perform a more comprehensive redesign of parts of date fields and that is unlikely to happen by 8.2, we will at least consider integrating this PR.

You can also take a look at the slightly related PR #10041, which transfers the client time zone to the server (if available - will not be on IE11).

asashour:
Thanks for hinting.

Getting the client time zone to the server doesn't affect this PR, since the user determines which `ZoneId` is used by the `AbstractDateField`, which can be client-based or server-side logic (e.g. from database).

The interesting question, couldn't we send the time zone name to the client, and it detects the timezone name for the specified date, unfortunately it seems not currently possible, since the below with latest Chrome gives `Central European Summer Time`

    var tz = Intl.DateTimeFormat().resolvedOptions().timeZone;
    var options = { timeZoneName: 'long',  timeZone: tz };

    var format = new Intl.DateTimeFormat('en', options);
    var parts = format.formatToParts(new Date(2017, 6, 6));
    for (var p in parts) {
        if (parts[p].type == 'timeZoneName') {
            alert(parts[p].value);
        }
    }

And changing the options`timeZoneName` to `short` returns `GMT+2`, but what is actually needed is `CEST`, which is determined by the server and sent to the client in the PR.

hesara:
Dis a partial review with some comments.

---

Reviewed 3 of 4 files at r3, 1 of 2 files at r4.
Review status: 4 of 7 files reviewed at latest revision, 8 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/DateTimeService.java, line 330 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwuyxrF73_6xGVBy88:-KuwuyxrF73_6xGVBy89:b-auejn8) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/DateTimeService.java#L330)):*
> ```Java
>      * @param formatStr
>      *            The format string that might contain MMM or MMMM
>      * @param timeZoneJSON
> ```

I'm hesitating on whether this should use the JSON or rather pre-parsed TimeZoneInfo or TimeZone. Those would allow parsing the info just once, and give a cleaner API.

---

*[client/src/main/java/com/vaadin/client/DateTimeService.java, line 343 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwuLizHQdvUlNUxFtm:-KuwuLizHQdvUlNUxFtn:bgsythi) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/DateTimeService.java#L343)):*
> ```Java
>         formatStr = formatMonthNames(date, formatStr);
>         formatStr = formatDayNames(date, formatStr);
>         formatStr = formatTimeZone(date, formatStr, timeZoneJSON);
> ```

This should probably be done before month formatting, as some month names could contain a "z" in some languages.

---

*[client/src/main/java/com/vaadin/client/DateTimeService.java, line 459 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwvkMMighhGFiE-Brz:-KuwvkMMighhGFiE-Bs-:b-ma0vjl) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/DateTimeService.java#L459)):*
> ```Java
>         }
>         int end = start;
>         while (end + 1 < formatStr.length()
> ```

This loop takes a bit of time to understand (dealing with multiple "z"s, handling quotes, skipping extra "z"s) so there could be a comment for what is being done in replaceTimeZone().
Do multiple "z"s have a different meaning?

---

*[client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java, line 74 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwxwBJJPb3qc6KTu6p:-KuwxwBJJPb3qc6KTu6q:b-nfrfhw) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java#L74)):*
> ```Java
> 
>     /** For internal use only. May be removed or replaced in the future. */
>     public String timeZoneJSON;
> ```

I'd perhaps prefer a pre-parsed value here, set by the connector.

---

*[client/src/main/java/com/vaadin/client/ui/datefield/AbstractTextualDateConnector.java, line 49 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kux62LqOddyaeeXnTgY:-Kux62LqOddyaeeXnTgZ:b-ftouru) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/ui/datefield/AbstractTextualDateConnector.java#L49)):*
> ```Java
>             getWidget().setFormatString(uidl.getStringAttribute("format"));
>         }
>         if (uidl.hasAttribute("timeZoneJSON")) {
> ```

Could this be transferred in the state instead, and pre-parse it on reception? Or would that be costly? See also comments below.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 184 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kux6gcscr3T3nKFK6t3:-Kux6gcscr3T3nKFK6t4:bwi2f49) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L184)):*
> ```Java
>         }
> 
>         if (getZoneId() != null) {
> ```

I'd prefer this to be in the state, perhaps so that the "cached" JSON version in the state is reset when a ZoneId is set and regenerated in beforeClientResponse(boolean). This way, you'd regenerate and resend it only when it changes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844#-:-Kux9Rb9EopcpdaZD1mM:b-2fpmvf)*

asashour:
Review status: 4 of 7 files reviewed at latest revision, 8 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/DateTimeService.java, line 330 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwuyxrF73_6xGVBy88:-KuxCzqKStkB1oG2UV_w:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/DateTimeService.java#L330)):*

I'm hesitating on whether this should use the JSON or rather pre-parsed TimeZoneInfo or TimeZone. Those would allow parsing the info just once, and give a cleaner API.
</blockquote></details>

Done.

---

*[client/src/main/java/com/vaadin/client/DateTimeService.java, line 343 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwuLizHQdvUlNUxFtm:-KuxCDJeDvg2u_7ECc1W:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/DateTimeService.java#L343)):*

This should probably be done before month formatting, as some month names could contain a "z" in some languages.
</blockquote></details>

Done.

---

*[client/src/main/java/com/vaadin/client/DateTimeService.java, line 459 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwvkMMighhGFiE-Brz:-KuxFluv59LUKFURQcyt:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/DateTimeService.java#L459)):*

This loop takes a bit of time to understand (dealing with multiple "z"s, handling quotes, skipping extra "z"s) so there could be a comment for what is being done in replaceTimeZone().
Do multiple "z"s have a different meaning?
</blockquote></details>

Done.

---

*[client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java, line 74 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuwxwBJJPb3qc6KTu6p:-Kuxbqy_Te2kMhd4Ol9F:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java#L74)):*

I'd perhaps prefer a pre-parsed value here, set by the connector.
</blockquote></details>

Done.

---

*[client/src/main/java/com/vaadin/client/ui/datefield/AbstractTextualDateConnector.java, line 49 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kux62LqOddyaeeXnTgY:-KuxfhDZrMNMB83TpEzj:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/client/src/main/java/com/vaadin/client/ui/datefield/AbstractTextualDateConnector.java#L49)):*

Could this be transferred in the state instead, and pre-parse it on reception? Or would that be costly? See also comments below.
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 184 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kux6gcscr3T3nKFK6t3:-KuxfWiOn9l1bLa77sn-:b-gv0gu6) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L184)):*

I'd prefer this to be in the state, perhaps so that the "cached" JSON version in the state is reset when a ZoneId is set and regenerated in beforeClientResponse(boolean). This way, you'd regenerate and resend it only when it changes.
</blockquote></details>

Done. I hope I correctly understood. Another PR should be to also migrate all properties to the State

---

*[uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java, line 17 at r1](https://reviewable.io:443/reviews/vaadin/framework/8844#-KfopYSfjN-0jWCwSU9Q:-Kuxck9Pl5-WyijIbKPv:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/c3dc6234bbbb6709b8b2e198c2d7c402d282c981/uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java#L17)):*

> AbstractReindeerTestUI

Please just use AbstractTestUI instead
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

hesara:
Reviewed 1 of 7 files at r2, 1 of 2 files at r4, 5 of 5 files at r5.
Review status: 7 of 8 files reviewed at latest revision, 12 unresolved discussions.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 184 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kux6gcscr3T3nKFK6t3:-KuxlzI8r3rDyPbdfNVA:bd1fhsd) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L184)):*

Done. I hope I correctly understood. Another PR should be to also migrate all properties to the State
</blockquote></details>

There are quite a few components that should be migrated to transfer data in state (and for transient information with RPC), but that can also be a breaking change especially for add-ons extending the components. This is anyway the direction we'd like to go but we haven't had the time to convert everything.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 222 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxmSCgGUt0wB6NzDcU:-KuxmSCgGUt0wB6NzDcV:bunmnlu) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L222)):*
> ```Java
>             timeZoneJSON = null;
>         }
>         getState().timeZoneJSON = timeZoneJSON;
> ```

This could check if the information has actually changed (based on getState(false)) and only call getState() that marks the field as dirty if it has.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 56 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxGDa6lhhQi6lmNxPP:-KuxGDa6lhhQi6lmNxPQ:b-7cqexg) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L56)):*
> ```Java
>      *            the {@link ZoneId} to get the daylight transitions from
>      * @param locale
>      *            the locale the locale to translate the name
> ```

"the locale the locale" and also otherwise a bit unclear parameter description

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 67 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxGP-v0nVDxa8agAe_:-KuxGP-v0nVDxa8agAea:bl95yw1) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L67)):*
> ```Java
>         TimeZone timeZone = TimeZone.getTimeZone(zoneId);
>         List<Long> transtionsList = new ArrayList<>();
>         
> ```

(extra whitespace here and below)

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 70 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxGmVc0k3neq-NzG8z:-KuxGmVc0k3neq-NzG9-:bdk0aig) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L70)):*
> ```Java
>         TimeZoneInfo info = new TimeZoneInfo();
>         
>         int endYear = LocalDate.now().getYear() + 20;
> ```

could make constants from the 20 here and 1980 below so that they would be easier to find if they need to be updated someday

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 77 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxH7uUfKyqPAK_lgBL:-KuxH7uUfKyqPAK_lgBM:b-ohj0m7) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L77)):*
> ```Java
>                     ZoneOffsetTransition t = rules.nextTransition(i.toInstant());
>                     i = t.getInstant().atZone(zoneId);
>                     if (i.toLocalDate().getYear() != year) {
> ```

Can it happen that there is no transition for a given year, or only for a specific year? This could happen e.g. when a country switches back and forth between using daylight time and not like Russia did? Could we get stuck in this loop in some cases?

---

*[shared/src/main/java/com/vaadin/shared/ui/datefield/AbstractDateFieldState.java, line 52 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxmhqW0cpABDGvaWR8:-KuxmhqXZf4YzJqZfSl2:bv2excv) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/shared/src/main/java/com/vaadin/shared/ui/datefield/AbstractDateFieldState.java#L52)):*
> ```Java
>     /**
>      * The JSON used to construct a TimeZone on the client side, can be
>      * {@code null}.
> ```

could add a since tag for this as well

---

*[uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java, line 43 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kuxn5Q3Sz-j3uBY-voW:-Kuxn5Q3Sz-j3uBY-voX:bdk1ssm) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java#L43)):*
> ```Java
>         zoneIdComboBox.setItems(zoneIdSet);
>         addComponent(zoneIdComboBox);
>     
> ```

(extra whitespace here and below on empty lines)

---

*[uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java, line 44 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kuxu7UCZXUiy9zBR8-Q:-Kuxu7UCZXUiy9zBR8-R:buvwd5z) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java#L44)):*
> ```Java
> import com.vaadin.tests.tb3.MultiBrowserTest;
> 
> public class DateTimeFieldZoneIdTest extends MultiBrowserTest {
> ```

Does the test work correctly regardless of system settings (locale, time zone)?

---

*[uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java, line 190 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxssvNY3TPiNeYDODB:-KuxssvNY3TPiNeYDODC:b-iobmpa) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java#L190)):*
> ```Java
>     }
> 
>     private void openDatePicker(AbstractDateFieldElement dateField) {
> ```

DateFieldElement actually has some relevant methods (openPopup(), getValue(), getInputElement()) if you can use it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:
Review status: 3 of 8 files reviewed at latest revision, 9 unresolved discussions.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 222 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxmSCgGUt0wB6NzDcU:-Kuy7A5DN6lnTtrTP8yV:b-drxhe0) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L222)):*

This could check if the information has actually changed (based on getState(false)) and only call getState() that marks the field as dirty if it has.
</blockquote></details>

Done. Now I see the pattern, and determination is done in `setZoneId`

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 56 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxGDa6lhhQi6lmNxPP:-Kuy8nRkxzkcDwsUyKZ-:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L56)):*

"the locale the locale" and also otherwise a bit unclear parameter description
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 67 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxGP-v0nVDxa8agAe_:-Kuy8ojEOLq9vDu75rhY:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L67)):*

(extra whitespace here and below)
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 70 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxGmVc0k3neq-NzG8z:-Kuy9bhoSSJ608ifwQCY:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L70)):*

could make constants from the 20 here and 1980 below so that they would be easier to find if they need to be updated someday
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 77 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxH7uUfKyqPAK_lgBL:-Kuy9cEegvq25vg8h1vF:bveb15p) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L77)):*

Can it happen that there is no transition for a given year, or only for a specific year? This could happen e.g. when a country switches back and forth between using daylight time and not like Russia did? Could we get stuck in this loop in some cases?
</blockquote></details>

This is already handled, the `break` is for the `while(true)`, and the outer`for` loop iterates over all years.

---

*[shared/src/main/java/com/vaadin/shared/ui/datefield/AbstractDateFieldState.java, line 52 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxmhqW0cpABDGvaWR8:-Kuy9vmKZOaWoX0wZ8k3:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/shared/src/main/java/com/vaadin/shared/ui/datefield/AbstractDateFieldState.java#L52)):*

could add a since tag for this as well
</blockquote></details>

Done.

---

*[uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java, line 43 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kuxn5Q3Sz-j3uBY-voW:-KuyAL2bF-59ev9vXjo1:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/uitest/src/main/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneId.java#L43)):*

(extra whitespace here and below on empty lines)
</blockquote></details>

Done.

---

*[uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java, line 44 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kuxu7UCZXUiy9zBR8-Q:-KuyAZqNkfnukb3iP9m9:b-vr2lrg) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java#L44)):*

Does the test work correctly regardless of system settings (locale, time zone)?
</blockquote></details>

I believe so, the `init()` sets the time zone to `Brazil`, and individual tests handle the default name, in Tokyo, Berlin, with German language, and in Lebanon with German language

---

*[uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java, line 190 at r5](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxssvNY3TPiNeYDODB:-KuyEk1qUyWtbVCbOp4k:b-mu99eg) ([raw file](https://github.com/vaadin/framework/blob/03eb40820ca9f5ce76077a47a2ce6d9ec21af80a/uitest/src/test/java/com/vaadin/tests/components/datefield/DateTimeFieldZoneIdTest.java#L190)):*

DateFieldElement actually has some relevant methods (openPopup(), getValue(), getInputElement()) if you can use it.
</blockquote></details>

Done. Thanks for hinting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

hesara:
Looking quite good, I'll finish reviewing the last file when I can focus a bit better on checking the logic.

---

Reviewed 4 of 5 files at r6.
Review status: 7 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 469 at r6](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuyGA61zbuueTixjMIB:-KuyGA61zbuueTixjMIC:b-ff6aqk) ([raw file](https://github.com/vaadin/framework/blob/9b1039ae3026b89d6abe8cbed7157ad3bcf0102d/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L469)):*
> ```Java
>             }
>             getState().timeZoneJSON = timeZoneJSON;
>             markAsDirty();
> ```

markAsDirty() is redundant as getState() does it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844#-:-KuyHlO7fjSWu_O2Y-ZN:b2kbfh0)*

asashour:
Review status: 6 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/ui/AbstractDateField.java, line 469 at r6](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuyGA61zbuueTixjMIB:-KuydfHrhWj0x7-wyZsj:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/9b1039ae3026b89d6abe8cbed7157ad3bcf0102d/server/src/main/java/com/vaadin/ui/AbstractDateField.java#L469)):*

markAsDirty() is redundant as getState() does it
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

hesara:
Reviewed 1 of 5 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 77 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxH7uUfKyqPAK_lgBL:-Kv1rsX6IBYc-JphQWkU:bufeuiy) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L77)):*

This is already handled, the `break` is for the `while(true)`, and the outer`for` loop iterates over all years.
</blockquote></details>

An eternal loop in the while(true) is what I meant, though what would actually happen is an NPE e.g. if a country stops switching between summer and winter time, and nextTransition(i) returns null.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 77 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1sHplgNzzFvy1yV_4:-Kv1sHplgNzzFvy1yV_5:b-5zb02x) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L77)):*
> ```Java
>         ZoneRules rules = zoneId.getRules();
>         TimeZone timeZone = TimeZone.getTimeZone(zoneId);
>         List<Long> transtionsList = new ArrayList<>();
> ```

transitionsList

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 93 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1tRXjgbSaWKpWsz4j:-Kv1tRXjgbSaWKpWsz4k:b2xs62g) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L93)):*
> ```Java
>                         break;
>                     }
>                     long epocHours = Duration
> ```

Is this the format GWT TimeZoneInfo expects? Why is this in hours but the duration in minutes on the same list?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 77 at r4](https://reviewable.io:443/reviews/vaadin/framework/8844#-KuxH7uUfKyqPAK_lgBL:-Kv2Yy5ntyP1OTEa8Dmo:b-tkbxfs) ([raw file](https://github.com/vaadin/framework/blob/2e1557b9fbbd9208ddcd53fb64473e4fd963f477/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L77)):*

An eternal loop in the while(true) is what I meant, though what would actually happen is an NPE e.g. if a country stops switching between summer and winter time, and nextTransition(i) returns null.
</blockquote></details>

Done. You are right.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 77 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1sHplgNzzFvy1yV_4:-Kv2YTlbW-CAmV76xcnV:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L77)):*

transitionsList
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 93 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1tRXjgbSaWKpWsz4j:-Kv2_vRzBeaSh9WgRds4:b-18pi8m) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L93)):*

Is this the format GWT TimeZoneInfo expects? Why is this in hours but the duration in minutes on the same list?
</blockquote></details>

Well, it is the expected format, see [here](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/client/constants/TimeZoneConstants.properties).

I didn't add a JUnit test to test a fixed string value, because the result depends on the `STARTING_YEAR`/`YEARS_FROM_NOW`, and so the JUnit test and the implementation will simply be the same logic!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:
Review status: 7 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 93 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1tRXjgbSaWKpWsz4j:-Kv2bi_m7vP3erCiL1hh:bwguxia) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L93)):*

Well, it is the expected format, see [here](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/client/constants/TimeZoneConstants.properties).

I didn't add a JUnit test to test a fixed string value, because the result depends on the `STARTING_YEAR`/`YEARS_FROM_NOW`, and so the JUnit test and the implementation will simply be the same logic!
</blockquote></details>

In [TimeZone](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/client/TimeZone.java#L188) the comparison of transition in hours, but the value returned is in [minutes](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/shared/TimeZone.java#L26).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

hesara:
Reviewed 1 of 1 files at r10.

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 93 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1tRXjgbSaWKpWsz4j:-Kv5xYJRAMQFDqP6BVog:b8op1km) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L93)):*

In [TimeZone](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/client/TimeZone.java#L188) the comparison of transition in hours, but the value returned is in [minutes](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/shared/TimeZone.java#L26).
</blockquote></details>

Based on this, my impression is that TimeZoneOffset.getDuration() would return a positive or a negative number that is the change from the previous period, whereas the values expected in the JSON format would be offsets from the standard time and not the previous period. I see the Math.max below trying to fix the values. That could in theory go wrong if some transitions back are not of the expected size, but in practice it looks like the data in TimeZoneConstants has no negative values and all transitions are either a full hour or half an hour from standard time and then back to standard time, so this is probably ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:

---

*[server/src/main/java/com/vaadin/util/TimeZoneUtil.java, line 93 at r8](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv1tRXjgbSaWKpWsz4j:-Kv6CIme9B2U-2so2aEb:bjlzptp) ([raw file](https://github.com/vaadin/framework/blob/0632283b43e50a54d96d3fd4a945c7293c33953d/server/src/main/java/com/vaadin/util/TimeZoneUtil.java#L93)):*

Based on this, my impression is that TimeZoneOffset.getDuration() would return a positive or a negative number that is the change from the previous period, whereas the values expected in the JSON format would be offsets from the standard time and not the previous period. I see the Math.max below trying to fix the values. That could in theory go wrong if some transitions back are not of the expected size, but in practice it looks like the data in TimeZoneConstants has no negative values and all transitions are either a full hour or half an hour from standard time and then back to standard time, so this is probably ok.
</blockquote></details>

Yes, ZoneOffsetTransition.getDuration() has javadoc: "The duration will be positive for a gap and negative for an overlap."

However, [TimeZone.getDaylightAdjustment()](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/i18n/shared/TimeZone.java#L28) returns positive or 0.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

hesara:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv6a5pJo5NMAn0A5gtd:-Kv6a5pJo5NMAn0A5gte:b-7zsse0) (no related file):*
If you could update the empty since tags to say "8.2", I think we can merge this today. For any changes already merged to master, I made the updates in another PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

asashour:
Review status: 4 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8844#-Kv6a5pJo5NMAn0A5gtd:-Kv6cC1o-_vuFpGSOGzH:b-896fix) (no related file):*

If you could update the empty since tags to say "8.2", I think we can merge this today. For any changes already merged to master, I made the updates in another PR.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844)*

hesara:
:lgtm:

---

Reviewed 4 of 4 files at r11.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8844#-:-Kv6cR1U7_TG-HMTRJ8w:bnfp4nl)*

PR #8927, opened 2017-03-24
Fix infinite loop in data communication

### LINKED ISSUE #8479: Infinite loop if DataProvider returns no rows
If `size` in a DataProvider returns 1 but `fetch` returns an empty stream, Grid will ask again and again and again for data although there is none (e.g. somebody has deleted the row between the `size` call and the `fetch` call). The UI shows a row without data, a blinking loading indicator and completely stops working.

ahie:
Fixes #8479

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8927)

### DIFF
--- a/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
+++ b/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
@@ -334,7 +334,7 @@ public class DataCommunicator<T> extends AbstractExtension {
                             inMemorySorting, filter))
                     .collect(Collectors.toList());
 
-            if (!initial && rowsToPush.size() < limit) {
+            if (!initial && rowsToPush.size() == 0) {
                 triggerReset = true;
             }
 
--- a/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
+++ b/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
@@ -334,7 +334,7 @@ public class DataCommunicator<T> extends AbstractExtension {
                             inMemorySorting, filter))
                     .collect(Collectors.toList());
 
-            if (!initial && rowsToPush.size() == 0) {
+            if (!initial && !reset && rowsToPush.size() == 0) {
                 triggerReset = true;
             }
 
--- a/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
+++ b/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
@@ -338,7 +338,7 @@ public class DataCommunicator<T> extends AbstractExtension {
                 triggerReset = true;
             }
 
-            pushData(offset, rowsToPush.stream());
+            pushData(offset, rowsToPush);
         }
 
         if (!updatedData.isEmpty()) {
@@ -399,18 +399,17 @@ public class DataCommunicator<T> extends AbstractExtension {
      * @param data
      *            data objects to send as an iterable
      */
-    protected void pushData(int firstIndex, Stream<T> data) {
+    protected void pushData(int firstIndex, List<T> data) {
         JsonArray dataArray = Json.createArray();
 
         int i = 0;
-        List<T> collected = data.collect(Collectors.toList());
-        for (T item : collected) {
+        for (T item : data) {
             dataArray.set(i++, getDataObject(item));
         }
 
         rpc.setData(firstIndex, dataArray);
-        handler.addActiveData(collected.stream());
-        handler.cleanUp(collected.stream());
+        handler.addActiveData(data.stream());
+        handler.cleanUp(data.stream());
     }
 
     /**
--- a/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
+++ b/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
@@ -323,15 +323,22 @@ public class DataCommunicator<T> extends AbstractExtension {
         }
 
         Range requestedRows = getPushRows();
+        boolean triggerReset = false;
         if (!requestedRows.isEmpty()) {
             int offset = requestedRows.getStart();
             int limit = requestedRows.length();
 
             @SuppressWarnings({ "rawtypes", "unchecked" })
-            Stream<T> rowsToPush = getDataProvider().fetch(new Query(offset,
-                    limit, backEndSorting, inMemorySorting, filter));
+            List<T> rowsToPush = (List<T>) getDataProvider()
+                    .fetch(new Query(offset, limit, backEndSorting,
+                            inMemorySorting, filter))
+                    .collect(Collectors.toList());
 
-            pushData(offset, rowsToPush);
+            if (!initial && rowsToPush.size() < limit) {
+                triggerReset = true;
+            }
+
+            pushData(offset, rowsToPush.stream());
         }
 
         if (!updatedData.isEmpty()) {
@@ -344,7 +351,7 @@ public class DataCommunicator<T> extends AbstractExtension {
         }
 
         setPushRows(Range.withLength(0, 0));
-        reset = false;
+        reset = triggerReset;
         updatedData.clear();
     }
 
--- a/server/src/test/java/com/vaadin/data/provider/ReplaceDataProviderTest.java
+++ b/server/src/test/java/com/vaadin/data/provider/ReplaceDataProviderTest.java
@@ -79,7 +79,7 @@ public class ReplaceDataProviderTest {
                 beans1);
 
         dataCommunicator.setDataProvider(dataProvider, null);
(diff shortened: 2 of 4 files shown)


hesara:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 338 at r1](https://reviewable.io:443/reviews/vaadin/framework/8927#-KfzTbKnBbOVeKo7cfMr:-KfzTbKnBbOVeKo7cfMs:b-2wb58l) ([raw file](https://github.com/vaadin/framework/blob/70786bcd1562f1a41568ebe50a27775b64508ad3/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L338)):*
> ```Java
> 
>             if (!initial && rowsToPush.size() < limit) {
>                 triggerReset = true;
> ```

I think this might be too aggressive - as Teemu said, maybe we should only do this when the size equals zero, even though that might lead to another round trip to check if there really are no more items available.

---

*[uitest/src/main/java/com/vaadin/tests/components/grid/DataCommunicatorInfiniteLoop.java, line 26 at r1](https://reviewable.io:443/reviews/vaadin/framework/8927#-KfzTvWow6f1NQe7RHF6:-KfzTvWow6f1NQe7RHF7:b-csx0oo) ([raw file](https://github.com/vaadin/framework/blob/70786bcd1562f1a41568ebe50a27775b64508ad3/uitest/src/main/java/com/vaadin/tests/components/grid/DataCommunicatorInfiniteLoop.java#L26)):*
> ```Java
> 
>         @Override
>         public int size(Query<String, Void> query) {
> ```

I think it is a real case where some DPs return a subset of the requested rows and another request is needed to get the rest. Thus, this could use size 1/2 and return a stream of one item if the query starts from zero index to cover more real cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

tsuoanttila:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[uitest/src/main/java/com/vaadin/tests/components/grid/DataCommunicatorInfiniteLoop.java, line 26 at r1](https://reviewable.io:443/reviews/vaadin/framework/8927#-KfzTvWow6f1NQe7RHF6:-KfzVmkZS_-MW4emQDwo:b-m258iq) ([raw file](https://github.com/vaadin/framework/blob/70786bcd1562f1a41568ebe50a27775b64508ad3/uitest/src/main/java/com/vaadin/tests/components/grid/DataCommunicatorInfiniteLoop.java#L26)):*

I think it is a real case where some DPs return a subset of the requested rows and another request is needed to get the rest. Thus, this could use size 1/2 and return a stream of one item if the query starts from zero index to cover more real cases.
</blockquote></details>

to test this issue without a complex implementation, why not use ListDataProvider, remove one item from the list and not call .refreshAll(). End result should be the same?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

ahie:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 338 at r1](https://reviewable.io:443/reviews/vaadin/framework/8927#-KfzTbKnBbOVeKo7cfMr:-KfzcOqqIvMr0XkAqEcW:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/70786bcd1562f1a41568ebe50a27775b64508ad3/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L338)):*

I think this might be too aggressive - as Teemu said, maybe we should only do this when the size equals zero, even though that might lead to another round trip to check if there really are no more items available.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

hesara:
Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

hesara:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

pleku:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 338 at r3](https://reviewable.io:443/reviews/vaadin/framework/8927#-Kg0dvV_tdTERCZB2M0r:-Kg0dvV_tdTERCZB2M0s:b-m26dq6) ([raw file](https://github.com/vaadin/framework/blob/b8a742c9a6b538090ad12af0def15406262ca665/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L338)):*
> ```Java
> 
>             if (!initial && !reset && rowsToPush.size() == 0) {
>                 triggerReset = true;
> ```

So what happens if the size has been e.g 100, client is requesting rows 60-100, but then the fetch returns only 30 ? Does the client keep on requesting ?

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 341 at r3](https://reviewable.io:443/reviews/vaadin/framework/8927#-Kg0goafH5kNWky_BCSE:-Kg0goafH5kNWky_BCSF:b-cr96ap) ([raw file](https://github.com/vaadin/framework/blob/b8a742c9a6b538090ad12af0def15406262ca665/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L341)):*
> ```Java
>             }
> 
>             pushData(offset, rowsToPush.stream());
> ```

Could `pushData` be changed to accept a List instead since it is collecting it it again anyway

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

ahie:

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 338 at r3](https://reviewable.io:443/reviews/vaadin/framework/8927#-Kg0dvV_tdTERCZB2M0r:-KgDFOENcljybZxP_hOY:brlln8q) ([raw file](https://github.com/vaadin/framework/blob/b8a742c9a6b538090ad12af0def15406262ca665/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L338)):*

So what happens if the size has been e.g 100, client is requesting rows 60-100, but then the fetch returns only 30 ? Does the client keep on requesting ?
</blockquote></details>

Yes, it keeps requesting.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 341 at r3](https://reviewable.io:443/reviews/vaadin/framework/8927#-Kg0goafH5kNWky_BCSE:-KgDFrvErA7TSWjD43X0:b-4pm7kg) ([raw file](https://github.com/vaadin/framework/blob/b8a742c9a6b538090ad12af0def15406262ca665/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L341)):*

Could `pushData` be changed to accept a List instead since it is collecting it it again anyway
</blockquote></details>

Would not be backwards compatible, since `pushData` is protected.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

pleku:

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 341 at r3](https://reviewable.io:443/reviews/vaadin/framework/8927#-Kg0goafH5kNWky_BCSE:-KgDJUWnhKUhRdWd5ibR:bqumars) ([raw file](https://github.com/vaadin/framework/blob/b8a742c9a6b538090ad12af0def15406262ca665/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L341)):*

Would not be backwards compatible, since `pushData` is protected.
</blockquote></details>

Well you cannot pick this for 8.0 anyways I think since 8.0 and 8.1 versions of this class have forked, `sentDataToClient´ should not be existing in 8.0 I think (made in HierarchicalDataProvider patch)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

pleku:
Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8927)*

pleku:
Should not be picked for 8.0 branch, needs a separate PR.

casz69:
The reset method inform to client-side Datasource that all data has been invalidated.

¿ Is there some method to inform client-side Datasource that size has been changed ?

casz69:
Framework 8.1.0 has the same problem.

tsuoanttila:
Hi @casz69 

Can you open a  ticket about this issue including a brief description what you are trying to achieve?

PR #8983, opened 2017-03-31
Make getValueProvider return ValueProvider

Legioth:
The instance can only be a ValueProvider, but the getter previously
returned a less specific type. Making the getter return ValueProvider
makes it easier to reuse the provider in other contexts that require
ValueProvider, e.g. when adding filters to a ListDataProvider.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8983)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -69,7 +69,6 @@ import com.vaadin.server.EncodeResult;
 import com.vaadin.server.Extension;
 import com.vaadin.server.JsonCodec;
 import com.vaadin.server.SerializableComparator;
-import com.vaadin.server.SerializableFunction;
 import com.vaadin.server.SerializableSupplier;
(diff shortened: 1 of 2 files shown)


pleku:
Reviewed 2 of 2 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8983#-KgZO4Bwouhqx4AI86Cd:-KgZO4Bwouhqx4AI86Ce:b-4rayur) (no related file):*
To my understanding, this is not a breaking change, since nobody should have been able to use this with something else than `ValueProvider` or `SerializableFunction<T, ? extends V>` as the returned type. In the former case they just would have an unnecessary cast.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8983)*

Legioth:
This change is not backwards compatible – it breaks binary compatibility by changing the return type. The change is, however, still source compatible.

Issue #8416, opened 2017-02-01
Allow usage of valid 'html location' in CustomLayout


Knoobie:
Currently it's only possible to add a component with the attribute location in the html layout. 
Location isn't a valid html attribut on most elements. 

> ( Example: ```html<div location="id"></div>``` )

How about adding additional support for data-location, so the W3C Markup Validation won't return unnaccessary errors? Additional it's an improvement for the german accessibility guideline, where a valid W3C Markup is a required (*).

* to reduce unnecessary points deduction - see http://testen.bitv-test.de/index.php?a=di&iid=27&s=n

taurus227:
For the record, this is the relevant HTML5 spec section:
[Embedding custom non-visible data with the data-* attributes](https://www.w3.org/TR/html5/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes)

PR #8990, opened 2017-03-31
Update layout-customlayout.asciidoc

silvioincalza:
change attribute name from data-location to location

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8990)


hesara:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/8990#-KgmE9qzEZ6o7HCLYtjz:-KgmE9qzEZ6o7HCLYtk-:b11khfa) (no related file):*
Note that since 8.0.4 and 8.1.0.alpha2, both the more standards compliant "data-location" and the old "location" attribute should work (see #8416). If we want to change the documentation to reflect the situation in the first 8.0 releases, we should probably also qualify the sentence before that with version information.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/8990)*

hesara:
The latest released versions 8.0.4+ and 8.1.0.alpha2+ both support the new syntax, which is more standards compliant. As the documentation text mentions that both are supported and the example is using the new recommended syntax, closing this PR.

PR #9092, opened 2017-04-17
Fix problem when reading a nested Design based on a GridLayout

Artur-:
The parent Design must leave child handling to the nested design, when
there are no children defined in the parent Design. This fixes a problem
with com.vaadin.ui.GridLayout$OutOfBoundsException when the reading
operation of the parent design tries to set number of rows according to
its empty content.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9092)


hesara:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9092)*

PR #9131, opened 2017-04-21
add necessary override and implementations (FocusNotifier & BlurNotifier)

knoobie:
Fix the missing Implementation (FieldEvents.FocusNotifier & FieldEvents.BlurNotifier)  in AbstractTextField. The fields TextField, TextArea and PasswordField are the only focusable input fields where this Implementation is missing.

Update (1):

Well, okay. My intention was to create another pull request for the new commit. But now it's all here. Sorry about that.

- Removes X-UA-Compatible because since Vaadin 8.0 you only support IE > 11 / Edge, so this is not needed anymore
- The chrome=1 project part of the X-UA-Compatible was never updated since 2014 
- Best practice would be (if needed by a client) that he should add it in a server config.

Update (2):

- I would propose to remove the body scroll="auto" part as well, because it adds unnecessary validation errors to the site and adds little to no value for the end user, what do you think?

Update (3):

- Fixed #3944 
- Tested this fix on a local machine against a Firefox and Firefox Dev Edition, both works now ( The FocusListener get's called even when the User clicks the Label or Input to change the Value from true to false or vice versa)
- I didn't want to remove the whole if-block to reduce the chance that this could break something in IE (Don't have one to test it.)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9131)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VCheckBox.java
+++ b/client/src/main/java/com/vaadin/client/ui/VCheckBox.java
@@ -55,7 +55,7 @@ public class VCheckBox extends com.google.gwt.user.client.ui.CheckBox
             el = DOM.getNextSibling(el);
         }
 
-        if (BrowserInfo.get().isWebkit()) {
+        if (BrowserInfo.get().isWebkit() || BrowserInfo.get().isFirefox()) {
             // Webkit does not focus non-text input elements on click
             // (#11854)
             addClickHandler(new ClickHandler() {
--- a/server/src/main/java/com/vaadin/server/BootstrapHandler.java
+++ b/server/src/main/java/com/vaadin/server/BootstrapHandler.java
@@ -387,12 +387,6 @@ public abstract class BootstrapHandler extends SynchronizedRequestHandler {
         head.appendElement("meta").attr("http-equiv", "Content-Type").attr(
                 "content", ApplicationConstants.CONTENT_TYPE_TEXT_HTML_UTF_8);
 
-        /*
-         * Enable Chrome Frame in all versions of IE if installed.
-         */
-        head.appendElement("meta").attr("http-equiv", "X-UA-Compatible")
-                .attr("content", "IE=11;chrome=1");
-
         Class<? extends UI> uiClass = context.getUIClass();
 
         String viewportContent = null;
--- a/server/src/main/java/com/vaadin/ui/AbstractTextField.java
+++ b/server/src/main/java/com/vaadin/ui/AbstractTextField.java
@@ -22,6 +22,7 @@ import java.util.Objects;
 import org.jsoup.nodes.Attributes;
 import org.jsoup.nodes.Element;
 
+import com.vaadin.event.FieldEvents;
 import com.vaadin.event.FieldEvents.BlurEvent;
 import com.vaadin.event.FieldEvents.BlurListener;
 import com.vaadin.event.FieldEvents.FocusEvent;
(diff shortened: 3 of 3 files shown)


Knoobie:
I would love to see, that least, the first commit in 8.1 so that we can use the official vaadin branch again in our software. 

I can remove the other changes if you think this are breaking changes and shouldn't be done in a .x version. 

ping @hesara

hesara:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[server/src/main/java/com/vaadin/ui/AbstractTextField.java, line 48 at r1](https://reviewable.io:443/reviews/vaadin/framework/9131#-KjbvQambIn3FD941RPM:-KjbvQambIn3FD941RPN:b-ljgyfd) ([raw file](https://github.com/vaadin/framework/blob/d4d27fdbf8b90b46997bd4d5944bcb5cfbd5ce2d/server/src/main/java/com/vaadin/ui/AbstractTextField.java#L48)):*
> ```Java
>  */
> public abstract class AbstractTextField extends AbstractField<String>
>         implements HasValueChangeMode, FieldEvents.FocusNotifier, FieldEvents.BlurNotifier {
> ```

@ahie was there some special reason why ATF was not extending these? I guess all text fields should be focusable even if some other fields might not be.

---

*[server/src/main/java/com/vaadin/ui/AbstractTextField.java, line 205 at r1](https://reviewable.io:443/reviews/vaadin/framework/9131#-KjbrSrXUPnJrndaHfK9:-KjbrSrXUPnJrndaHfKA:bs6wss) ([raw file](https://github.com/vaadin/framework/blob/d4d27fdbf8b90b46997bd4d5944bcb5cfbd5ce2d/server/src/main/java/com/vaadin/ui/AbstractTextField.java#L205)):*
> ```Java
>      * @see Registration
>      */
>     @Override 
> ```

extra trailing whitespace

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9131)*

hesara:
See the comments from reviewable.

I'd suggest splitting this to separate PRs with the relevant justifications (currently in PR description) split to the commit messages (just cherry pick each change to a new branch, update the commit message, push the branch and make a PR) - that improves the chances of them going through as unrelated changes won't block each other. If all of these would have been good as-is, I could have merged them as a batch without squashing them, but that rarely happens.

ahie:

---

*[server/src/main/java/com/vaadin/ui/AbstractTextField.java, line 48 at r1](https://reviewable.io:443/reviews/vaadin/framework/9131#-KjbvQambIn3FD941RPM:-Kjc2-3d-zj2scoVaBtE:bh5mp4o) ([raw file](https://github.com/vaadin/framework/blob/d4d27fdbf8b90b46997bd4d5944bcb5cfbd5ce2d/server/src/main/java/com/vaadin/ui/AbstractTextField.java#L48)):*

@ahie was there some special reason why ATF was not extending these? I guess all text fields should be focusable even if some other fields might not be.
</blockquote></details>

Not that I can recall. The original vaadin 8 ATF was implemented before fully moving to the `Registration` based interfaces. Apparently this was neglected along the way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9131)*

Knoobie:
Thank you, @hesara for the feedback. I recreated three PRs for you to easily review the changes. I will close this PR.

PR #9138, opened 2017-04-24
Detect Chrome on iOS as Chrome + Webkit

Artur-:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9138)

### DIFF
--- a/client/src/main/java/com/vaadin/client/BrowserInfo.java
+++ b/client/src/main/java/com/vaadin/client/BrowserInfo.java
@@ -253,6 +253,10 @@ public class BrowserInfo {
         return browserDetails.isSafari();
     }
 
+    public boolean isSafariOrIOS() {
+        return browserDetails.isSafariOrIOS();
+    }
+
     @Deprecated
     public boolean isIE8() {
         return isIE() && getBrowserMajorVersion() == 8;
--- a/client/src/main/java/com/vaadin/client/ResourceLoader.java
+++ b/client/src/main/java/com/vaadin/client/ResourceLoader.java
@@ -329,7 +329,7 @@ public class ResourceLoader {
             linkElement.setType("text/css");
             linkElement.setHref(url);
 
-            if (BrowserInfo.get().isSafari()) {
+            if (BrowserInfo.get().isSafariOrIOS()) {
                 // Safari doesn't fire any events for link elements
                 // See http://www.phpied.com/when-is-a-stylesheet-really-loaded/
                 Scheduler.get().scheduleFixedPeriod(new RepeatingCommand() {
--- a/client/src/main/java/com/vaadin/client/WidgetUtil.java
+++ b/client/src/main/java/com/vaadin/client/WidgetUtil.java
@@ -496,7 +496,7 @@ public class WidgetUtil {
                     // updated when collapsing/expanding columns
                     // Also appeared in Safari 5.1 with webkit 534 (#7667)
                     if ((BrowserInfo.get().isChrome() || (BrowserInfo.get()
-                            .isSafari()
+                            .isSafariOrIOS()
                             && BrowserInfo.get().getWebkitVersion() >= 534))
                             && (scrollleft > 0 || elem.getScrollLeft() > 0)) {
                         int scrollvalue = scrollleft;
@@ -794,7 +794,7 @@ public class WidgetUtil {
             com.google.gwt.dom.client.Element el, String p)
     /*-{
         try {
-    
+
         if (el.currentStyle) {
             // IE
             return el.currentStyle[p];
@@ -809,7 +809,7 @@ public class WidgetUtil {
         } catch (e) {
             return "";
         }
-    
+
      }-*/;
 
     /**
@@ -823,7 +823,7 @@ public class WidgetUtil {
         try {
             el.focus();
         } catch (e) {
-    
+
         }
     }-*/;
 
@@ -1119,7 +1119,7 @@ public class WidgetUtil {
        if ($wnd.document.activeElement) {
            return $wnd.document.activeElement;
        }
-    
+
        return null;
      }-*/;
 
@@ -1190,11 +1190,11 @@ public class WidgetUtil {
     /*-{
         var top = elem.offsetTop;
         var height = elem.offsetHeight;
-    
+
         if (elem.parentNode != elem.offsetParent) {
           top -= elem.parentNode.offsetTop;
         }
-    
+
         var cur = elem.parentNode;
         while (cur && (cur.nodeType == 1)) {
           if (top < cur.scrollTop) {
@@ -1203,12 +1203,12 @@ public class WidgetUtil {
           if (top + height > cur.scrollTop + cur.clientHeight) {
             cur.scrollTop = (top + height) - cur.clientHeight;
           }
-    
+
           var offsetTop = cur.offsetTop;
           if (cur.parentNode != cur.offsetParent) {
             offsetTop -= cur.parentNode.offsetTop;
           }
-    
+
           top += offsetTop - cur.scrollTop;
           cur = cur.parentNode;
         }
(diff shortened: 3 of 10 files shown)


hesara:
Reviewed 1 of 2 files at r1.

---

*[shared/src/main/java/com/vaadin/shared/VBrowserDetails.java, line 83 at r1](https://reviewable.io:443/reviews/vaadin/framework/9138#-KiY_CXlxkNnb5ADDPgM:-KiY_CXlxkNnb5ADDPgN:bp8tvn3) ([raw file](https://github.com/vaadin/framework/blob/a93b3029b9d9e5cfd52eb7c661fb9a5a5c32bab7/shared/src/main/java/com/vaadin/shared/VBrowserDetails.java#L83)):*
> ```Java
>         // browser name
>         isChrome = userAgent.indexOf(" chrome/") != -1
>                 || userAgent.indexOf(" crios/") != -1;
> ```

I'm not 100% convinced whether we should treat Chrome on iOS as Chrome or as something different, as AFAIK it still uses the Safari rendering engine (possibly with some differences in what features are enabled) - to be discussed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

Artur-:

---

*[shared/src/main/java/com/vaadin/shared/VBrowserDetails.java, line 83 at r1](https://reviewable.io:443/reviews/vaadin/framework/9138#-KiY_CXlxkNnb5ADDPgM:-KiY_hYhUlqloVrFWtsN:bxoa6ov) ([raw file](https://github.com/vaadin/framework/blob/a93b3029b9d9e5cfd52eb7c661fb9a5a5c32bab7/shared/src/main/java/com/vaadin/shared/VBrowserDetails.java#L83)):*

I'm not 100% convinced whether we should treat Chrome on iOS as Chrome or as something different, as AFAIK it still uses the Safari rendering engine (possibly with some differences in what features are enabled) - to be discussed
</blockquote></details>

The options were as I saw them: Detect as Safari with correct version or detect as Chrome with correct version number. The user agent does not contain the Safari version number so it was quite easy to exclude that option. With this approach you can detect that it is Chrome on iOS by checking `isChrome() && isIOS()`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

Artur-:

---

*[shared/src/main/java/com/vaadin/shared/VBrowserDetails.java, line 83 at r1](https://reviewable.io:443/reviews/vaadin/framework/9138#-KiY_CXlxkNnb5ADDPgM:-KiYaZSBFuXYC-0Llv1Y:b-d6yqpq) ([raw file](https://github.com/vaadin/framework/blob/a93b3029b9d9e5cfd52eb7c661fb9a5a5c32bab7/shared/src/main/java/com/vaadin/shared/VBrowserDetails.java#L83)):*

The options were as I saw them: Detect as Safari with correct version or detect as Chrome with correct version number. The user agent does not contain the Safari version number so it was quite easy to exclude that option. With this approach you can detect that it is Chrome on iOS by checking `isChrome() && isIOS()`
</blockquote></details>

The real world problem is that right now,`isES6Supported()` returns `false` for Chrome on iOS even though it should return `true` because Chrome on iOS is detected as Safari 5

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

hesara:

---

*[shared/src/main/java/com/vaadin/shared/VBrowserDetails.java, line 83 at r1](https://reviewable.io:443/reviews/vaadin/framework/9138#-KiY_CXlxkNnb5ADDPgM:-KiYamltSdki121DVWC7:b-7nlejh) ([raw file](https://github.com/vaadin/framework/blob/a93b3029b9d9e5cfd52eb7c661fb9a5a5c32bab7/shared/src/main/java/com/vaadin/shared/VBrowserDetails.java#L83)):*

The real world problem is that right now,`isES6Supported()` returns `false` for Chrome on iOS even though it should return `true` because Chrome on iOS is detected as Safari 5
</blockquote></details>

My concern is the dozen or so direct or indirect references to isChrome(), and whether they make sense after this change (and to a lesser extent, future such). However, the number seems to be low enough that we could check each to see if it needs to be adjusted e.g. with !isIOS() or not. Making this a non-blocker, but checking the calls to VBrowserDetails/BrowserInfo/WebBrowser.isChrome() would be good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

Artur-:
Review status: 1 of 2 files reviewed at latest revision, all discussions resolved.

---

*[shared/src/main/java/com/vaadin/shared/VBrowserDetails.java, line 83 at r1](https://reviewable.io:443/reviews/vaadin/framework/9138#-KiY_CXlxkNnb5ADDPgM:-KiZXd8VtWi-6oOXErHx:btnixmn) ([raw file](https://github.com/vaadin/framework/blob/a93b3029b9d9e5cfd52eb7c661fb9a5a5c32bab7/shared/src/main/java/com/vaadin/shared/VBrowserDetails.java#L83)):*

My concern is the dozen or so direct or indirect references to isChrome(), and whether they make sense after this change (and to a lesser extent, future such). However, the number seems to be low enough that we could check each to see if it needs to be adjusted e.g. with !isIOS() or not. Making this a non-blocker, but checking the calls to VBrowserDetails/BrowserInfo/WebBrowser.isChrome() would be good.
</blockquote></details>

I think there is an opposite problem: there are a few places where `ìsSafari` is used and those must use `isSafari() || isIOS()`. I will make some more changes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

Artur-:
Review status: 1 of 10 files reviewed at latest revision, all discussions resolved.

---

*[shared/src/main/java/com/vaadin/shared/VBrowserDetails.java, line 83 at r1](https://reviewable.io:443/reviews/vaadin/framework/9138#-KiY_CXlxkNnb5ADDPgM:-KiZitkumVOMeJgLS5Bx:buajkhk) ([raw file](https://github.com/vaadin/framework/blob/a93b3029b9d9e5cfd52eb7c661fb9a5a5c32bab7/shared/src/main/java/com/vaadin/shared/VBrowserDetails.java#L83)):*

I think there is an opposite problem: there are a few places where `ìsSafari` is used and those must use `isSafari() || isIOS()`. I will make some more changes
</blockquote></details>

Done. Should really use `isWebkit()` for Webkit based browsers but "funnily" enough Chrome is also detected as `isWebkit` as it reports that in the user agent...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

hesara:
Reviewed 9 of 9 files at r2.

---

*[client/src/main/java/com/vaadin/client/ui/VButton.java, line 178 at r2](https://reviewable.io:443/reviews/vaadin/framework/9138#-Ki_1xS0YK_5WvicJsIx:-Ki_1xS0YK_5WvicJsIy:b5av874) ([raw file](https://github.com/vaadin/framework/blob/184327fcacc032cfa3563a7e42444d444f422417/client/src/main/java/com/vaadin/client/ui/VButton.java#L178)):*
> ```Java
>             boolean isPhantomClickPossible = browserInfo.isSafariOrIOS()
>                     && browserInfo.isTouchDevice()
>                     && browserInfo.getBrowserMajorVersion() == 8;
> ```

Does this major version check make sense on Chrome? On the other hand, I guess we only support new enough versions of iOS, so not a blocker.

---

*[compatibility-client/src/main/java/com/vaadin/v7/client/ui/VScrollTable.java, line 5601 at r2](https://reviewable.io:443/reviews/vaadin/framework/9138#-Ki_2N0exzGz3iImdJC-:-Ki_2N0fOCDNBF3brDcq:b-63vobn) ([raw file](https://github.com/vaadin/framework/blob/184327fcacc032cfa3563a7e42444d444f422417/compatibility-client/src/main/java/com/vaadin/v7/client/ui/VScrollTable.java#L5601)):*
> ```Java
>                 Style wrapperStyle = cell.getFirstChildElement().getStyle();
>                 int wrapperWidth = width;
>                 if (BrowserInfo.get().isSafariOrIOS()
> ```

(I wonder if this is still an issue with the supported iOS versions?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9138)*

PR #9124, opened 2017-04-20
cherry picks master->8.0

elmot:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9124)

PR #9106, opened 2017-04-19
Add/fill in missing @since tags for 8.1

hesara:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9106)

PR #9147, opened 2017-04-25
Remove extra sinces

elmot:
Depends on #9124
Related to #9106

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9147)


tsuoanttila:
These since tags have been updated to correct version in both 8.0 and master.

PR #9164, opened 2017-04-26
Open VaadinServlet and BootstrapHandler classes for external use

Artur-:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9164)

### DIFF
--- a/server/src/main/java/com/vaadin/server/BootstrapHandler.java
+++ b/server/src/main/java/com/vaadin/server/BootstrapHandler.java
@@ -59,6 +59,7 @@ import elemental.json.JsonObject;
 import elemental.json.impl.JsonUtil;
 
 /**
+ * Handles the initial request to start the application.
  *
  * @author Vaadin Ltd
  * @since 7.0.0
@@ -75,6 +76,9 @@ public abstract class BootstrapHandler extends SynchronizedRequestHandler {
      */
     public static final String IGNORE_RESTART_PARAM = "ignoreRestart";
 
+    /**
+     * Provides context information for the bootstrap process.
+     */
     protected class BootstrapContext implements Serializable {
 
         private final VaadinResponse response;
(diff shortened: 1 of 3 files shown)


hesara:
Reviewed 2 of 2 files at r1.

---

*[server/src/main/java/com/vaadin/server/BootstrapHandler.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/framework/9164#-KihyN8iGcOF4NyaSN-i:-KihyN8iGcOF4NyaSN-j:bpkinx7) ([raw file](https://github.com/vaadin/framework/blob/35bcd31dcd30bbeaecf97a609f2da6868280daac/server/src/main/java/com/vaadin/server/BootstrapHandler.java#L78)):*
> ```Java
>     public static final String IGNORE_RESTART_PARAM = "ignoreRestart";
> 
>     public class BootstrapContext implements Serializable {
> ```

(at least minimal) javadoc when making this public

---

*[server/src/main/java/com/vaadin/server/BootstrapHandler.java, line 189 at r1](https://reviewable.io:443/reviews/vaadin/framework/9164#-KihxMjR7LjMZFmHoJ0X:-KihxMjR7LjMZFmHoJ0Y:b-9p3ifq) ([raw file](https://github.com/vaadin/framework/blob/35bcd31dcd30bbeaecf97a609f2da6868280daac/server/src/main/java/com/vaadin/server/BootstrapHandler.java#L189)):*
> ```Java
>     }
> 
>     public static class BootstrapUriResolver extends VaadinUriResolver {
> ```

As far as I can tell, to be able to actually override the behavior here, one would need to override BootstrapContext.getUriResolver(). However, BootstrapContext is created deep inside BootstrapHandler.synchronizedHandleRequest(), which does not look like it could be overridden in any sensible way as it uses private methods. Perhaps the creation of a BootstrapContext in BootstrapHandler.synchronizedHandleRequest() should also be extracted so that it is possible to use a custom context and thus also a custom URI resolver.

Furthermore, the public classes should have at least minimal javadoc.

---

*[server/src/main/java/com/vaadin/server/BootstrapHandler.java, line 280 at r1](https://reviewable.io:443/reviews/vaadin/framework/9164#-KihyWKKjYkVOFaKI3rN:-KihyWKKjYkVOFaKI3rO:b-wfowd7) ([raw file](https://github.com/vaadin/framework/blob/35bcd31dcd30bbeaecf97a609f2da6868280daac/server/src/main/java/com/vaadin/server/BootstrapHandler.java#L280)):*
> ```Java
>      *         internal protocol
>      */
>     public static String resolveFrontendUrl(VaadinSession session) {
> ```

now as this is static, it is not really possible to override this behavior in any way - see the comment above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9164)*

Artur-:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[server/src/main/java/com/vaadin/server/BootstrapHandler.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/framework/9164#-KihyN8iGcOF4NyaSN-i:-KiiPzXDPBR9iYNrw4J3:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/35bcd31dcd30bbeaecf97a609f2da6868280daac/server/src/main/java/com/vaadin/server/BootstrapHandler.java#L78)):*

(at least minimal) javadoc when making this public
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/server/BootstrapHandler.java, line 189 at r1](https://reviewable.io:443/reviews/vaadin/framework/9164#-KihxMjR7LjMZFmHoJ0X:-Kii2_HoclGej387atqn:bjzg716) ([raw file](https://github.com/vaadin/framework/blob/35bcd31dcd30bbeaecf97a609f2da6868280daac/server/src/main/java/com/vaadin/server/BootstrapHandler.java#L189)):*

As far as I can tell, to be able to actually override the behavior here, one would need to override BootstrapContext.getUriResolver(). However, BootstrapContext is created deep inside BootstrapHandler.synchronizedHandleRequest(), which does not look like it could be overridden in any sensible way as it uses private methods. Perhaps the creation of a BootstrapContext in BootstrapHandler.synchronizedHandleRequest() should also be extracted so that it is possible to use a custom context and thus also a custom URI resolver.

Furthermore, the public classes should have at least minimal javadoc.
</blockquote></details>

This is done only to be able to call `BootstrapContext.getUriResolver()`

---

*[server/src/main/java/com/vaadin/server/BootstrapHandler.java, line 280 at r1](https://reviewable.io:443/reviews/vaadin/framework/9164#-KihyWKKjYkVOFaKI3rN:-KiiCz49G0JGKr8zzE51:b-msaoxf) ([raw file](https://github.com/vaadin/framework/blob/35bcd31dcd30bbeaecf97a609f2da6868280daac/server/src/main/java/com/vaadin/server/BootstrapHandler.java#L280)):*

now as this is static, it is not really possible to override this behavior in any way - see the comment above
</blockquote></details>

But it's possible to call it and possible to override it in the caller, like before

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9164)*

hesara:
Reviewed 2 of 2 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9164)*

Issue #9172, opened 2017-04-27
Warnings printed to console log when grid header is touched


mvysny:
I'm playing with Vaadin 8.1-SNAPSHOT; when I access a very simple Vaadin 8.1-SNAPSHOT based application from Android phone and attach remotely to the Chrome, the following messages are seen in the javascript console:

```
com.vaadin.DefaultWidgetSet-0.js:4343 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080
ueb @ com.vaadin.DefaultWidgetSet-0.js:4343
Xfb @ com.vaadin.DefaultWidgetSet-0.js:4226
Rh @ com.vaadin.DefaultWidgetSet-0.js:2534
Uh @ com.vaadin.DefaultWidgetSet-0.js:5235
(anonymous) @ com.vaadin.DefaultWidgetSet-0.js:3680
com.vaadin.DefaultWidgetSet-0.js:6091 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080
NVc @ com.vaadin.DefaultWidgetSet-0.js:6091
Qeb @ com.vaadin.DefaultWidgetSet-0.js:6083
Ds @ com.vaadin.DefaultWidgetSet-0.js:5791
us @ com.vaadin.DefaultWidgetSet-0.js:5307
Reb @ com.vaadin.DefaultWidgetSet-0.js:5305
ueb @ com.vaadin.DefaultWidgetSet-0.js:4343
Xfb @ com.vaadin.DefaultWidgetSet-0.js:4226
Rh @ com.vaadin.DefaultWidgetSet-0.js:2534
Uh @ com.vaadin.DefaultWidgetSet-0.js:5235
(anonymous) @ com.vaadin.DefaultWidgetSet-0.js:3680
com.vaadin.DefaultWidgetSet-0.js:4343 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080
ueb @ com.vaadin.DefaultWidgetSet-0.js:4343
Xfb @ com.vaadin.DefaultWidgetSet-0.js:4226
Rh @ com.vaadin.DefaultWidgetSet-0.js:2534
Uh @ com.vaadin.DefaultWidgetSet-0.js:5235
(anonymous) @ com.vaadin.DefaultWidgetSet-0.js:3680
[...]

PR #9196, opened 2017-04-28
Fixes #8632: Touching Grid column header on Android tablet does nothing

### LINKED ISSUE #8632: Touching Grid column header on Android tablet does nothing
When you open this page on iPad or Android tablet (for example Nexus 10), touching the "H1" text in the Grid header will do nothing: https://demo.vaadin.com/sampler/#ui/grids-and-trees/grid/features . The expected behavior is that the Grid is sorted according to the touched column. This works properly on the Desktop browser.

One will have to touch the right part of the Grid header cell (the invisible sorting icon), in order for the sorting to be performed. 
However, on my very simple application with vanilla Valo theme and the default widgetset, the Grid sorts properly even when touched on the caption text.

This looks to be some weird (Android/iPad)+Grid+(maybe custom theme) interaction, but I'm at loss what may be happening here. The Sampler's Vaadin version is 7.7.3, but this problem also manifests on the customer project which runs on 7.7.4.

mvysny:
The fix allows both column-reorder handler and column-sort handlers to run
in parallel; on touchend it is decided whether we will perform sort or
reodering.

The patch solves #8632. Upon applying, the patch will enable user to sort on touch screens even with column reordering enabled.
However, the patch has several drawbacks:

1. It introduces a not-so-apparent interaction between HeaderCellDragStartHandler and HeaderDefaultRowEventHandler (TOUCHSTART is now able to
   bubble to HeaderDefaultRowEventHandler since it is no longer marked as handled by HeaderCellDragStartHandler).
2. Its functionality may/may not depend on the fact that starting with Chrome 56, NativePreviewHandler is not allowed to stop propagation of events for passive elements (see #9172 for details).
3. Prior patching, long-click would trigger multisorting after 500ms. However, the patch now enables the column dragging handler to run at the same time with the column sorting handler
   (since it is not clear whether the user wishes to do column drag or sort, we have to watch out for both in parallel).
   The problem is that making the grid to sort+repoll data after 500ms while still dragging will effectively disrupt the dragging process which is highly annoying. Therefore I have delayed the multisort until after the touchend event; this degrades user experience when multisorting but keeps the dragging experience unhindered which I consider to be of higher priority. Yet, I am unsure whether this is acceptable.

If the multisort UX degradation is acceptable, then I would propose to modify this patch to resolve point 1. and merge HeaderCellDragStartHandler and HeaderDefaultRowEventHandler. What do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9196)


tsuoanttila:
Reviewed 1 of 1 files at r1.

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 7752 at r1](https://reviewable.io:443/reviews/vaadin/framework/9196#-KoMyHp7PObhjKiA4Cf2:-KoMyHp7PObhjKiA4Cf3:b-hqvqyi) ([raw file](https://github.com/vaadin/framework/blob/60b205e3e6de30d1112319034939230a60a3bc86/client/src/main/java/com/vaadin/client/widgets/Grid.java#L7752)):*
> ```Java
>                 }
> 
>                 if (rowEventTouchStartingPoint == null) {
> ```

this check is common for two if blocks. Can these be combined?

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 7761 at r1](https://reviewable.io:443/reviews/vaadin/framework/9196#-KoMyFOEx7Xzzi5JQlRq:-KoMyFOEx7Xzzi5JQlRr:b6ve7wn) ([raw file](https://github.com/vaadin/framework/blob/60b205e3e6de30d1112319034939230a60a3bc86/client/src/main/java/com/vaadin/client/widgets/Grid.java#L7761)):*
> ```Java
>             } else if (BrowserEvents.TOUCHCANCEL
>                     .equals(event.getDomEvent().getType())) {
>                 if (event.getDomEvent().getTouches().length() > 1) {
> ```

This check is same for all the if blocks. Can these be combined?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9196)*

mvysny:

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 7752 at r1](https://reviewable.io:443/reviews/vaadin/framework/9196#-KoMyHp7PObhjKiA4Cf2:-KoQX_FgM98VFwAog6kG:b-ikxigp) ([raw file](https://github.com/vaadin/framework/blob/60b205e3e6de30d1112319034939230a60a3bc86/client/src/main/java/com/vaadin/client/widgets/Grid.java#L7752)):*

this check is common for two if blocks. Can these be combined?
</blockquote></details>

Unfortunately probably it can't. Those if branches differ; combining them will probably only make the code more complex.

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 7761 at r1](https://reviewable.io:443/reviews/vaadin/framework/9196#-KoMyFOEx7Xzzi5JQlRq:-KoQY-kyrhHJrVEe_X5K:bpa9bk4) ([raw file](https://github.com/vaadin/framework/blob/60b205e3e6de30d1112319034939230a60a3bc86/client/src/main/java/com/vaadin/client/widgets/Grid.java#L7761)):*

This check is same for all the if blocks. Can these be combined?
</blockquote></details>

Actually the last if block (the BrowserEvent.CLICK) doesn't contain such check. I believe this check was included also in the original code, so I tried to not to touch it in the patch itself.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9196)*

tsuoanttila:
:lgtm: Tested and works like a charm. The Grid.java file would need a major refactor on some of it's if-then-else structures, so let's _not_ do it here now, but later when we are not in such a rush

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9196#-:-Kr0nbqHsR08hgi-wEqo:bky9swp)*

PR #9204, opened 2017-05-01
Wrong method is used

dunand:
binder.bind(p);
should be replace by
binder.bindInstanceFields(p);

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9204)


dunand:
Update to documentation.

PR #9249, opened 2017-05-05
Move TestBench API to the correct package

Artur-:
Fixes #9241

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9249)

PR #9263, opened 2017-05-07
Add methods for setting the date as an ISO value for date fields

Artur-:
Depends on #9249

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9263)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java
+++ b/client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java
@@ -19,12 +19,14 @@ package com.vaadin.client.ui;
 import java.util.Date;
 
 import com.google.gwt.aria.client.Roles;
+import com.google.gwt.dom.client.Element;
 import com.google.gwt.event.dom.client.ChangeEvent;
 import com.google.gwt.event.dom.client.ChangeHandler;
 import com.google.gwt.event.dom.client.DomEvent;
 import com.google.gwt.event.dom.client.KeyCodes;
 import com.google.gwt.event.dom.client.KeyDownEvent;
 import com.google.gwt.event.dom.client.KeyDownHandler;
+import com.google.gwt.i18n.client.DateTimeFormat;
 import com.google.gwt.user.client.ui.TextBox;
 import com.vaadin.client.BrowserInfo;
 import com.vaadin.client.Focusable;
@@ -54,6 +56,8 @@ public abstract class VAbstractTextualDate<R extends Enum<R>>
         HandlesAriaRequired, KeyDownHandler {
 
     private static final String PARSE_ERROR_CLASSNAME = "-parseerror";
+    private static final String ISO_DATE_TIME_PATTERN = "yyyy-MM-dd'T'HH:mm:ss";
+    private static final String ISO_DATE_PATTERN = "yyyy-MM-dd";
 
     /** For internal use only. May be removed or replaced in the future. */
     public final TextBox text;
@@ -78,6 +82,7 @@ public abstract class VAbstractTextualDate<R extends Enum<R>>
             addDomHandler(this, KeyDownEvent.getType());
         }
         add(text);
+        publishJSHelpers(getElement());
     }
 
     /**
@@ -378,9 +383,65 @@ public abstract class VAbstractTextualDate<R extends Enum<R>>
         fireEvent(event);
     }
 
-    @Override
+    /**
+     * Publish methods/properties on the element to be used from JavaScript.
+     *
+     * @since
+     */
+    private native void publishJSHelpers(Element root)
+    /*-{
+        var self = this;
+        root.setISOValue = $entry(function (value) {
+           self.@VAbstractTextualDate::setISODate(*)(value);
+        });
+        root.getISOValue = $entry(function () {
+           return self.@VAbstractTextualDate::getISODate()();
+        });
+    }-*/;
+
+    /**
+     * Sets the value of the date field as a locale independent ISO date
+     * (yyyy-MM-dd'T'HH:mm:ss or yyyy-MM-dd depending on whether this is a date
+     * field or a date and time field).
+     *
+     * @param isoDate
+     *            the date to set in ISO8601 format, or null to clear the date
+     *            value
+     * @since
+     */
     public void setISODate(String isoDate) {
-        super.setISODate(isoDate);
+        if (isoDate == null) {
+            setDate(null);
+        } else {
+            Date date = getIsoFormatter().parse(isoDate);
+            setDate(date);
+        }
         updateDateVariables();
+    }
+
+    /**
+     * Gets the value of the date field as a locale independent ISO date
+     * (yyyy-MM-dd'T'HH:mm:ss or yyyy-MM-dd depending on whether this is a date
+     * field or a date and time field).
+     *
+     * @return the current date in ISO8601 format, or null if no date is set
+     *
+     * @since
+     */
+    public String getISODate() {
+        Date date = getDate();
+        if (date == null) {
+            return null;
+        } else {
+            return getIsoFormatter().format(date);
+        }
+    }
+
+    private DateTimeFormat getIsoFormatter() {
+        if (supportsTime()) {
+            return DateTimeFormat.getFormat(ISO_DATE_TIME_PATTERN);
+        } else {
+            return DateTimeFormat.getFormat(ISO_DATE_PATTERN);
+        }
     }
 }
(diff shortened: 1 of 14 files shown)


Artur-:
Review status: 0 of 32 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9263#-Kjff8usqwJCm--I-Enj:-Kjff8usqwJCm--I-Enk:blufns6) (no related file):*
Might make more sense to provide a getValue/setValue which always use ISO8601 dates instead and handling parsing/formatting internally. Would be easier to use and less fragile

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

Artur-:
Review status: 0 of 14 files reviewed at latest revision, all discussions resolved.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9263#-Kjff8usqwJCm--I-Enj:-KjhJCE4KhsdpRD6wbJQ:bw98qae) (no related file):*

Might make more sense to provide a getValue/setValue which always use ISO8601 dates instead and handling parsing/formatting internally. Would be easier to use and less fragile
</blockquote></details>

Changed to this approach

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

ahie:
Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9263#-Kjff8usqwJCm--I-Enj:-KkQ8Adq91zof8aSVfbo:b2xfafi) (no related file):*

Changed to this approach
</blockquote></details>

As I understand, these additions are to make testing easier. Instead of setting the date and format on the client side component, wouldn't it be possible to just have a getter on the components to get the time in milliseconds from epoch (GMT?) and have helper methods in the testbench element to do comparisons? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

Artur-:
Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9263#-Kjff8usqwJCm--I-Enj:-KkQBV1Rughd4036rfaB:bi43yyy) (no related file):*

As I understand, these additions are to make testing easier. Instead of setting the date and format on the client side component, wouldn't it be possible to just have a getter on the components to get the time in milliseconds from epoch (GMT?) and have helper methods in the testbench element to do comparisons? 
</blockquote></details>

There is one part of this PR which is actually not strictly needed: the addition of `createFormatString` and `setFormatString`. They can be removed if you want to keep the old API.

The relevant part is being able to set the date/time from JS using the published method. Using timestamps instead of the ISO date format would be another way to do it but I don't see the benefits from it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

tsuoanttila:
Reviewed 14 of 14 files at r1.

---

*[client/src/main/java/com/vaadin/client/ui/VDateField.java, line 274 at r1](https://reviewable.io:443/reviews/vaadin/framework/9263#-KmQSmzoozVwQThcicRD:-KmQSmzoozVwQThcicRE:b-mgcym) ([raw file](https://github.com/vaadin/framework/blob/242f724d44e92855249c6e7a660625ca6d4c4251/client/src/main/java/com/vaadin/client/ui/VDateField.java#L274)):*
> ```Java
>         var self = this;
>         root.setISOValue = $entry(function (value) {
>            self.@VDateField::setISODate(*)(value);
> ```

Do these methods properly send the values to the server? Having these mostly for testing purpose kinda requires that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

Artur-:
Review status: 10 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/VDateField.java, line 274 at r1](https://reviewable.io:443/reviews/vaadin/framework/9263#-KmQSmzoozVwQThcicRD:-KmQZM16UC13n8fx6dLb:bu7hcsb) ([raw file](https://github.com/vaadin/framework/blob/242f724d44e92855249c6e7a660625ca6d4c4251/client/src/main/java/com/vaadin/client/ui/VDateField.java#L274)):*

Do these methods properly send the values to the server? Having these mostly for testing purpose kinda requires that.
</blockquote></details>

Expanded test and fixed stuff and things

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

tsuoanttila:
*[client/src/main/java/com/vaadin/client/ui/VDateField.java, line 274 at r1](https://reviewable.io:443/reviews/vaadin/framework/9263#-KmQSmzoozVwQThcicRD:-KmR-p7SBREsrZIb_UK8:b-yxvxip) ([raw file](https://github.com/vaadin/framework/blob/242f724d44e92855249c6e7a660625ca6d4c4251/client/src/main/java/com/vaadin/client/ui/VDateField.java#L274)):*

Expanded test and fixed stuff and things
</blockquote></details>

These helpers are not used by any other element implementation except the ones that are actually VAbstractTextualDate ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

tsuoanttila:
Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

Artur-:
Review status: 12 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/VDateField.java, line 274 at r1](https://reviewable.io:443/reviews/vaadin/framework/9263#-KmQSmzoozVwQThcicRD:-KmR2snx3U18AMgFZo9v:b4ktw0m) ([raw file](https://github.com/vaadin/framework/blob/242f724d44e92855249c6e7a660625ca6d4c4251/client/src/main/java/com/vaadin/client/ui/VDateField.java#L274)):*

These helpers are not used by any other element implementation except the ones that are actually VAbstractTextualDate ?
</blockquote></details>

Yes, that's why I just moved them to `VAbstractTextualDate` (the feature could be but is not implemented for inline date fields)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

tsuoanttila:
Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

tsuoanttila:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9263#-Kjff8usqwJCm--I-Enj:-KmR4Bal-uZaRdQAdszr:bo5hfbz) (no related file):*

There is one part of this PR which is actually not strictly needed: the addition of `createFormatString` and `setFormatString`. They can be removed if you want to keep the old API.

The relevant part is being able to set the date/time from JS using the published method. Using timestamps instead of the ISO date format would be another way to do it but I don't see the benefits from it?
</blockquote></details>

This change seems pretty ok for me. Asking @elmot a second opinion on this one as well. Dismissing @ahie so we can continue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

elmot:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9263#-Kjff8usqwJCm--I-Enj:-KmRFgGyqWiTj6Q1g4-S:bjvwh9k) (no related file):*

This change seems pretty ok for me. Asking @elmot a second opinion on this one as well. Dismissing @ahie so we can continue.
</blockquote></details>

I am not absolutely happy with LocalDate on server side vs Date on client side, and not-so-correct usage of ISO name but I believe it's acceptable.

---

*[client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java, line 59 at r3](https://reviewable.io:443/reviews/vaadin/framework/9263#-KmRDBb1sWCFrmDjQze4:-KmRCgeLibFw0-ubnd1b:b-99j4ns) ([raw file](https://github.com/vaadin/framework/blob/0dbf73533876b60edfcf05d4f33acad971e3c79e/client/src/main/java/com/vaadin/client/ui/VAbstractTextualDate.java#L59)):*
> ```Java
> 
>     private static final String PARSE_ERROR_CLASSNAME = "-parseerror";
>     private static final String ISO_DATE_TIME_PATTERN = "yyyy-MM-dd'T'HH:mm:ss";
> ```

This is not absolutely correct ISO format. It should contain 'Z' suffix for  absolute date in UTC. Since the field is based on Date, it should be like that. But I think we can survive with present version, but it's better not to note ISO standard then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9263)*

PR #9270, opened 2017-05-08
Removes unnecessary X-UA-Compatible

knoobie:
Intention behind the commit:
* remove unnecessary overload
* removes X-UA-Compatible because since Vaadin 8.0 you only support IE > 11 / Edge, so this is not needed anymore
* the chrome=1 project part of the X-UA-Compatible was never updated since 2014
* best practice would be (if needed by a client) that he should add it in a server config.

I would propose to remove the body scroll="auto" part as well, because it adds unnecessary validation errors to the site and adds little to no value for the end user, what do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9270)

### DIFF
--- a/server/src/main/java/com/vaadin/server/BootstrapHandler.java
+++ b/server/src/main/java/com/vaadin/server/BootstrapHandler.java
@@ -515,12 +515,6 @@ public abstract class BootstrapHandler extends SynchronizedRequestHandler {
         head.appendElement("meta").attr("http-equiv", "Content-Type").attr(
                 "content", ApplicationConstants.CONTENT_TYPE_TEXT_HTML_UTF_8);
 
-        /*
-         * Enable Chrome Frame in all versions of IE if installed.
-         */
-        head.appendElement("meta").attr("http-equiv", "X-UA-Compatible")
-                .attr("content", "IE=11;chrome=1");
-
         Class<? extends UI> uiClass = context.getUIClass();
 
         String viewportContent = null;
--- a/server/src/main/java/com/vaadin/server/BootstrapHandler.java
+++ b/server/src/main/java/com/vaadin/server/BootstrapHandler.java
@@ -515,6 +515,9 @@ public abstract class BootstrapHandler extends SynchronizedRequestHandler {
         head.appendElement("meta").attr("http-equiv", "Content-Type").attr(
                 "content", ApplicationConstants.CONTENT_TYPE_TEXT_HTML_UTF_8);
 
+        // Force IE 11 to use IE 11 mode.
+        head.appendElement("meta").attr("http-equiv", "X-UA-Compatible").attr("content", "IE=11");
+
         Class<? extends UI> uiClass = context.getUIClass();
 
         String viewportContent = null;


Knoobie:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9270#-Kjc7jgUv-2doKPPaU6i:-Kjc8PyjpOynku5AaTjL:bialooj) (no related file):*

If you have IE11 in compatibility mode, won't X-UA-Compatible then force it into IE11 mode?
</blockquote></details>

I'm not sure. Only affected browser: IE 11 (Edge won't implement it - https://msdn.microsoft.com/en-us/library/ff955275(v=vs.85).aspx) My intention was to remove legacy code that shouldn't be needed anymore because of the drop of legacy browsers and of course: best practice would be that an apache httpd server serves the right header to the user. Vaadin shouldn't force this. But I can understand if you think that this shouldn't be done in a 8.x release without proper testing and for example a user poll. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9270)*

hesara:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9270#-Kjc7jgUv-2doKPPaU6i:-Kr5nOmv4QesD83NQPKV:bfm5) (no related file):*
<details><summary><i>Previously, knoobie (Knoobie) wrote…</i></summary><blockquote>

I'm not sure. Only affected browser: IE 11 (Edge won't implement it - https://msdn.microsoft.com/en-us/library/ff955275(v=vs.85).aspx) My intention was to remove legacy code that shouldn't be needed anymore because of the drop of legacy browsers and of course: best practice would be that an apache httpd server serves the right header to the user. Vaadin shouldn't force this. But I can understand if you think that this shouldn't be done in a 8.x release without proper testing and for example a user poll. 
</blockquote></details>

Based on some digging, it would seem to me that X-UA-Compatible would force IE11 to use the IE11 mode (see https://msdn.microsoft.com/en-us/library/ff405803(v=vs.85).aspx ) over most compatibility mode settings in the browser, which is the desired behavior for instance in corporate environments that use compatibility mode settings in group policies etc.

Alternatively, the value could be "edge" which does not mean MS Edge but rather the latest supported mode, but it would probably be better to minimize unnecessary changes.

The Chrome Frame part of the line is unnecessary, but I would not remove the X-UA-Compatible line completely unless someone convinces me it serves no purpose.

@knoobie would you prefer to modify the PR to only remove the Chrome Frame part, or shall we just close this PR?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9270)*

knoobie:
@hesara  I've updated the PR to only remove the chrome part. I can understand your concerns about this. It could be kinda drastic and some developer could have been irritated by such a change in a minor/bugfix release withour proper explanation.

hesara:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9270)*

PR #9352, opened 2017-05-17
remove warning on log when using shortcut on disabled connector

### LINKED ISSUE #6951: Actions for hidden components are sent to server
**Originally by @mstahv**

___
Shortcut actions for disabled components are sent to server. The server side framework blocks the call, but causes nasty warnings:

2015-07-09 12:17:43.429  WARN 36713 --- [io-8080-exec-10] com.vaadin.event.ConnectorActionManager  : Ignoring action for disabled connector com.vaadin.ui.Button, caption=Save

___
 Imported from https://dev.vaadin.com/ issue #18449

OlliTietavainenVaadin:
fixes #6951 : "Shortcut actions for disabled components are sent to server. The server side framework blocks the call, but causes nasty warnings"

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9352)

### DIFF
--- a/server/src/main/java/com/vaadin/event/ConnectorActionManager.java
+++ b/server/src/main/java/com/vaadin/event/ConnectorActionManager.java
@@ -20,7 +20,6 @@ import java.util.logging.Logger;
 import com.vaadin.event.Action.Container;
 import com.vaadin.server.ClientConnector;
 import com.vaadin.server.VariableOwner;
-import com.vaadin.server.communication.ServerRpcHandler;
 import com.vaadin.ui.Component;
 
 /**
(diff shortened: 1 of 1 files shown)


hesara:
Reviewed 1 of 1 files at r1.

---

*[server/src/main/java/com/vaadin/event/ConnectorActionManager.java, line 73 at r1](https://reviewable.io:443/reviews/vaadin/framework/9352#-KkPG6vDsibIEqOVkY8M:-KkPG6vDsibIEqOVkY8N:b-mhbr4u) ([raw file](https://github.com/vaadin/framework/blob/4eba6399334434e28ada8a4c25829f3cde456218/server/src/main/java/com/vaadin/event/ConnectorActionManager.java#L73)):*
> ```Java
>     @Override
>     public void handleAction(Action action, Object sender, Object target) {
>         if (!connector.isConnectorEnabled()) {
> ```

(Ideally, we would not even paint actions for disabled connectors, but in practice that would be tricky to do reliably and doesn't remove the need for this change as the server and the client can still have a different picture of the enabled state. I guess this change is sufficient.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9352)*

PR #9366, opened 2017-05-18
Rename findAndFilterDependencies to findDependencies

Legioth:
The old findDependencies method is made private since it's only an
implementation detail and should not be used from outside the class.

At the same time, the new findAndFilterDependencies method is renamed to
findDependencies since there's no longer necessary to have a name that
distinguishes it from the variant that doesn't do any filtering.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9366)


hesara:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9366)*

PR #9387, opened 2017-05-19
Fix menu item selection in tests

hesara:
When the parent span was not used, the size of the wrong span was used
in the calculation of mouse locations for activating a menu item. This
occasionally leads to the wrong submenu being opened on IE as the
cursor passed over the corner of the next item.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9387)


elmot:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9387)*

PR #9400, opened 2017-05-22
Allow setting text and url as data for IE11

wbadam:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9400)

### DIFF
--- a/client/src/main/java/com/vaadin/client/extensions/DragSourceExtensionConnector.java
+++ b/client/src/main/java/com/vaadin/client/extensions/DragSourceExtensionConnector.java
@@ -17,6 +17,7 @@ package com.vaadin.client.extensions;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import com.google.gwt.animation.client.AnimationScheduler;
 import com.google.gwt.dom.client.DataTransfer;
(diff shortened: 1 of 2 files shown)


wbadam:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9400#-KkkfAWCjRCDhESn27h6:-KkkfAWDmKa98zAlK8C9:b-xmtjgy) (no related file):*
TODO: test if it is actually possible to set url as transfer data for IE, I couldn't receive any data other than text so far.

Should be possible according to the second example here https://msdn.microsoft.com/en-us/library/ms536436(v=vs.85).aspx

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9400)*

hesara:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9400#-KkkfAWCjRCDhESn27h6:-Kn-CGYEf-LT2K75xoSX:bj21vb1) (no related file):*
<details><summary><i>Previously, wbadam (Adam Wagner) wrote…</i></summary><blockquote>

TODO: test if it is actually possible to set url as transfer data for IE, I couldn't receive any data other than text so far.

Should be possible according to the second example here https://msdn.microsoft.com/en-us/library/ms536436(v=vs.85).aspx
</blockquote></details>

code looks ok to me, but I haven't tested this 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9400)*

pleku:
Since according to author @wbadam it actually never did not work for IE11, closing this PR for now since no point in adding something that doesn't even work.

PR #9428, opened 2017-05-24
Throw exception if user modifies same binding step twice

### LINKED ISSUE #9427: Builder doesn't throw exception if it's accessed twice
I just spent quite a while wondering why my converter wasn't working:
```
BindingBuilder<Bean, String> builder = binder.forMemberField(field);
builder.withConverter(new StringToDoubleConverter("Double conversion failed"));
builder.bind(s);
```

In hindsight it's glaringly obvious I should have called `bind` to the builder that is returned from `withConverter` instead of the original builder, but it would have been really nice if it had let me know I'm trying to do something stupid.

- Vaadin version: 8.0.6

tsuoanttila:
Adding a converter to a binding step does not mark it to be bound
allowing it to be later re-used leading into unexpected cast exceptions.

Fixes #9427

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9428)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -711,6 +711,9 @@ public class Binder<BEAN> implements Serializable {
             checkUnbound();
             Objects.requireNonNull(converter, "converter cannot be null");
 
+            // Mark this step to be bound to prevent modifying multiple times.
+            bound = true;
+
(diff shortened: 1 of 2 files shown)


elmot:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9428)*

PR #9433, opened 2017-05-25
ListDataProvider#size now uses the offset and limit provided by the query

Darsstar:
It is my understanding that dp.size(someQueryWithOffsetOrLimit) should be equal to dp.fetch(theSameQueryAsBefore).count(). Currently that is not true.
I am willing to create a test, but I'm not sure where I should add it.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9433)

### DIFF
--- a/server/src/main/java/com/vaadin/data/provider/ListDataProvider.java
+++ b/server/src/main/java/com/vaadin/data/provider/ListDataProvider.java
@@ -84,7 +84,7 @@ public class ListDataProvider<T>
 
     @Override
     public int size(Query<T, SerializablePredicate<T>> query) {
-        return (int) getFilteredStream(query).count();
+        return (int) getFilteredStream(query).skip(query.getOffset()).limit(query.getLimit()).count();
     }
 
     private Stream<T> getFilteredStream(


elmot:
Good catch!
Please add the test to `server/src/test/java/com/vaadin/data/provider`

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9433#-:-Kl2Q9Yhuzjd_p6AW7FC:bqrx4xo)*

tsuoanttila:
Since the `DataProvider#size` method is about finding the size of the data set with given filter, it should always contain `offset = 0, limit = Integer.MAX_VALUE`. That said, using those limits in our default implementations would of course be nice, but we can't rely on the application or backend developer to use these limits when he's supposed to return the size of the backend.

Quote from the `DataProvider#size` JavaDoc: `@return the size of the data provider`

elmot:
Partly agree, but the API for the moment does not look consistent. I'd prefer to have the code from my suggestion merged because couple of arithmetic operations do not really matter performance-wise, but make the API more solid.

---

---

*[server/src/main/java/com/vaadin/data/provider/ListDataProvider.java, line 87 at r1](https://reviewable.io:443/reviews/vaadin/framework/9433#-Kl2V3PHAFmXERJlzwHU:-Kl2V3PHAFmXERJlzwHV:bn8p63e) ([raw file](https://github.com/vaadin/framework/blob/d4b12dee44059c83c5ec8758093abe8819c92919/server/src/main/java/com/vaadin/data/provider/ListDataProvider.java#L87)):*
> ```Java
>     @Override
>     public int size(Query<T, SerializablePredicate<T>> query) {
>         return (int) getFilteredStream(query).skip(query.getOffset()).limit(query.getLimit()).count();
> ```

Looks like this might be optimized as`Math.max(0, Math.min(query.getLimit(), getFilteredStream(query).count() - query.getOffset()))`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9433#-:-KlObO01dM8GpQTN-F8i:b-fjlxzx)*

tsuoanttila:
I'm closing this PR based on the discussion in the original issue location tsuoanttila/GridExtensionPack#20

It was concluded that following the limits on a size query is not the thing to do. Thank you @Darsstar for the PR in my add-on and this one in the Framework, even though only one of them got merged!

PR #9443, opened 2017-05-26
Make component iterators unmodifiable

### LINKED ISSUE #9404: Layout iterator.remove() modifies internal list, causing inconsistent state
Verified with 8.0.6.

The iterator() method of AbstractOrderedLayout, CssLayout (and probably others) just calls iterator() on the internal `LinkedList<Component> components`. As a result, calls to iterator.remove() are passed trough, and the component is removed from the layout's internal components list, but w/o doing the other things required to properly remove a component, leaving the layout in an inconsistent state.

Why is this important:
Calling iterator.remove() will (visually) remove the component or not, depending on other circumstances. This means that behaviour might change when something in a completely different part of the application is changed, leading to "interesting" debugging.
Also, the layout is _always_ left in an inconsistent state when iterator.remove() is used.

Example:
```
		setContent(layout);

		Button button = new Button("Remove");
		layout.addComponent(button);
		button.addClickListener(e -> {
			Iterator<Component> it = layout.iterator();
			if (it.hasNext()) {
				final int before = layout.getComponentCount();
				Component removed = it.next();
				it.remove();
				final int after = layout.getComponentCount();

				System.err.println("Count " + before + "->" + after);
				System.err.println("Removed still has parent: " + removed.getParent());

				// You can see the removed component still has state, BUT this
				// will cause repaint for some reason; let's leave it commented
				// for a more clear example:
				// System.err.println("Removed still has alignment (state): " +
				// layout.getComponentAlignment(removed));

			} else {
				System.err.println("Layout is empty - WHAT are you clicking on?");
			}
		});

		button = new Button("Refresh");
		layout.addComponent(button);
		button.addClickListener(e -> {
			layout.markAsDirty();
		});

	}
```

hesara:
Document that component iterators are typically unmodifiable, and make
iterator.remove() throw an exception.

Document the exception of AbstractSplitPanel that actually supports
removing components through the iterator.

Fixes #9404

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9443)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/AbsoluteLayout.java
+++ b/server/src/main/java/com/vaadin/ui/AbsoluteLayout.java
@@ -16,6 +16,7 @@
 package com.vaadin.ui;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -85,7 +86,9 @@ public class AbsoluteLayout extends AbstractLayout
      */
     @Override
     public Iterator<Component> iterator() {
-        return componentToCoordinates.keySet().iterator();
+        return Collections
+                .unmodifiableCollection(componentToCoordinates.keySet())
+                .iterator();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


elmot:
Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9443)*

PR #9467, opened 2017-06-01
Improve logging when bean validation is not available

### LINKED ISSUE #9464: When expression library is missing, BeanValidationBinder will print misleading error message
When the Expression Library is missing, BeanValidationBinder will disregard the exception stating that fact and instead will errorneously print that
```
BeanValidationBinder cannot be used because a JSR-303 Bean Validation implementation not found on the classpath. Use Binder instead
```
This is misleading since I *have* the JSR-303 bean validation api (`javax.validation:validation-api:1.0.0.GA`) on the classpath.

The exception which should be at least logged at error level, and which actually points to the core of the problem:
```
javax.validation.ValidationException: HV000183: Unable to initialize 'javax.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead
```

Adding `org.glassfish.web:javax.el:2.2.6` as a dependency will solve this issue.
Vaadin 8.0.x and 8.1.0.beta1

hesara:
Closes #9464

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9467)

### DIFF
--- a/server/src/main/java/com/vaadin/data/BeanValidationBinder.java
+++ b/server/src/main/java/com/vaadin/data/BeanValidationBinder.java
@@ -54,7 +54,7 @@ public class BeanValidationBinder<BEAN> extends Binder<BEAN> {
             throw new IllegalStateException(
                     BeanValidationBinder.class.getSimpleName()
                             + " cannot be used because a JSR-303 Bean Validation "
-                            + "implementation not found on the classpath. Use "
+                            + "implementation not found on the classpath or could not be initialized. Use "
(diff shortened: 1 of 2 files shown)


elmot:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9467#-:-Kls-iyt5kepRBamJXJM:bnfp4nl)*

PR #9479, opened 2017-06-05
Re(set) calendar time after calculating rollAmount

sableeyed:
by calling calendar.setTime(today);
after rollAmount has been initialized and given a value, we can reset the calendar's internal time back to the correct date. As it is currently implemented; calling calendar.getTime() will result in a wrong date, it returns the first of the month with the correct time.
the function addInitialEvents() is called AFTER the calendar's time has been subtracted from (by calculating the rollAmount), as such this results in an incorrect internal time and the 'Day' button does not switch to the correct day. Instead it shows the first of the month here too.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9479)


hesara:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9479#-:-Kr594AHWplm4-sraSoB:bnfp4nl)*

PR #9508, opened 2017-06-08
Add pluggable connector id generator support

Legioth:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9508)

### DIFF
--- a/server/src/main/java/com/vaadin/server/ConnectorIdGenerationEvent.java
+++ b/server/src/main/java/com/vaadin/server/ConnectorIdGenerationEvent.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2000-2016 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.server;
+
+import java.util.EventObject;
+
+import com.vaadin.shared.Connector;
+
+/**
+ * Event object containing information related to connector id generation.
+ *
+ * @author Vaadin Ltd
+ * @since 8.1
+ */
+public class ConnectorIdGenerationEvent extends EventObject {
+
+    private final VaadinSession session;
+    private final Connector connector;
+
+    /**
+     * Creates a new event for the given session and connector.
+     *
+     * @param session
+     *            the session for which a connector id is needed, not
+     *            <code>null</code>
+     * @param connector
+     *            the connector that should get an id, not <code>null</code>
+     */
+    public ConnectorIdGenerationEvent(VaadinSession session,
+            Connector connector) {
+        super(session.getService());
+
+        assert session != null;
+        assert connector != null;
+
+        this.session = session;
+        this.connector = connector;
+    }
+
+    /**
+     * Gets the session for which connector id is needed.
+     *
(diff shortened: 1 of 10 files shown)


tsuoanttila:
Reviewed 10 of 10 files at r1.

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1511 at r1](https://reviewable.io:443/reviews/vaadin/framework/9508#-KmQ7tqHg--9gfCNLAiB:-KmQ7tqIddGN6fPDf3Cs:b-wuv12s) ([raw file](https://github.com/vaadin/framework/blob/83f112dc07748f007adc0537b89497d94fb89709/server/src/main/java/com/vaadin/server/VaadinService.java#L1511)):*
> ```Java
>      *
>      */
>     protected ConnectorIdGenerator initConenctorIdGenerator(
> ```

how easy it is to override this to select the specific one to use?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9508)*

Legioth:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1511 at r1](https://reviewable.io:443/reviews/vaadin/framework/9508#-KmQ7tqHg--9gfCNLAiB:-KmQYA3nOwympak8343-:brk5bla) ([raw file](https://github.com/vaadin/framework/blob/83f112dc07748f007adc0537b89497d94fb89709/server/src/main/java/com/vaadin/server/VaadinService.java#L1511)):*

how easy it is to override this to select the specific one to use?
</blockquote></details>

As easy or hard as anything else that requires a custom `VaadinService` subclass, i.e. you'd need to define your own `VaadinServlet` subclass with an overridden `createServletService` method.

Do you have any idea on how to simplify this?

The actual implementation of this method would in most such cases be quite straightforward: either return the desired implementation without even looking at the provided list, or alternatively find the right instance from the list by filtering it based on the provider type.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9508)*

tsuoanttila:
*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1511 at r1](https://reviewable.io:443/reviews/vaadin/framework/9508#-KmQ7tqHg--9gfCNLAiB:-KmQZh6JK3vUYqoDZ0VT:b-cb2j2j) ([raw file](https://github.com/vaadin/framework/blob/83f112dc07748f007adc0537b89497d94fb89709/server/src/main/java/com/vaadin/server/VaadinService.java#L1511)):*

As easy or hard as anything else that requires a custom `VaadinService` subclass, i.e. you'd need to define your own `VaadinServlet` subclass with an overridden `createServletService` method.

Do you have any idea on how to simplify this?

The actual implementation of this method would in most such cases be quite straightforward: either return the desired implementation without even looking at the provided list, or alternatively find the right instance from the list by filtering it based on the provider type.
</blockquote></details>

Well one thing that comes to mind is to provide an API to override this with a Function<List<ConnectorIdGenreator>, ConnectorIdGenerator> but that should be done in quite early phase and not sure how to access APIs like that (yet). I guess the implementation as of right now is "good enough".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9508)*

Legioth:

---

*[server/src/main/java/com/vaadin/server/VaadinService.java, line 1511 at r1](https://reviewable.io:443/reviews/vaadin/framework/9508#-KmQ7tqHg--9gfCNLAiB:-KmQ_ttEXww4jMgIyqrV:b55qfag) ([raw file](https://github.com/vaadin/framework/blob/83f112dc07748f007adc0537b89497d94fb89709/server/src/main/java/com/vaadin/server/VaadinService.java#L1511)):*

Well one thing that comes to mind is to provide an API to override this with a Function<List<ConnectorIdGenreator>, ConnectorIdGenerator> but that should be done in quite early phase and not sure how to access APIs like that (yet). I guess the implementation as of right now is "good enough".
</blockquote></details>

The question would still be what kind of mechanism the application developer has to provide that `Function` implementation. Whatever we come up with for that question, it can probably be implemented later on, based on the current low-level implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9508)*

PR #9532, opened 2017-06-14
Add internal state validators for HasValue

tsuoanttila:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9532)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -1190,7 +1190,8 @@ public class Binder<BEAN> implements Serializable {
         getStatusLabel().ifPresent(label -> label.setValue(""));
 
         return createBinding(field, createNullRepresentationAdapter(field),
-                this::handleValidationStatus);
+                this::handleValidationStatus)
+                        .withValidator(field.getDefaultValidator());
     }
 
     /**
--- a/server/src/main/java/com/vaadin/data/HasValue.java
+++ b/server/src/main/java/com/vaadin/data/HasValue.java
@@ -224,7 +224,7 @@ public interface HasValue<V> extends Serializable {
      * Returns the current value of this object, wrapped in an {@code Optional}.
      * <p>
      * The {@code Optional} will be empty if the value is {@code null} or
-     * {@code isEmpty()} returns {@code true}. 
+     * {@code isEmpty()} returns {@code true}.
      *
      * @return the current value, wrapped in an {@code Optional}
      */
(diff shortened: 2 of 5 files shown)


elmot:
Reviewed 4 of 5 files at r1.
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/HasValue.java, line 301 at r1](https://reviewable.io:443/reviews/vaadin/framework/9532#-Kn-aBHCGiPju0GR27Rf:-Kn-aBHCGiPju0GR27Rg:b7yuul3) ([raw file](https://github.com/vaadin/framework/blob/7d2b282e0fa631dff31090fd31e3bc325f9c91dc/server/src/main/java/com/vaadin/data/HasValue.java#L301)):*
> ```Java
> 
>     /**
>      * Returns a validator that checks the internal state of the HasValue. This
> ```

Is the javadoc really clear for an app developer?
What about text like "This method should be overwritten for components with internal value validation like text input for dates. If user text is unparseableas a date, then that should be exposed to `Binder` by using default component validator".

---

*[server/src/main/java/com/vaadin/data/HasValue.java, line 305 at r1](https://reviewable.io:443/reviews/vaadin/framework/9532#-Kn-a2IKCEcd4wUaMWdG:-Kn-a2IL6LRh0nRAkej8:by18c3v) ([raw file](https://github.com/vaadin/framework/blob/7d2b282e0fa631dff31090fd31e3bc325f9c91dc/server/src/main/java/com/vaadin/data/HasValue.java#L305)):*
> ```Java
>      * date. A faulty input from user can be seen as a validation failure.
>      * 
>      * @return internal state validator
> ```

`@since`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9532)*

elmot:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9532)*

anasmi:
Still some problem exists with validation

- Set malformed date --> call a binder validation --> remove date field content-->call a binder validation another time => the validation failed. (Field is set to be in non-required mode)
-Also setting malformed input (abc, for example) --> validating returns error, but no visual representation then setting the right value makes textfield red
visualvalidation

Code to reproduce: 
`	
             
		DateField dateField = new DateField("Date");
		dateField.setParseErrorMessage("Parse error");
		dateField.setDateOutOfRangeMessage("Out of range");
		layout.addComponent(dateField);
		dateField.addValueChangeListener(event -> {
			System.out.println(dateField.getValue());
		});
		Label errorLabel = new Label();
		layout.addComponent(errorLabel);

		Binder<Void> binder = new Binder<>();
		binder.forField(dateField).withStatusLabel(errorLabel).bind(o -> dateField.getEmptyValue(), null);

		Button button = new Button("Validate!");
		button.addClickListener(event1 -> {
			if (binder.validate().isOk()) {
				System.out.println("Correct");
			} else {
				System.out.println(dateField.isEmpty() + "Error!");
			}
		});
		layout.addComponent(button);

		setContent(layout);`

Issue #7964, opened 2016-12-13
Empty rows in table after scrolling


OlliTietavainenVaadin:
This is a resurfacing of an old issue https://dev.vaadin.com/ticket/6160, which I can't seem to find from GitHub. 

The issue from the original ticket:"
When scrolling up and down (right scrollbar) in a table and pagination is activated, the table is not always rendered correctly.
The table contains empty rows.
After refresh or scrolling up again the table is renderded correct again (empty rows disapeared).
"

The original issue was reported for Vaadin 6, but I was able to reproduce it with 7.7.6 with the latest Chrome on Windows 10. 

The test scenario in the attached project contains an artificial Thread.sleep() in a generated column's generateCell methods, which simulates loading data from a backend service. The table's pageLength is set to 25 and cacheRate is 1.

Reproducing the issue not completely reliable, but I can usually cause it in less than a minute of scrolling up and down.

Attached maven test project:
[scroll-bug-test.zip](https://github.com/vaadin/framework/files/648643/scroll-bug-test.zip)

marianavarroit:
The same for me. I reproduced the same issue with lazy queries (for huge amount of date) using Table and BeanItemContainer. Sometimes when scrolling, some cells dissapear: The problem is with generated columns. Columns without generators are shown perfectly!

bednar:
Same issue affect our application with Vaadin 7.7.6. Is there any progress for fix this issue @tsuoanttila ?

mlindfors:
I'm here to keep this particular issue alive as well. A client has a product and their users are hitting this same issue quite often. I did a rather ugly workaround, but a real fix in framework would be preferred. And yes, this is still happening in Vaadin 7.7.9.

Anyway, as far as I can tell, the descriptions in #5030 are quite accurate in describing the root cause. At least I came to similar conclusions separately before actually even finding that ticket. The methods for working around it aren't quite right, though. As far as I can tell, cacheRate has very little effect on the matter.

Things indeed do break in the method removeUnnecessaryRows inside Table's server side class. To me it looks like there's something off in calculating the amount of rows to be removed, if the rows are to be removed from the beginning of the buffer. In certain cases the thing removes all the rows, I got this to happen when pageBufferFirstIndex was 0 and minPageBufferIndex was something above 0. I suppose if the amount of rows to remove was currentlyCachedRowCount - maxBufferSize, it would be a bit more sane. Unless TreeTable uses those in a totally different manner, in which case it would break, again.

[...]

PR #9551, opened 2017-06-16
Fix an issue, which occasionally leads to Table and TreeTable to have…

mlindfors:
… empty rows (#7964)

There's an intermittently happening issue with both Table and TreeTable, which results in row data disappearing.
This change removes a method which is probably a vestigial one from over five years ago and other changes are handling the things the method used to perform. Currently the method removes rows deemed unnecessary from the row buffer. The problem is, those rows are visible to the user and removing causes row contents to be lost.
Also included are manually runnable test cases which demonstrate that this removal actually prevents the issue from happening.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9551)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Table.java
+++ b/server/src/main/java/com/vaadin/ui/Table.java
@@ -1749,9 +1749,6 @@ public class Table extends AbstractSelect implements Action.Container,
         if (rows > 0) {
             pageBufferFirstIndex = firstIndex;
         }
-        if (getPageLength() != 0) {
-            removeUnnecessaryRows();
-        }
 
         setRowCacheInvalidated(true);
         markAsDirty();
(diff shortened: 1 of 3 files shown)


mlindfors:
To verify that this didn't break anything, I actually went and ran through all the manually runnable tests for both Table and TreeTable. As far as I can tell, this change doesn't break anything in those, since the method is practically never even called in such a way that it would actually do anything.

hesara:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9551#-Kn-IaVKH8IdGmGZ7aVs:-Kn-IaVKH8IdGmGZ7aVt:b-u6fchd) (no related file):*
While the removal of visible rows sounds very wrong, I believe the method was there for a reason. Probably partly related to https://dev.vaadin.com/ticket/10470 and partly to reduce/eliminate memory leaks when scrolling back and forth in the table for a long time. Does anything else do such cleanup there?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9551)*

mlindfors:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9551#-Kn-IaVKH8IdGmGZ7aVs:-Kn-OXhzVYMftPqdfTtZ:b-ttxnrz) (no related file):*

While the removal of visible rows sounds very wrong, I believe the method was there for a reason. Probably partly related to https://dev.vaadin.com/ticket/10470 and partly to reduce/eliminate memory leaks when scrolling back and forth in the table for a long time. Does anything else do such cleanup there?
</blockquote></details>

To my understanding, the method removeUnnecessaryRows() was added in https://dev.vaadin.com/ticket/6160 (see changeset https://dev.vaadin.com/changeset/24879/svn which is connected to the ticket). And I do think the descriptions in #5030 are accurate. I believe the method getVisibleCellsNoCache produces a buffer, which contains just the visible rows and no excess, and it's that exact buffer which gets cropped by removeUnnecessaryRows(), resulting in visible rows being emptied. This is my main rationale for suspecting that the whole method has become vestigial and actually being harmful when it does something, which doesn't happen that often to begin with. It probably was there for a reason when it was created over five years ago, but has since become obsolete.

The rarity of this whole issue occurring is a proof of that. You can also just set a breakpoint to the method, after the point where it determines if the cache size is too large for its tastes or not. I ran all the manual tests through in debug mode and a breakpoint there, and it got hit once. With the test cases in this patch proving the issue, it gets hit all the time and rows go missing.

The main worker method it is calling, unregisterComponentsAndPropertiesInRows, gets called from other places as well and apparently with a better timing and more sane parameters, so similar cleaning gets done regardless. And unregisterPropertiesAndComponents is doing cleaning duties as well, and this gets called every time getVisibleCellsNoCache is executed.

Furthermore, as far as I can tell, the changeset in https://dev.vaadin.com/ticket/10470 just changes VScrollTable, it has just been merged together with the more extensive changes produced by the old trac ticket 6160.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9551)*

PR #9604, opened 2017-06-28
Fix invalid index in drop target Grid

pleku:
Was using length - 1 and 0 for getting the row element,
which caused client side exception when dragging over header or footer when the grid had enough rows.
This had no effect on UX, only exceptions thrown. Now using visible range instead.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9604)


hesara:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9604)*

PR #9620, opened 2017-07-02
Fixes #9594 : Use ZoneId rather than ZoneOffset for conversion between model (Date) and presentation of LocalDateTime

### LINKED ISSUE #9594: LocalDateTimeToDateConverter should uses ZoneId#getRules
LocalDateTimeToDateConverter only supports static ZoneOffsets but zone offsets for time zones with DST are dynamic.   

For bugs, please provide at minimum the following information:
- Vaadin version:  8.0.6
- Description of the bug: (see below)
- If possible, minimal reproducible example: (see below)

Will happily offer a pull request.  It's a 3 line fix:

CURRENT CODE BASE HAS:

```
public class LocalDateTimeToDateConverter
        implements Converter<LocalDateTime, Date> {

    private ZoneOffset zoneOffset;

    /**
     * Creates a new converter using the given time zone.
     *
     * @param zoneOffset
     *            the time zone offset to use, not <code>null</code>
     */
    public LocalDateTimeToDateConverter(ZoneOffset zoneOffset) {
        this.zoneOffset = Objects.requireNonNull(zoneOffset,
                "Zone offset cannot be null");
    }

    @Override
    public Result<Date> convertToModel(LocalDateTime localDate,
            ValueContext context) {
        if (localDate == null) {
            return Result.ok(null);
        }

        return Result.ok(Date.from(localDate.toInstant(zoneOffset)));
    }

    @Override
    public LocalDateTime convertToPresentation(Date date,
            ValueContext context) {
        if (date == null) {
            return null;
        }

        return Instant.ofEpochMilli(date.getTime()).atZone(zoneOffset)
                .toLocalDateTime();
    }

}

```
This is easily enhanced - and is a drop-in replacement since ZoneOffset is a subclass of ZoneId with trivial rules that just return the ZoneOffset itself.

```
public class LocalDateTimeToDateConverter
        implements Converter<LocalDateTime, Date> {

    private ZoneId zoneId;

    /**
     * Creates a new converter using the given time zone.
     *
     * @param zoneId
     *            the time zone id to use, not <code>null</code>
     */
    public LocalDateTimeToDateConverter(ZoneId zoneId) {
        this.zoneId = Objects.requireNonNull(zoneId,
                "Zone ID cannot be null");
    }

    @Override
    public Result<Date> convertToModel(LocalDateTime localDate,
            ValueContext context) {
        if (localDate == null) {
            return Result.ok(null);
        }

        return Result.ok(Date.from(localDate.atZone(zoneId).toInstant()));
    }

    @Override
    public LocalDateTime convertToPresentation(Date date,
            ValueContext context) {
        if (date == null) {
            return null;
        }

        return Instant.ofEpochMilli(date.getTime()).atZone(zoneId).toLocalDateTime();
    }

}

```

eepstein:
LocalDateToDateConverter (correctly) uses ZoneId, whereas LocalDateTimeToDateConverter (incorrectly) uses a ZoneOffset. This fix aligns the two Converter implementations and makes the latter one more robust.

A ZoneOffset is a fixed time difference, e.g., +07:00, whereas a time zone - represented by a ZoneId - is more dynamic, including features like Daylight-Savings Time.  A ZoneId returns one or more ZoneOffsets via its ZoneRules method. (ZoneOffsets have trivial ZoneRules that simply return the ZoneOffset.)

Since the date/time being displayed may be from any date on the calendar, the ZoneOffset imposes a negative limitation. Using ZoneId instead gets us past that limitation and allows a more robust set of conversion rules.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9620)

### DIFF
--- a/server/src/main/java/com/vaadin/data/converter/LocalDateTimeToDateConverter.java
+++ b/server/src/main/java/com/vaadin/data/converter/LocalDateTimeToDateConverter.java
@@ -15,17 +15,17 @@
  */
 package com.vaadin.data.converter;
 
-import com.vaadin.data.Converter;
-import com.vaadin.data.Result;
-import com.vaadin.data.ValueContext;
-import com.vaadin.ui.DateTimeField;
-import com.vaadin.ui.InlineDateTimeField;
-
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.Date;
 import java.util.Objects;
+
+import com.vaadin.data.Converter;
+import com.vaadin.data.Result;
+import com.vaadin.data.ValueContext;
+import com.vaadin.ui.DateTimeField;
+import com.vaadin.ui.InlineDateTimeField;
 
 /**
  * A converter that converts between <code>LocalDateTime</code> and
@@ -35,8 +35,7 @@ import java.util.Objects;
  * @author Vaadin Ltd
  * @since 8.0
  */
-public class LocalDateTimeToDateConverter
-    implements Converter<LocalDateTime, Date> {
+public class LocalDateTimeToDateConverter implements Converter<LocalDateTime, Date> {
 
     private ZoneId zoneId;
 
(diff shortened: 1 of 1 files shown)


hesara:
No specific test, but as ZoneId is a superclass of ZoneOffset, we are implicitly testing the functionality that was already there.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9620#-:-Ko6NVUN1qcTGhOYG8Yh:bod8id0)*

eepstein:
Thanks for merging.

Note on unit tests: I can add a ZoneId-specific test case if that's useful.  I hesitated after waiting for an answer to whether this was the best branch or whether a 8.0.x would be better.

PR #9648, opened 2017-07-06
Tidy up the Vaadin OSGi whiteboard component

timothyjward:
The Vaadin OSGi integration component uses Declarative Services, but it does some odd things:
 * It uses the whiteboard service's own bundle context to get hold of the service instance
 * It has an asymmetric get/release for the whiteboard service, which could leak instances over time
 * It releases service instances that it is still actively using

This change tidies up the service lifecycle by delegating the get/release to the Service Component Runtime managing the container (specifically by injecting the service instance). Using this injection also ensures that the Vaadin whiteboard service is obtained using this component's bundle context.

This change also simplifies the code a little by using the reference as the key to track the registrations. Different references for the same service are required to be equal so there is no issue with doing this.

This change does not alter the fact that the whiteboard service's bundle context is used to register the Http Whiteboard servlet, as this may be the intended behaviour. As a result the component should be prepared for an IllegalStateException when unregistering the service whiteboard service, which may already have been unregistered by the OSGi framework if the target bundle is stopping.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9648)

### DIFF
--- a/osgi-integration/pom.xml
+++ b/osgi-integration/pom.xml
@@ -56,7 +56,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.1</version>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java
+++ b/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java
@@ -48,9 +48,9 @@ import com.vaadin.server.VaadinServlet;
  *
  * @since 8.1
  */
-@Component(immediate = true)
+@Component
 public class VaadinServletRegistration {
-    private Map<Long, ServiceRegistration<?>> registeredServlets = Collections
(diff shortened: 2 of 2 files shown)


elmot:
Thanks, @timothyjward , for your contribution.

@mmerruko, please have a look.

mmerruko:
@elmot The PR seems good to me. I'm sure I wouldn't have been able to do as good of a cleanup but it is something that I neglected. I tested with a sample app in Karaf and it runs as expected.

Thanks @timothyjward!

hesara:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9648)*

QNENet:
Timothy Ward responded to a question I asked on the osgi-dev mail list. I thought the particular piece of code looked strange and Tim agreed and he took the initiative to send the PR.
The info in the Vaadin book is very terse and hard to get a grip on.
If we had a working OSGi example that we could run we might be able to offer  further comment.
A video or even the Full Stack Bakers converted to OSGi would be of immense value.
I have proposed a talk about Vaadin and OSGi at the upcoming EclipseCon in Oct. 
https://www.eclipsecon.org/europe2017/session/joy-osgi-and-vaadin-ux
If possible I would like to be able to get my teeth into the Vaadin approach ASAP.

PR #9688, opened 2017-07-18
Fix Grid header merging on init

### LINKED ISSUE #9686: Vaadin 8.1.0 rc1 - Grid HeaderRow setText() after join does not update caption
Hi,

1/ I have a grid and join two columns - and want to assign a new caption - just like the grid documentation example - but I don't create a new row - just use the default one.
The following code worked fine in 7.7.x 

HeaderRow def = grid.getDefaultHeaderRow();
Grid.HeaderCell c1 = def.getCell("col1 name");
Grid.HeaderCell c2 = def.getCell("col2 name");
def.join(c1,c2).setText("joined name here");

2/ Under 8.1.0 rc1, I do something very similar.

HeaderRow def = grid.getDefaultHeaderRow();
HeaderCell c1 = rr.getCell("col1 name");
HeaderCell c2 = rr.getCell("col2 name");
def.join(c1,c2).setText("joined name here");

What happens now is the column keeps the name of the second column ("col2 name") - it
does not have the new "joined name here" label applied.

(If I add the line c2.setCaption("ABC"); above the last/join line, then the ABC label is displayed. Looks like the join sets the label to the caption of the last column and the setText then does nothing. Adding and removing the setText() line does not change the caption.)

Thanks

tsuoanttila:
Fixes #9686

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9688)

### DIFF
--- a/client/src/main/java/com/vaadin/client/connectors/grid/ColumnConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/ColumnConnector.java
@@ -20,6 +20,7 @@ import com.vaadin.client.annotations.OnStateChange;
 import com.vaadin.client.connectors.AbstractRendererConnector;
 import com.vaadin.client.extensions.AbstractExtensionConnector;
 import com.vaadin.client.widgets.Grid.Column;
+import com.vaadin.client.widgets.Grid.HeaderCell;
 import com.vaadin.shared.data.DataCommunicatorConstants;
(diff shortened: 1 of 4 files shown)


hesara:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9688)*

PR #9704, opened 2017-07-20
Unify basic instance creation and related error reporting

Legioth:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9704)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -2261,12 +2261,10 @@ public class Binder<BEAN> implements Serializable {
     private HasValue<?> makeFieldInstance(
             Class<? extends HasValue<?>> fieldClass) {
         try {
-            return fieldClass.newInstance();
-        } catch (InstantiationException | IllegalAccessException e) {
-            throw new IllegalStateException(
-                    String.format("Couldn't create an '%s' type instance",
-                            fieldClass.getName()),
-                    e);
+            return ReflectTools.createInstance(fieldClass);
+        } catch (IllegalArgumentException e) {
+            // Rethrow as the exception type declared for bindInstanceFields
+            throw new IllegalStateException(e);
         }
     }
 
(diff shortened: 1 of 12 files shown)


hesara:
Reviewed 11 of 12 files at r1.
Review status: 11 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/src/main/java/com/vaadin/util/ReflectTools.java, line 273 at r1](https://reviewable.io:443/reviews/vaadin/framework/9704#-KpxatQxOoRWcmO7E9WO:-KpxatQxOoRWcmO7E9WP:bqvfer3) ([raw file](https://github.com/vaadin/framework/blob/c19f30bde4c9104dbcc1ead2475ad73b883547bb/server/src/main/java/com/vaadin/util/ReflectTools.java#L273)):*
> ```Java
>      * @return an instance of the class
>      */
>     public static <T> T createInstance(Class<T> cls) {
> ```

The problem with the current approach (static method calling constructor.newInstance()) is that it is not possible to override it. However, it should be possible to make the creation code possible to register and plug in here from Spring, CDI etc. if we want to support DI so this is a step in the correct direction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9704)*

Legioth:
Review status: 11 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/src/main/java/com/vaadin/util/ReflectTools.java, line 273 at r1](https://reviewable.io:443/reviews/vaadin/framework/9704#-KpxatQxOoRWcmO7E9WO:-KqMVJMkEjb1VNwlStFl:bcifwmr) ([raw file](https://github.com/vaadin/framework/blob/c19f30bde4c9104dbcc1ead2475ad73b883547bb/server/src/main/java/com/vaadin/util/ReflectTools.java#L273)):*

The problem with the current approach (static method calling constructor.newInstance()) is that it is not possible to override it. However, it should be possible to make the creation code possible to register and plug in here from Spring, CDI etc. if we want to support DI so this is a step in the correct direction.
</blockquote></details>

One step at a time. This change doesn't remove any capabilities, while it adds improved and unified error messaging. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9704)*

tsuoanttila:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9704#-:-KqRhX_3OWOMQaDdycRB:bnfp4nl)*

PR #9761, opened 2017-08-01
Fix moving child to first position in tree data

### LINKED ISSUE #9760: TreeData#moveAfterSibling method throws exception when calling with non root item and null sibling
When trying to move a non root item to the first position on its level, an exception is thrown saying that `java.lang.IllegalArgumentException: Items '<item>' and 'null' don't have the same parent`

Item can be moved to first position when it's a root item.

Add condition to handle case when `null` is passed as sibling.

The bug was introduced in the initial version of the feature in https://github.com/vaadin/framework/issues/9674

wbadam:
Resolves  #9760

Added explicit condition for handling null sibling
Added tests

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9761)

### DIFF
--- a/server/src/main/java/com/vaadin/data/TreeData.java
+++ b/server/src/main/java/com/vaadin/data/TreeData.java
@@ -438,7 +438,8 @@ public class TreeData<T> implements Serializable {
      * @param item
      *         the item to be moved
      * @param sibling
-     *         the item after which the moved item will be located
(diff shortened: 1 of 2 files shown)


hesara:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9761)*

PR #9779, opened 2017-08-03
Added GridDragger to get rid of boilerplate code required for drag dr…

sknitelius:
Currently drag drop operations between or within Grids require alot of boiler plate code. 

I have boiled this down to the GridDragger, it allows dragging for reordering within a Grid as well as Drag Drop operations between different Grids.

Usage simple reordering of items within a Grid
`new GridDragger(myGrid);`

Adding extra listeners to drag events:
```
GridDragger gridDragger = new GridDragger(myGrid);
gridDragger.getGridDropTarget().addDropListener(event -> {...});
gridDragger.getGripDragTarget().addStartDragListener(event -> {...});
```
Setting GridDragger to drag between two grids:
`new GridDragger(sourceGrid, targetGrid);`

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9779)


hesara:
Thank you for your contribution! I think this is a really great helper for many users, but it does have some limitations (only works with ListDataProvider at both ends, fixed drag effect and always moving rows, ...). The limitation of only applying to ListDataProvider is hard to get around in a generic way, though it might be possible to expose more API in GridDragger (perhaps in the form of methods to override, or callbacks/lambdas to set) that allows customizing such things.

I'll ask for another opinion in the team (@tsuoanttila, @pleku), but my opinion would be that this would be better published as an add-on than in the framework in its current form, with the current limitations. I suspect there is demand also for other similar helpers for TreeGrid and Tree that could fit in the same add-on.

sknitelius:
Some of the things you mentioned are easy enough to expose.

I am not quite clear on what you mean by moving rows, essentially this mostly applies for the reorder functionality or do you mean adding the option to always add to end?  

I am not quite sure when ListDataProvider becomes an issue,  maybe when dropping to none list structures? 
This could be handled in a multitude of ways (e.g. instanceof ...). 

A higher level API is always limiting in certain regards, the question is which use cases should be covered for it to be accepted in Vaadin-framework.

 I would hate to publish this as `yet-another-addon` since I believe it greatly simplifies the standard use-cases, specialized use-cases could still be implemented using the established `GridDragSource` and `GridDragTarget`.

hesara:
What I mean about ListDataProvider is that the framework tries hard to make all of its functionality available to users of all data providers including lazy ones (AbstractBackEndDataProvider/CallbackDataProvider), not just ListDataProvider or other InMemoryDataProvider instances such as those that get created by setItems(...). This has been one of the main reasons why there is no functionality related to modifying data providers through a component (or any other location than the one where the type of the data provider is known and fixed) in Vaadin Framework 8.

As for my comment about moving, that relates to the two-grid case where there are situations where you might not want to remove items from the source grid even when dragging them to some order in the second one. This is a smaller issue than that of different DataProviders.

sknitelius:
Ok - I see the issue now, without an IDE right now, I would say this can be solved via `instanceof` or callbacks as you mentioned. 

Since the only functionality required is adding new elements and optionally removing elements.

sknitelius:
How about when using the existing constructors it would work by Default against the ListDataProvider, for custom jobs you could define the logic via constructors and setters:

```
public GridDragger(myGrid, DropTargetWriter dropTargetWriter)
public GridDragger(myGrid, DropTargetWriter dropTargetWriter, DragSourceWriter dragSourceWriter)

public GridDragger(myGrid, targetGrid, DropTargetWriter dropTargetWriter)
public GridDragger(myGrid, targetGrid, DropTargetWriter dropTargetWriter, DragSourceWriter dragSourceWriter)

public void setDropTargetWriter(DragTargetWriter dragTargetWriter();
public void setDragSourceWriter(DragSourceWriter dargSourceWriter();
```

So the setup would look something like this:

```
new GridDragger(myGrid, (index, items) -> {
    //custom implementation for adding items.
}, items -> {
   //custom remove source items implementations.
});

new GridDragger(myGrid, targetGrid (index, items) -> {
    //custom implementation for adding items.
}, items -> {
   //custom remove source items implementations.
});
```

sknitelius:
Okay - I have reworked the GridDragger to incorporate callbacks for the source and target grids.

pleku:
Sorry @sknitelius for us being a bit inactive around this. Thank you for this contribution, I do agree that it should be included in the core FW, but we just did not want to postpone 8.1 final even further.

We will put this into the backlog for the next minor and decide in the coming weeks on how we can include this in that release. We are currently working more actively on other parts of the FW, so we will probably get to look at this immediately.

There will be some changes required, do you wish to do those yourself when requested, or we can merge this and then do additional PRs for the changes ?

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9779#-:-KsqmPHP6Asnt6CABjk3:b5qvj7p)*

sknitelius:
@pleku no worries, I am quite aware that its not that trivial and that it has to be checked properly before being it can be included into a release.

tsuoanttila:
Review status: 0 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHC5zElrWD2rIyXtfu:-KtHC5zFGdmUbBRiqdO1:bwomylz) (no related file):*
This change could use a simple test UI where the different ways to use this GridDragger are shown. 

Also there seems to be no way to remove this dragger once it has been set.

---

*[server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java, line 134 at r4](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHAlOa6gGhb7SBNudZ:-KtHAlOa6gGhb7SBNud_:b-8am312) ([raw file](https://github.com/vaadin/framework/blob/f3fe64c0c2e63122a91988bf4297219c1f66233b/server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java#L134)):*
> ```Java
>      */
>     public GridDragger(Grid<T> source, Grid<T> target, DropMode dropMode) {
>         checkAndInitalizeGridWriter(source, target);
> ```

This method is always called before setting the source and target writers. Should the init method content be moved to some protected/private method?

---

*[server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java, line 154 at r4](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHB8daJzUTJh65soo2:-KtHB8daJzUTJh65soo3:b-awwyc0) ([raw file](https://github.com/vaadin/framework/blob/f3fe64c0c2e63122a91988bf4297219c1f66233b/server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java#L154)):*
> ```Java
>     }
> 
>     public void setGridTargetWriter(GridTargetWriter<T> gridTargetWriter) {
> ```

Public methods without javadoc.

---

*[server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java, line 162 at r4](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHBTsvyCk1EmH4Ke34:-KtHBTsvyCk1EmH4Ke35:bo1z85w) ([raw file](https://github.com/vaadin/framework/blob/f3fe64c0c2e63122a91988bf4297219c1f66233b/server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java#L162)):*
> ```Java
>     }
> 
>     public void setGrid(GridDropTargetIndex<T> gridDropTargetIndex) {
> ```

Confusing method name and parameter combination

---

*[server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java, line 187 at r4](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHBZGQfTzBgPWWQ5CN:-KtHBZGQfTzBgPWWQ5CO:b-qok7lz) ([raw file](https://github.com/vaadin/framework/blob/f3fe64c0c2e63122a91988bf4297219c1f66233b/server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java#L187)):*
> ```Java
>      * @param addToEnd add items to end of Grid.
>      */
>     public void setAddToEnd(boolean addToEnd) {
> ```

Should this information be somehow passed to the GridDropTargetIndex interface so custom implementations can take it into account? Or should this always mean that it should return Integer.MAX_VALUE without asking from the method? Also, no getter for this boolean.

---

*[server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java, line 196 at r4](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHBqTKG1ckUxGPbzcw:-KtHBqTKG1ckUxGPbzcx:b-p487ir) ([raw file](https://github.com/vaadin/framework/blob/f3fe64c0c2e63122a91988bf4297219c1f66233b/server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java#L196)):*
> ```Java
>      * @param removeFromSource set to false to keep items in source Grid.
>      */
>     public void removeFromSourceGrid(boolean removeFromSource) {
> ```

The method name sounds like I'm removing this extension from the source grid. Could use a different name

---

*[server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java, line 207 at r4](https://reviewable.io:443/reviews/vaadin/framework/9779#-KtHAPj95420QcuI8INX:-KtHAPjAgq17T1R57WkJ:b-2755fq) ([raw file](https://github.com/vaadin/framework/blob/f3fe64c0c2e63122a91988bf4297219c1f66233b/server/src/main/java/com/vaadin/ui/components/grid/GridDragger.java#L207)):*
> ```Java
>         if (gridSourceWriter == null) {
>             this.gridSourceWriter = (items) -> {
>                 ListDataProvider listDataProvider = (ListDataProvider) source.getDataProvider();
> ```

Unsafe cast. In case of a user using GridDragger together with any other DataProvider,  this will throw a runtime class cast exception. Could use at least a check that the current DataProvider is of the current type before setting the source writer. Same for the other 2 setters here. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9779)*

sknitelius:
Branch has gone missing - resubmitted new pull request #10036.

hesara:
Superceded by #10036.

PR #9789, opened 2017-08-06
Recalculate tree column width when component size changes

wbadam:
Resolves #9788

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9789)

PR #9791, opened 2017-08-07
Implement SimpleManagedLayout for CompositeConnector

tsuoanttila:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9791)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java
@@ -20,8 +20,10 @@ import com.google.gwt.user.client.ui.Label;
 import com.google.gwt.user.client.ui.Widget;
 import com.vaadin.client.ComponentConnector;
 import com.vaadin.client.ConnectorHierarchyChangeEvent;
+import com.vaadin.client.DirectionalManagedLayout;
 import com.vaadin.client.HasComponentsConnector;
 import com.vaadin.client.ui.AbstractHasComponentsConnector;
+import com.vaadin.client.ui.SimpleManagedLayout;
 import com.vaadin.shared.AbstractComponentState;
 import com.vaadin.shared.MouseEventDetails;
 import com.vaadin.shared.ui.Connect;
@@ -35,7 +37,8 @@ import com.vaadin.ui.Composite;
  * @since 8.1
  */
 @Connect(value = Composite.class, loadStyle = LoadStyle.EAGER)
-public class CompositeConnector extends AbstractHasComponentsConnector {
+public class CompositeConnector extends AbstractHasComponentsConnector
+        implements SimpleManagedLayout {
 
     private ComponentConnector childConnector;
 
@@ -97,8 +100,20 @@ public class CompositeConnector extends AbstractHasComponentsConnector {
     }
 
     @Override
-    protected void sendContextClickEvent(MouseEventDetails details, EventTarget eventTarget) {
-        //Do nothing, because Composite is not an actual component, and the event
-        //must be handled in inner components.
+    protected void sendContextClickEvent(MouseEventDetails details,
+            EventTarget eventTarget) {
+        // Do nothing, because Composite is not an actual component, and the
+        // event must be handled in inner components.
+    }
+
+    @Override
+    public void layout() {
+        // Pass on the layout to the appropriate method in child connector
+        if (childConnector instanceof SimpleManagedLayout) {
+            ((SimpleManagedLayout) childConnector).layout();
+        } else if (childConnector instanceof DirectionalManagedLayout) {
+            ((DirectionalManagedLayout) childConnector).layoutVertically();
+            ((DirectionalManagedLayout) childConnector).layoutHorizontally();
+        }
     }
 }
--- a/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java
@@ -112,8 +112,8 @@ public class CompositeConnector extends AbstractHasComponentsConnector
         if (childConnector instanceof SimpleManagedLayout) {
             ((SimpleManagedLayout) childConnector).layout();
         } else if (childConnector instanceof DirectionalManagedLayout) {
-            ((DirectionalManagedLayout) childConnector).layoutVertically();
             ((DirectionalManagedLayout) childConnector).layoutHorizontally();
+            ((DirectionalManagedLayout) childConnector).layoutVertically();
         }
     }
 }
--- a/uitest/src/main/java/com/vaadin/tests/components/tree/TreeOnBrowserResize.java
+++ b/uitest/src/main/java/com/vaadin/tests/components/tree/TreeOnBrowserResize.java
@@ -1,0 +1,24 @@
+package com.vaadin.tests.components.tree;
+
+import com.vaadin.annotations.Widgetset;
+import com.vaadin.server.VaadinRequest;
+import com.vaadin.tests.components.AbstractTestUI;
+import com.vaadin.ui.Tree;
+
+@Widgetset("com.vaadin.DefaultWidgetSet")
+public class TreeOnBrowserResize extends AbstractTestUI {
+
+    @Override
+    protected void setup(VaadinRequest request) {
+        Tree<String> tree = new Tree<>();
+        tree.setItems(
+                "Lorem ipsum dolor sit amet, consectetur adipiscing elit. "
+                        + "Ut a ante congue, dictum elit vitae, mollis justo. "
+                        + "Nunc porttitor, eros et eleifend accumsan, quam dolor venenatis tortor, "
+                        + "in euismod lorem massa quis nisi. In pretium viverra tincidunt. ");
+
+        tree.setSizeFull();
+        addComponent(tree);
+    }
+
+}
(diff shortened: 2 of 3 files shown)


tsuoanttila:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9791#-KqwKb7_Rr_LnNE9jHFJ:-KqwKb7_Rr_LnNE9jHFK:b-veteup) (no related file):*
Related to PR #9789 

This is a bit more generic approach to solving it for any composite that contains a component that is aware of layouting. The sizes of components inside the Composite do not get updated when changing the size of the UI. 

@wbadam, please take a look and test if this fixes your use-case!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

Artur-:

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/framework/9791#-KqwN753Ux2FeMbgMPbx:-KqwN753Ux2FeMbgMPby:b-2og1ku) ([raw file](https://github.com/vaadin/framework/blob/b22c0567813604c7f98045425ef49e422929c23d/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L41)):*
> ```Java
> @Connect(value = Composite.class, loadStyle = LoadStyle.EAGER)
> public class CompositeConnector extends AbstractHasComponentsConnector
>         implements SimpleManagedLayout {
> ```

Wouldn't this mean that ALL composites are suddenly measured by the LayoutManager? What's the performance impact? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

wbadam:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9791#-KqwKb7_Rr_LnNE9jHFJ:-KqwRBu8x82R96Eh1xC-:b-e9u8lw) (no related file):*

Related to PR #9789 

This is a bit more generic approach to solving it for any composite that contains a component that is aware of layouting. The sizes of components inside the Composite do not get updated when changing the size of the UI. 

@wbadam, please take a look and test if this fixes your use-case!

</blockquote></details>

Seems to be working like a dream :+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

tsuoanttila:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/framework/9791#-KqwN753Ux2FeMbgMPbx:-KqwUJnMyjHDUE7jLuzN:b-cmlo1n) ([raw file](https://github.com/vaadin/framework/blob/b22c0567813604c7f98045425ef49e422929c23d/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L41)):*

Wouldn't this mean that ALL composites are suddenly measured by the LayoutManager? What's the performance impact? 
</blockquote></details>

The implementation only checks if the childConnector actually implements a ManagedLayout and calls the appropriate methods. I have seen no change in the layout phase length when testing this with Grid, TreeGrid and VerticalSplitPanel. If you have an idea on how to make a good stress test to force out some comparable numbers, I'm all ears.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

hesara:
Reviewed 2 of 2 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 114 at r2](https://reviewable.io:443/reviews/vaadin/framework/9791#-Kqwqs-YYuaA5r_BfyRV:-Kqwqs-YYuaA5r_BfyRW:b-va95r3) ([raw file](https://github.com/vaadin/framework/blob/008ceca3a50b7d0d5fdc13590dcf1a7530913234/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L114)):*
> ```Java
>         if (childConnector instanceof SimpleManagedLayout) {
>             ((SimpleManagedLayout) childConnector).layout();
>         } else if (childConnector instanceof DirectionalManagedLayout) {
> ```

This is the unclear case for me - is it safe to simply request both layouts in a sequence here? @legioth ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

Legioth:
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/framework/9791#-KqwN753Ux2FeMbgMPbx:-Kr0jnqn8iMZBX8yY4m-:b34ff3u) ([raw file](https://github.com/vaadin/framework/blob/b22c0567813604c7f98045425ef49e422929c23d/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L41)):*

The implementation only checks if the childConnector actually implements a ManagedLayout and calls the appropriate methods. I have seen no change in the layout phase length when testing this with Grid, TreeGrid and VerticalSplitPanel. If you have an idea on how to make a good stress test to force out some comparable numbers, I'm all ears.
</blockquote></details>

The potential for performance problems is that LayoutManager does extra work and measuring to know that it should call `layout()` even if the method itself is a no-op. The stress test would thus be to look at the layout time with lots of components that are not managed layouts, e.g. 100x `Composite<Label>`. The biggest potential for issues would be in a UI that would otherwise not need any measuring (e.g. only using simple components and CssLayout).

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 114 at r2](https://reviewable.io:443/reviews/vaadin/framework/9791#-Kqwqs-YYuaA5r_BfyRV:-Kr0kchE3heBNkdWQnbT:b200tzf) ([raw file](https://github.com/vaadin/framework/blob/008ceca3a50b7d0d5fdc13590dcf1a7530913234/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L114)):*

This is the unclear case for me - is it safe to simply request both layouts in a sequence here? @legioth ?
</blockquote></details>

This approach is probably safe, but causes some redundant work in cases where layout would only have been needed in one direction. If necessary, this can be optimized by exposing a method for checking whether `LayoutManager.currentDependencyTree.layoutQueueInDirection` contains the id of this connector.

The opposite approach of implementing `DirectionalManagedLayout` and delegating to`SimpleManagedLayout.layout()` in both `layoutVertically()` and `layoutHorizontally()` would instead cause `layout()` to be called twice.

One theoretical fix here might also be to call `layoutHorizontally` before `layoutVertically()` since that's the order `LayoutManager.doLayout()` would call them in cases where layout is needed in both directions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

tsuoanttila:
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/framework/9791#-KqwN753Ux2FeMbgMPbx:-Kr4stscZWKBOOFIYNzL:bfuqsus) ([raw file](https://github.com/vaadin/framework/blob/b22c0567813604c7f98045425ef49e422929c23d/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L41)):*

The potential for performance problems is that LayoutManager does extra work and measuring to know that it should call `layout()` even if the method itself is a no-op. The stress test would thus be to look at the layout time with lots of components that are not managed layouts, e.g. 100x `Composite<Label>`. The biggest potential for issues would be in a UI that would otherwise not need any measuring (e.g. only using simple components and CssLayout).
</blockquote></details>

The question of how to make it a stable test that does not depend on the current workload on the testing cluster is still unanswered. I tested the case of multiple DateFields in Composites and a DateField in nested composites. Neither of them showed any significant increase in layout phase time.

---

*[client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java, line 114 at r2](https://reviewable.io:443/reviews/vaadin/framework/9791#-Kqwqs-YYuaA5r_BfyRV:-Kr4stLP3sA6kpYARJwv:bb41hqx) ([raw file](https://github.com/vaadin/framework/blob/008ceca3a50b7d0d5fdc13590dcf1a7530913234/client/src/main/java/com/vaadin/client/ui/composite/CompositeConnector.java#L114)):*

This approach is probably safe, but causes some redundant work in cases where layout would only have been needed in one direction. If necessary, this can be optimized by exposing a method for checking whether `LayoutManager.currentDependencyTree.layoutQueueInDirection` contains the id of this connector.

The opposite approach of implementing `DirectionalManagedLayout` and delegating to`SimpleManagedLayout.layout()` in both `layoutVertically()` and `layoutHorizontally()` would instead cause `layout()` to be called twice.

One theoretical fix here might also be to call `layoutHorizontally` before `layoutVertically()` since that's the order `LayoutManager.doLayout()` would call them in cases where layout is needed in both directions.
</blockquote></details>

Swapped horizontal and vertical layout around.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791)*

hesara:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9791#-:-Kr4ux0Ib-DyTsLhbsyY:bnfp4nl)*

PR #9800, opened 2017-08-09
Fix occasional empty rows in Table and TreeTable

### LINKED ISSUE #5030: Flattened empty rows in table by specifying cache rate and scrolling
**Originally by _cc-msa_**

___
# DescriptionThe issue can be recognized by flattened empty rows inside the table. The test app below contains everything necessary to reproduce the flattened empty rows.

- It is very important that you add custom components into the table cells, therefore I add one Button each row. However, the issue should be reproducible with any component e.g. a CssLayout containing more components.
* Then also set the cache rate to some rational number but not an integer. In the app below I used 1.1 to reproduce the flattened empty rows. With an integer number I could not produce any empty rows, i.e. the cache rate must be some rational number.

```java
public class TableRowsTestUI extends UI {
  @Override
  protected void init(VaadinRequest request) {
    Table table = new Table();
    table.setSizeFull();
    table.setCacheRate(1.1);
    table.addContainerProperty(0, Button.class, null);
    for (int i = 0; i < 100; i++) {
      table.addItem(new Object[] { new Button() }, i);
    }
    setContent(table);
  }
}
```

# Test caseThe following test case can be used to automatically produce the empty rows. The test assumes that you run the app on localhost:8080.
* First the window dimensions are set to a fixed resolution to be sure the issue can be produced on any display. To get the empty rows the test will scroll to the bottom of the table by doing "tablebody.scrollTop = tablebody.scrollHeight". Afterwards the test will scroll by 1000 to the top by subtracting 1000 from the current scrollTop property.
* The Thread.sleep(2000) calls in between are done to keep the test simple, otherwise you can also use WebDriverWait.until. However, it is important that you wait for the table to complete the rendering, i.e. adding or removing cells.
* At the end the test checks the existence of a button inside each row.
```java
@Test
public void testTableRows() throws InterruptedException {
  // given
  driver.manage().window().setSize(new Dimension(1024, 768));
  driver.get("http://localhost:8080");
  ((TestBenchCommands) driver).waitForVaadin();
  final WebElement tablebody = driver.findElement(By.className("v-table-body-wrapper"));
  final By cellcontentsByXpath = By.xpath(".//tr[@class='v-table-row' or @class='v-table-row-odd']");

  // when
  JavascriptExecutor jse = (JavascriptExecutor) driver;
  jse.executeScript("arguments[0].scrollTop = arguments[0].scrollHeight", tablebody);
  Thread.sleep(2000); // for demo purposes, wait.until is the correct way
  jse.executeScript("arguments[0].scrollTop -= 1000", tablebody);
  Thread.sleep(2000);
  // you can scroll again by 1000 to the top on the table body to see the flattened empty rows

  // then
  List<WebElement> rows = tablebody.findElements(cellcontentsByXpath);
  for (WebElement row : rows) {
    try {
      row.findElement(By.className("v-button"));
    } catch (NoSuchElementException e) {
      fail();
    }
  }
}
```

# InvestigationWhy empty rows?

To get an idea why the rows are empty you can have a look at the paintContent method of the class Table. Later on through paintContent the rows will get painted with the method paintRow. Before it paints the cells of one row the paintRow method checks the visibility of each cell using the LegacyCommunicationManager. The visibility will be false for those flattened empty rows because they have no parent, i.e. parent equals null. This behaviour is correct and the question why parent equals null comes up.

This situation leads to the idea that the cells of the wrong rows were removed, hence parent equaling null. The cells will be removed using the method unregisterComponent in the class Table. However, there are two relevant methods which are calling the unregisterComponent method.

1. unregisterPropertiesAndComponents (uPAC)
1. unregisterComponentsAndPropertiesInRows (uCAPIR)

While the method uPAC will be called very often during scrolling, uCAPIR will almost never be called. The only case uCAPIR will be invoked is if I scroll to the top like in the test case and having set the cache rate to a rational number (no integer). Note that you should get the empty rows if and only if the method uCAPIR has been invoked.

What can be done to solve the issue?

There are possible ways to workaround the issue:
1. Do not use a rational number as the cache rate.
1. Overwrite the Table class and uncomment or remove the uCAPIR method call in the method removeUnnecessaryRows. Note here that without calling the uCAPIR method the amount of TR elements in the DOM does not increase indefinitely due to the uPAC method.

___
 Imported from https://dev.vaadin.com/ issue #13534

### LINKED ISSUE #7964: Empty rows in table after scrolling

hesara:
Picked from #9551 in 7.7

There's an intermittently happening issue with both Table and TreeTable,
which results in row data disappearing.
This change removes a method which is probably a vestigial one from over
five years ago and other changes are handling the things the method used
to perform. Currently the method removes rows deemed unnecessary from
the row buffer. The problem is, those rows are visible to the user and
removing causes row contents to be lost.
Also included are manually runnable test cases which demonstrate that
this removal actually prevents the issue from happening.

Fixes #7964
Fixes #5030

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9800)


tsuoanttila:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9800#-:-Kr5f99sP4UjBpg86wVt:bnfp4nl)*

PR #9818, opened 2017-08-14
Mark session as dirty on UI/session access

### LINKED ISSUE #7535: Explicitly set session attribute at the end of UI/session access to support clustering
**Originally by @hesara**

___
Some clustering solutions such as the default Tomcat clustering valve replicate information in a cluster based on HttpSession attributes that are explicitly set during the processing of a request. Therefore, Vaadin should explicitly mark which sessions have been touched so that the clustering solution does not need to try to track those with extra listeners or aggressively replicate everything.

___
 Imported from https://dev.vaadin.com/ issue #19648

hesara:
This helps clustering solutions know when the session attribute needs
to be replicated instead of having to always aggressively replicate
all attributes.

Resolves #7535

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9818)

### DIFF
--- a/server/src/main/java/com/vaadin/server/VaadinSession.java
+++ b/server/src/main/java/com/vaadin/server/VaadinSession.java
@@ -493,6 +493,15 @@ public class VaadinSession implements HttpSessionBindingListener, Serializable {
     }
 
     /**
+     * Stores this VaadinSession in the HTTP or portlet session.
+     *
+     * @since
+     */
+    public void storeInSession() {
+        service.storeSession(this, session);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


hesara:
Picked from https://dev.vaadin.com/review/#/c/13141 - probably needs a thorough review and maybe some testing.

Artur-:
A simpler approach in https://github.com/Artur-/vaadin/commit/ef2341e6570b31d289d8d3f8cd3d0baae6a04b97

hesara:
Closing this in favor of Artur's version (apparently not yet in PR).

PR #9844, opened 2017-08-18
Prevent edit on non editable column

GillHawk:
Currently if user tries to edit a cell that is not editable (the corresponding column is not editable), the edit overlay is shown even if is not possible to edit that cell. This behavior is not desired so is better to prevent edit.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9844)

### DIFF
--- a/client/src/main/java/com/vaadin/client/widgets/Grid.java
+++ b/client/src/main/java/com/vaadin/client/widgets/Grid.java
@@ -1557,6 +1557,12 @@ public class Grid<T> extends ResizeComposite implements HasSelectionHandlers<T>,
                 throw new IllegalStateException(
                         "Cannot edit row: editor is not enabled");
             }
+            
+            // if column is not editable just return
+            Column<?, T> column = grid.getVisibleColumn(columnIndexDOM);
+            if(column != null && !column.isEditable()) {
+                return;
+            }
 
             if (isWorkPending()) {
                 // Request pending a response, don't move try to start another


hesara:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 1560 at r1](https://reviewable.io:443/reviews/vaadin/framework/9844#-KrpJU_ZOWfjGfv9lkwN:-KrpJU_ZOWfjGfv9lkwO:b-8e464o) ([raw file](https://github.com/vaadin/framework/blob/c5283a0c2c4f235a4d60cd39fda4ea40d940e8c6/client/src/main/java/com/vaadin/client/widgets/Grid.java#L1560)):*
> ```Java
>                         "Cannot edit row: editor is not enabled");
>             }
>             
> ```

(extra whitespace on empty line

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 1562 at r1](https://reviewable.io:443/reviews/vaadin/framework/9844#-KrpJWkjhsVrAXKWv-nL:-KrpJWkjhsVrAXKWv-nM:bmwkgt8) ([raw file](https://github.com/vaadin/framework/blob/c5283a0c2c4f235a4d60cd39fda4ea40d940e8c6/client/src/main/java/com/vaadin/client/widgets/Grid.java#L1562)):*
> ```Java
>             
>             // if column is not editable just return
>             Column<?, T> column = grid.getVisibleColumn(columnIndexDOM);
> ```

I believe this will throw an IllegalStateException on -1. According to the javadoc, column -1 is valid input, and it is used by editRow(int).

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 1564 at r1](https://reviewable.io:443/reviews/vaadin/framework/9844#-KrpJsWnB3G0NneCsWn8:-KrpJsWnB3G0NneCsWn9:byyxuva) ([raw file](https://github.com/vaadin/framework/blob/c5283a0c2c4f235a4d60cd39fda4ea40d940e8c6/client/src/main/java/com/vaadin/client/widgets/Grid.java#L1564)):*
> ```Java
>             Column<?, T> column = grid.getVisibleColumn(columnIndexDOM);
>             if(column != null && !column.isEditable()) {
>                 return;
> ```

What impact does this have on the user experience when e.g. opening the editor from the keyboard and then moving it? Might be fine, but I don't know these parts of Grid well enough to be able to tell without some testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9844)*

hesara:
In addition to the review comments, this change broke the test GridEditorBufferedTest.testKeyboardSaveWithHiddenColumn().

(You can ignore the other test failure in the validation build - it is a flaky test which is addressed by another PR.)

GillHawk:
This test might fail because it expects that editor will always open and the closes. With this change the editor will not be opened anymore (fast isEditable check)

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 1562 at r1](https://reviewable.io:443/reviews/vaadin/framework/9844#-KrpJWkjhsVrAXKWv-nL:-KrpPLu-jmZhBsuOUNT1:bcquazw) ([raw file](https://github.com/vaadin/framework/blob/c5283a0c2c4f235a4d60cd39fda4ea40d940e8c6/client/src/main/java/com/vaadin/client/widgets/Grid.java#L1562)):*

I believe this will throw an IllegalStateException on -1. According to the javadoc, column -1 is valid input, and it is used by editRow(int).
</blockquote></details>

indeed, I think here an extracheck should help

---

*[client/src/main/java/com/vaadin/client/widgets/Grid.java, line 1564 at r1](https://reviewable.io:443/reviews/vaadin/framework/9844#-KrpJsWnB3G0NneCsWn8:-KrpO_9ySnf40k4vcIPO:b8h4xi5) ([raw file](https://github.com/vaadin/framework/blob/c5283a0c2c4f235a4d60cd39fda4ea40d940e8c6/client/src/main/java/com/vaadin/client/widgets/Grid.java#L1564)):*

What impact does this have on the user experience when e.g. opening the editor from the keyboard and then moving it? Might be fine, but I don't know these parts of Grid well enough to be able to tell without some testing.
</blockquote></details>

Try this scenario. Create a grid with 3 columns, make the last column editable. Try to edit a cell in the table using the keyboard (press Enter over a non editable cell). Check where is the focus.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9844#-:-KrpPvHinvssJKWtG6T3:bhhnhz8)*

GillHawk:
Abort pool request. I managed to achieve this feature on server side by overriding EditorImpl and now I can control edit behavior on cell level.

PR #9858, opened 2017-08-23
Fix null check in BackEndDataProvider

### LINKED ISSUE #9855: Dead code in com/vaadin/data/provider/BackEndDataProvider.java
Vaadin framework, at least for version 8.1.1, contains bug in framework/server/src/main/java/com/vaadin/data/provider/BackEndDataProvider.java at line 61:
Objects.requireNonNull("Sort builder cannot be null.");
This validation condition is always true. It never protect from wrong parameter value, because "builder" method parameter is omitted, constant non-null string given only.
I propose following change:
        Objects.requireNonNull(builder, "Sort builder cannot be null.");
With regards, ibatis-dao

hesara:
Fixes #9855

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9858)


ahie:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9858)*

PR #9862, opened 2017-08-23
Avoid detaching VL/HL children unnecessarily

### LINKED ISSUE #7713: VerticalLayout and HorizontalLayout reattaches elements when removing other elements
**Originally by @Legioth**

___
The logic in `AbstractOrderedLayoutConnector.onConnectorHierarchyChange` causes redundant attaching and detaching of elements when another element is removed from the layout. If the layout previously contained components a, b, and c, then when b is about to be removed the logic will see that c should be at index 1 but it's currently at index 2 (since b has in fact not yet been removed), so it will detach c from index 2 and insert it into index 1.

The main impact of this issue is a small reduction in performance. It might also cause loss of focus, selection or scroll position in children that do not gracefully handle reattaching, but that is actually a bug in that particular component since it should still also support situations where a reattach is inevitable.

This issue is similar to #7712, but for different components.

___
 Imported from https://dev.vaadin.com/ issue #19959

hesara:
When child components are removed from a
VerticalLayout/HorizontalLayout, do not remove other children from the
DOM if no other hierarchy changes are made.

Fixes #7713

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9862)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/orderedlayout/AbstractOrderedLayoutConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/orderedlayout/AbstractOrderedLayoutConnector.java
@@ -320,6 +320,24 @@ public abstract class AbstractOrderedLayoutConnector
         Profiler.leave(
                 "AOLC.onConnectorHierarchyChange temporarily remove spacing");
 
+        // first remove extra components to avoid extra detaches and attaches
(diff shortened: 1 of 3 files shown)


ahie:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9862)*

PR #9874, opened 2017-08-25
Required fields arn't marked when button with binder.isValid() is triggered

sknitelius:
When someone presses a Button that relies on `binder.isValid()` prior to entering all required values, the action is blocked as expected.

However the fields which haven't been touched aren't marked red.

Current work around is to trigger `binder.validate()` in the button action, this marks all empty required fields as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9874)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -1621,15 +1621,8 @@ public class Binder<BEAN> implements Serializable {
                     + "bean level validators have been configured "
                     + "but no bean is currently set");
         }
-        if (validateBindings().stream().filter(BindingValidationStatus::isError)
-                .findAny().isPresent()) {
-            return false;
-        }
-        if (getBean() != null && validateBean(getBean()).stream()
-                .filter(ValidationResult::isError).findAny().isPresent()) {
-            return false;
-        }
-        return true;
+        BinderValidationStatus<BEAN> validate = validate();
+        return validate.isOk();
     }
 
     /**


ahie:
Hi and thanks for the patch! `isValid` was originally introduced to provide a simple way to check the validation status without causing other side effects like drawing required and validation error indicators, which is what calling `validate` does. The javadocs are currently missing a note about this behavior, which should probably be fixed instead.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9874#-:-KsbcQX68zcC2V70ZeMx:bx6gi3t)*

sknitelius:
Just spotted the failed tests - my bad.

Back to the actual issue, I would argue the standard use case is to check pre-"submit" if all Fields have been filled correctly and hence trigger a redrawing.

Calling:
```
if (binder.validate().isOk()) {
   ....                
}
```
seems rather long winded and unintuitive. 

As opposed to:
```
if (binder.valid()) {
    ...
}
```
Which also happens to be the suggested way of handling validation in actions in the documentation.

Hence I would suggest turning it around and adding an explicit method that performs a side effect free validation.

ahie:
Changing the `validate` method to not fire `StatusChangeEvent`s would unfortunately cause a major breaking change. To get around breaking changes additional methods would need to be introduced. This would likely cause at least some additional confusion with the API and thus result in a worse developer experience overall.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9874#-:-KsmqaCdGJRXHeVtlnpl:b-9u6wqy)*

hesara:
After some discussion, we concluded that Binder currently works as designed. Added a clarification to the javadoc in #9930 and closing this PR.

PR #9910, opened 2017-08-31
Fix ListSelect scrolling behavior on IE

### LINKED ISSUE #9894: VListSelect scrolling behavior changed slightly after refactor (IE11)
**Problem:**
ListSelect component scroll doesn't work correctly with IE11 like before. When selecting e.g second option from list IE11 will also scroll list down to that option ("hiding" now first node). Chrome and Firefox seems to be working correctly. See screenshots.

Faulty: IE11 second element selected by default (7.7.x) and "hiding" first element in list:

Correct: IE11 second element selected by default (7.6.8 and earlier): 

**Fix:**
Adding "temporary hack"-section from VListSelect (line: 92, Vaadin: 7.6.8) restores functionality back it was before. 
Vaadin - VListSelect.java 7.7.0 | 7.6.8

- Vaadin version: 7.7.x+ (possibly also 8.x?)
- Last working version: 7.6.8

ahie:
Fixes #9894

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9910)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VListSelect.java
+++ b/client/src/main/java/com/vaadin/client/ui/VListSelect.java
@@ -70,6 +70,9 @@ public class VListSelect extends VOptionGroupBase {
 
     @Override
     public void buildOptions(UIDL uidl) {
+        int scrollTop = select.getElement().getScrollTop();
+        int rowCount = getRows();
+
         select.setMultipleSelect(isMultiselect());
 
         Set<String> previousKeys = new HashSet<String>();
@@ -110,6 +113,10 @@ public class VListSelect extends VOptionGroupBase {
 
         if (getRows() > 0) {
             select.setVisibleItemCount(getRows());
+        }
+
+        if (rowCount == getRows()) {
+            select.getElement().setScrollTop(scrollTop);
         }
     }
 
--- a/uitest/src/test/java/com/vaadin/tests/components/listselect/ListSelectJumpTest.java
+++ b/uitest/src/test/java/com/vaadin/tests/components/listselect/ListSelectJumpTest.java
@@ -1,0 +1,39 @@
+package com.vaadin.tests.components.listselect;
+
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.openqa.selenium.By;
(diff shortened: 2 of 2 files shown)


hesara:
Reviewed 2 of 2 files at r1.

---

*[client/src/main/java/com/vaadin/client/ui/VListSelect.java, line 118 at r1](https://reviewable.io:443/reviews/vaadin/framework/9910#-KssPPK237tME6AqIXyQ:-KssPPK3IKP3ZZWL-6li:bnw5vlp) ([raw file](https://github.com/vaadin/framework/blob/9a6f7c4a2363db3c3be060edce4429671011fe5e/client/src/main/java/com/vaadin/client/ui/VListSelect.java#L118)):*
> ```Java
>         }
> 
>         if (rowCount == getRows()) {
> ```

This feels a bit hacky, so should we only do this on IE if the other browsers work fine without this?
Should this have some comment about why we do this? That could also implicitly explain why we only do this if the row count hasn't changed.
Should this also be applied to FW8?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9910)*

ahie:

---

*[client/src/main/java/com/vaadin/client/ui/VListSelect.java, line 118 at r1](https://reviewable.io:443/reviews/vaadin/framework/9910#-KssPPK237tME6AqIXyQ:-Kswq5OVbGVPa7t8M32I:bdgbywy) ([raw file](https://github.com/vaadin/framework/blob/9a6f7c4a2363db3c3be060edce4429671011fe5e/client/src/main/java/com/vaadin/client/ui/VListSelect.java#L118)):*

This feels a bit hacky, so should we only do this on IE if the other browsers work fine without this?
Should this have some comment about why we do this? That could also implicitly explain why we only do this if the row count hasn't changed.
Should this also be applied to FW8?
</blockquote></details>

This does not apply to FW8 because the client widget has been rewritten. I don't see a reason to run this code exclusively on IE. It only ensures if the row count remains the same that rebuilding the options doesn't change the location of the scroll.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9910)*

Issue #9859, opened 2017-08-23
Upload throws StringIndexOutOfBoundsException


OlliTietavainenVaadin:
This static method in FileUploadHandler will fail if it reads a zero-length line:

```java
private static String readLine(InputStream stream) throws IOException {
        ByteArrayOutputStream bout = new ByteArrayOutputStream();
        int readByte = stream.read();
        while (readByte != LF) {
            if (readByte == -1) {
                throw new IOException(
                        "The multipart stream ended unexpectedly");
            }
            bout.write(readByte);
            readByte = stream.read();
        }
        byte[] bytes = bout.toByteArray();
        return new String(bytes, 0, bytes.length - 1, UTF8);
    }
```

If bytes.length is 0 on the last line of code, it will throw an exception because it tries to read a range from 0 to -1.

Exception thrown:  
```
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
 at java.lang.String.checkBounds(Unknown Source)
 at java.lang.String.<init>(Unknown Source)
 at com.vaadin.server.communication.FileUploadHandler.readLine(FileUploadHandler.java:312)
 at com.vaadin.server.communication.FileUploadHandler.doHandleSimpleMultipartFileUpload(FileUploadHandler.java:364)
 at com.vaadin.server.communication.FileUploadHandler.handleRequest(FileUploadHandler.java:288)
 at com.vaadin.server.VaadinService.handleRequest(VaadinService.java:1464)
```

- Vaadin version: 8.1.2

hesara:
Do you have a file to upload to reproduce this? Does this happen with empty files, or with files with empty lines, or only under some more complicated circumstances?

hesara:
The line numbers in the stack trace seem to match the version of FileUploadHandler that was in 8.0 and perhaps up to 8.1.0.beta1, but not in 8.1.0.beta2 nor in the 8.1.2 version.

hesara:
When do you see the issue? With an empty file? With a file that has an LF without a preceding CR? With consecutive LFs in the file? Some other case? Do you only see the issue in Liferay, or also in other environments? Which OS etc.?

OlliTietavainenVaadin:
According to the report, the file in question was a randomly generated large file (multi-gigabyte) uploaded in a Liferay portlet. Original test data was created on Windows with `fsutil file createnew dummy 5368709120`. @elmot recreated a similar file as follows:
[...]

PR #9914, opened 2017-08-31
Fix issue #1892: Uploading a large (artifical) file with Vaadin upload (com.vaadin.ui.upload)

robinvanspauwen:
Fixes issue #1892:

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html

Use this constructor:
String(byte[] bytes, Charset charset)
Constructs a new String by decoding the specified array of bytes using the specified charset.

Instead of this one:
String(byte[] bytes, int offset, int length, String charsetName)
Constructs a new String by decoding the specified subarray of bytes using the specified charset.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9914)

### DIFF
--- a/server/src/main/java/com/vaadin/server/communication/FileUploadHandler.java
+++ b/server/src/main/java/com/vaadin/server/communication/FileUploadHandler.java
@@ -315,7 +315,7 @@ public class FileUploadHandler implements RequestHandler {
             readByte = stream.read();
         }
         byte[] bytes = bout.toByteArray();
-        return new String(bytes, 0, bytes.length - 1, UTF8);
+        return new String(bytes, UTF8);
     }
 
     /**


hesara:
Is this for the issue #9859? The number mentioned in the changeset comment seems to be for the support system.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[server/src/main/java/com/vaadin/server/communication/FileUploadHandler.java, line 318 at r1](https://reviewable.io:443/reviews/vaadin/framework/9914#-KsvlaT3f31i0pabwSzm:-KsvlaT3f31i0pabwSzn:b-f28nfz) ([raw file](https://github.com/vaadin/framework/blob/37e6d584f62585ea6d186c62d1dc6b557cdebbb4/server/src/main/java/com/vaadin/server/communication/FileUploadHandler.java#L318)):*
> ```Java
>         }
>         byte[] bytes = bout.toByteArray();
>         return new String(bytes, UTF8);
> ```

Skipping the last byte is (as far as I can tell) because the spec requires a CRLF pair before a boundary. As the code above stops reading at an LF, the CR preceding it needs to be skipped. (Of course, it would be good to have comments mentioning this in the code.)

Based on the stack trace in #9859, there seems to be some case(s) where the client does not send a CRLF pair but just an LF, which breaks this implementation. As far as I can tell (but I might be wrong), all LFs that are a part of the multi-part format itself should be in the form of CRLF pairs. However, I don't know if the spec allows e.g. passing lone LFs in the file contents to the data part.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9914#-:-KsvmftWF8zllNr0qWgf:b2jhp5y)*

hesara:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[server/src/main/java/com/vaadin/server/communication/FileUploadHandler.java, line 318 at r1](https://reviewable.io:443/reviews/vaadin/framework/9914#-KsvlaT3f31i0pabwSzm:-Ksw17vmEKwcXEmTATxG:b-9n3hop) ([raw file](https://github.com/vaadin/framework/blob/37e6d584f62585ea6d186c62d1dc6b557cdebbb4/server/src/main/java/com/vaadin/server/communication/FileUploadHandler.java#L318)):*

Skipping the last byte is (as far as I can tell) because the spec requires a CRLF pair before a boundary. As the code above stops reading at an LF, the CR preceding it needs to be skipped. (Of course, it would be good to have comments mentioning this in the code.)

Based on the stack trace in #9859, there seems to be some case(s) where the client does not send a CRLF pair but just an LF, which breaks this implementation. As far as I can tell (but I might be wrong), all LFs that are a part of the multi-part format itself should be in the form of CRLF pairs. However, I don't know if the spec allows e.g. passing lone LFs in the file contents to the data part.
</blockquote></details>

Continuing related discussion in #9859.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9914)*

hesara:
Unfortunately this change breaks normal file uploads, at least the upload of an empty file in a non-portlet environment. It seems the interaction between doHandleSimpleMultipartFileUpload() and readLine() (and later SimpleMultiPartInputStream) is a bit more complicated, and this breaks the current implementation. 

Furthermore, it is not clear to me how this would fix the original issue #9859, which I did not manage to reproduce even though using files with lone LFs etc. A sample file for reproducing the issue and a specification of the environment used (versions of all relevant components, OS, ...) could help us investigate this further. I suspect the problem isn't with file size but something else.

Closing the PR, but it can be re-opened if you can fix the issues with the patch.

PR #9953, opened 2017-09-08
Update Binder to display updated validation status

jcgueriaud:
Hello,

I have a strange behaviour. I use the binder with readbean and writeBeanIfValid.
doWriteIfValid doesn't show updated field status.
For example, an empty required field is not filled by default then the status is ok.
writeBeanIfValid return false but the field status is not updated on the screen (not in error).

Is it a correct behaviour ? (field statuses are not refreshed after writebeanIfValid) 

[MyUI.java.zip](https://github.com/vaadin/framework/files/1288352/MyUI.java.zip)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9953)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -1462,9 +1462,11 @@ public class Binder<BEAN> implements Serializable {
         // If no validation errors then update bean
         if (bindingStatuses.stream().filter(BindingValidationStatus::isError)
                 .findAny().isPresent()) {
-            fireStatusChangeEvent(true);
-            return new BinderValidationStatus<>(this, bindingStatuses,
+            BinderValidationStatus<BEAN> validationStatus = new BinderValidationStatus<>(this, bindingStatuses,
                     Collections.emptyList());
+            fireStatusChangeEvent(true);
+            getValidationStatusHandler().statusChange(validationStatus);
+            return validationStatus;
         }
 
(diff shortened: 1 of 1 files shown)


jcgueriaud:
I add one comment on my "patch": Binder is a quite hard to understand and to modify.
So I just write something to get the "expected" behaviour which is in my opinon that writeBeanIfValid = validate then writeBean.
So in my opinion, writeBeanIfValid should be rewrite reusing validate function (instead of copy/paste like I do).
I can also call validate function juste after readBean and all the errors will be displayed (before calling writeBean).

Regards,

tsuoanttila:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/9953#-KuOu3Y_NtGGzZczVHBa:-KuOu3Y_NtGGzZczVHBb:b-93shd) (no related file):*
Hi @jcgueriaud !

Thanks for the Pull Request. However, there's another fix to the Binder, that does touch the same part of the Binder and makes more or less the same change. Can you test if the change in #9988 against your UI? If you don't have the time, I'll do it a bit later when I get back to this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/9953)*

jcgueriaud:
Hi @tsuoanttila 
I've just try your modifications and it does fix my problem.

Really nice patch :)
Thanks !

I close my pull request  because it's fixed by #9988

PR #9973, opened 2017-09-12
Fix bean validation not executed with setBean and ValueChange

tsuoanttila:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/9973)


tsuoanttila:
This was an alternative implementation for #9988 and they are more or less mutually exclusive.

tsuoanttila:
Implementation of this patch has been made easier in the #9988.

PR #10007, opened 2017-09-15
Add aria-sort to Grid

knoobie:
This pr adds basic support for aria-sort for Grid (https://www.w3.org/TR/wai-aria-1.1/#aria-sort). 

Simple explanation: 
- aria-sort is not added, if the column is not sortable
- aria-sort="none" is added, if the column is sortable but currently not sorted
- aria-sort="ascending" is added, if the column is sorted by asc
- aria-sort="descending" is added, if the column is sorted by desc
- aria-sort="other" is added, if more than 1 column is sorted, currently there is no aria-sort-order

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10007)


hesara:
:lgtm: assuming that cellElement is the element where screen readers etc. expect to find the attribute.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10007#-:-KuIUeHN470BKaE5g2U-:b-kaqkyr)*

knoobie:
@hesara You can find here (https://www.w3.org/TR/wai-aria-practices/examples/grid/dataGrids.html below Role, Property, State, and Tabindex Attributes) an example with definiton where it should be added.

PR #10031, opened 2017-09-19
Use interface not Implementation, as parameter

asashour:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10031)


hesara:
We had some discussion about these as several of these are in theory breaking changes (subclass overriding the method needs to be updated), but in practice all these are so specific and unlikely to be overridden, that we feel we can take these to 8.2.

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10031#-:-KuTB1wQm7Oa8JpvIWmK:b-ajet01)*

Issue #8820, opened 2017-03-13
Programmatically open the editor in Grid for an item based on the item


pleku:
Challenge is with knowing the index of the item in the grid, since that is not known on the server side. This would require also `scrollToItem(T item)` to work also

heruan:
What if the item is already selected? Can the editor already be opened for the selected item? I guess I'm missing something here, but how `select(T item)` is implemented right now if the key is unknown on the server side?

tsuoanttila:
There is a big difference in showing a selection on a row and showing the editor. Basically the editor wants to scroll to the location of the edited row and display the editor on top of it, when the selection can be done for any object at any time even when it's not in the view port. We don't need to know where in the data set the object is to select it, but we do need to know the location for the object we want to scroll to and edit.

jflamy:
A simpler requirement even is to open the grid in editable mode.  Manytimes, the grid is small and the user is expected to pick a row and edit it.  The current requirement to double-click or use enter is ABSOLUTELY NOT intuitive.

sirbris:
We really need this possibility since our users can add an entry to the grid manually and are used to be able to start editing directly (in Vaadin 7 it was possible like that). The necessary double click before that is annoying to the user and definitely not intuitive.

Wnt:
Here's one workaround for opening the editor programmatically from the server-side:

[![See the full sample on vaadinfiddle.com](https://vaadinfiddle.com/editor/preview-image/d3f33ce5e924ee24d41cf748e7bc7d3ef9a3bbdbe1694c490b6d9c7146e835a8_src_main_java_org_vaadin_vaadinfiddle_MyUI.java.png)](https://vaadinfiddle.com/editor/container/d3f33ce5e924ee24d41cf748e7bc7d3ef9a3bbdbe1694c490b6d9c7146e835a8/src/main/java/org/vaadin/vaadinfiddle/MyUI.java)

[See the full sample on vaadinfiddle.com](https://vaadinfiddle.com/editor/container/d3f33ce5e924ee24d41cf748e7bc7d3ef9a3bbdbe1694c490b6d9c7146e835a8/src/main/java/org/vaadin/vaadinfiddle/MyUI.java)

sirbris:
thank you very much for providing a workaround! works, but obviously not the best approach and it would require quiet some adjustments on our side. will there be a functionality like that provided by the framework at some time?

sirbris:
[...]

PR #10040, opened 2017-09-20
Grid edit row by index - server side

### LINKED ISSUE #8477: Programmatically open the editor row in Grid for an item based on row index
Vaadin version: 8.0-SNAPSHOT

Apparently, there is no way to programmatically open the edit menu for Vaadin 8 Grid component.
It was possible in Vaadin 7.
Can you add the method to Vaadin 8 Grid please?

pwilkin:
Opening grid editor from server side. Fixes #8477, will probably delegate #8820 to user responsibility (fetching index of item might be slow).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10040)

### DIFF
--- a/all/src/main/templates/release-notes.html
+++ b/all/src/main/templates/release-notes.html
@@ -113,7 +113,7 @@
         <li>Row height in <tt>Grid</tt> is replaced with separate header, body and footer row heights</li>
         <li><tt>GridState</tt> variable <tt>rowHeight</tt> has been replaced by three variables.</li>
         <li><tt>Button</tt> has a new constructor that may cause constructor calls with null as first parameter to be ambiguous.</li>
-        <li><tt>DataCommunicator</tt> method <tt>getDataProviderSize</tt> is now <tt>public</tt>, not <tt>protected</tt>.</li>
+        <li><tt>DataCommunicator</tt> methods <tt>getDataProviderSize</tt> and <tt>fetchItemsWithRange</tt> are now <tt>public</tt>, not <tt>protected</tt>.</li>
         <li><tt>Binder</tt> method <tt>getBindings</tt> now returns a Collection, not a Set.</li>
         <li><tt>BindingBuilder</tt> now works like a proper builder. Adding a converter will not mark Binding as <tt>bound</tt> allowing chaining to the same object.</li>
         <li><tt>ErrorLevel</tt> is removed from <tt>ErrorMessage</tt> and now <tt>com.vaadin.shared.ui.ErrorLevel</tt> should be used.</li>
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -70,7 +70,6 @@ public class EditorConnector extends AbstractExtensionConnector {
                     Scheduler.get().scheduleDeferred(new ScheduledCommand() {
                         @Override
                         public void execute() {
-                            serverInitiated = fromServer;
                             boolean canEdit = true;
                             if (fromServer) {
                                 currentEditedRow = rowIndex;
@@ -108,10 +107,7 @@ public class EditorConnector extends AbstractExtensionConnector {
 
                 @Override
                 public void confirmBind(final boolean bindSucceeded) {
-                    if (currentRequest != null) {   // this might be a ping-pong after a delayed
-                                                    // client-side callback to a server-side edit request
-                        endRequest(bindSucceeded);
-                    }
+                    endRequest(bindSucceeded);
                 }
 
                 @Override
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 import java.util.List;
 
 import com.google.gwt.core.client.Scheduler;
-import com.google.gwt.core.client.Scheduler.ScheduledCommand;
 import com.google.gwt.user.client.ui.Widget;
 import com.vaadin.client.ComponentConnector;
 import com.vaadin.client.ConnectorMap;
@@ -67,34 +66,26 @@ public class EditorConnector extends AbstractExtensionConnector {
                 @Override
                 public void bind(final int rowIndex, boolean fromServer) {
                     // call this deferred to avoid issues with editing on init
-                    Scheduler.get().scheduleDeferred(new ScheduledCommand() {
-                        @Override
-                        public void execute() {
-                            boolean canEdit = true;
-                            if (fromServer) {
-                                currentEditedRow = rowIndex;
-                                if (getParent().getDataSource().getRow(rowIndex) != null) { // already have data
-                                    canEdit = true;
-                                } else {
-                                    // will need to wait for available data, register listener if necessary on first try
-                                    if (dataAvailableListener == null) {
-                                        dataAvailableListener = (event) -> {
-                                            Range range = event.getAvailableRows();
-                                            if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
-                                                getParent().getWidget().editRow(currentEditedRow);
-                                                waitingForAvailableData = false;
-                                            }
-                                        };
-                                        getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
+                    Scheduler.get().scheduleDeferred(() -> {
+                        if (fromServer) {
+                            currentEditedRow = rowIndex;
+                            // might need to wait for available data, register listener if necessary on first try
+                            // if data is available, ensureAvailability will immediately trigger the handler anyway,
+                            // so no need for alternative "immediately available" logic
+                            if (dataAvailableListener == null) {
+                                dataAvailableListener = (event) -> {
+                                    Range range = event.getAvailableRows();
+                                    if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
+                                        getParent().getWidget().editRow(currentEditedRow);
+                                        waitingForAvailableData = false;
                                     }
-                                    getParent().getDataSource().ensureAvailability(rowIndex, 1);
-                                    waitingForAvailableData = true;
-                                    canEdit = false;
-                                }
+                                };
+                                getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
                             }
-                            if (canEdit) {
-                                getParent().getWidget().editRow(rowIndex);
-                            }
+                            waitingForAvailableData = true;
+                            getParent().getDataSource().ensureAvailability(rowIndex, 1);
+                        } else {
+                            getParent().getWidget().editRow(rowIndex);
                         }
                     });
                 }
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -49,6 +49,9 @@ import elemental.json.JsonObject;
 @Connect(EditorImpl.class)
 public class EditorConnector extends AbstractExtensionConnector {
 
+    private Integer currentEditedRow = null;
+    private boolean waitingForAvailableData = false;
+
     /**
      * EditorHandler for communicating with the server-side implementation.
      */
@@ -56,9 +59,6 @@ public class EditorConnector extends AbstractExtensionConnector {
         private EditorServerRpc rpc = getRpcProxy(EditorServerRpc.class);
         private EditorRequest<JsonObject> currentRequest = null;
         private boolean serverInitiated = false;
-        private DataAvailableHandler dataAvailableListener = null;
-        private Integer currentEditedRow = null;
-        private boolean waitingForAvailableData = false;
 
         public CustomEditorHandler() {
             registerRpc(EditorClientRpc.class, new EditorClientRpc() {
@@ -68,19 +68,9 @@ public class EditorConnector extends AbstractExtensionConnector {
                     // call this deferred to avoid issues with editing on init
                     Scheduler.get().scheduleDeferred(() -> {
                         currentEditedRow = rowIndex;
-                        // might need to wait for available data, register listener if necessary on first try
+                        // might need to wait for available data,
                         // if data is available, ensureAvailability will immediately trigger the handler anyway,
                         // so no need for alternative "immediately available" logic
-                        if (dataAvailableListener == null) {
-                            dataAvailableListener = (event) -> {
-                                Range range = event.getAvailableRows();
-                                if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
-                                    getParent().getWidget().editRow(currentEditedRow);
-                                    waitingForAvailableData = false;
-                                }
-                            };
-                            getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
-                        }
                         waitingForAvailableData = true;
                         getParent().getDataSource().ensureAvailability(rowIndex, 1);
                     });
@@ -231,6 +221,13 @@ public class EditorConnector extends AbstractExtensionConnector {
     protected void extend(ServerConnector target) {
         Grid<JsonObject> grid = getParent().getWidget();
         grid.getEditor().setHandler(new CustomEditorHandler());
+        grid.addDataAvailableHandler((event) -> {
+            Range range = event.getAvailableRows();
+            if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
+                getParent().getWidget().editRow(currentEditedRow);
+                waitingForAvailableData = false;
+            }
+        });
     }
 
     @Override
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -26,7 +26,6 @@ import com.vaadin.client.ConnectorMap;
 import com.vaadin.client.ServerConnector;
 import com.vaadin.client.annotations.OnStateChange;
 import com.vaadin.client.extensions.AbstractExtensionConnector;
-import com.vaadin.client.widget.grid.DataAvailableHandler;
 import com.vaadin.client.widget.grid.EditorHandler;
 import com.vaadin.client.widgets.Grid;
 import com.vaadin.client.widgets.Grid.Column;
@@ -112,9 +111,7 @@ public class EditorConnector extends AbstractExtensionConnector {
 
         @Override
         public void bind(EditorRequest<JsonObject> request) {
-            if (!handleServerInitiated(request)) {
-                startRequest(request);
-            }
+            startRequest(request);
             rpc.bind(getRowKey(request.getRow()));
         }
 
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -64,29 +64,25 @@ public class EditorConnector extends AbstractExtensionConnector {
             registerRpc(EditorClientRpc.class, new EditorClientRpc() {
 
                 @Override
-                public void bind(final int rowIndex, boolean fromServer) {
+                public void bind(final int rowIndex) {
                     // call this deferred to avoid issues with editing on init
                     Scheduler.get().scheduleDeferred(() -> {
-                        if (fromServer) {
-                            currentEditedRow = rowIndex;
-                            // might need to wait for available data, register listener if necessary on first try
-                            // if data is available, ensureAvailability will immediately trigger the handler anyway,
-                            // so no need for alternative "immediately available" logic
-                            if (dataAvailableListener == null) {
-                                dataAvailableListener = (event) -> {
-                                    Range range = event.getAvailableRows();
-                                    if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
-                                        getParent().getWidget().editRow(currentEditedRow);
-                                        waitingForAvailableData = false;
-                                    }
-                                };
-                                getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
-                            }
-                            waitingForAvailableData = true;
-                            getParent().getDataSource().ensureAvailability(rowIndex, 1);
-                        } else {
-                            getParent().getWidget().editRow(rowIndex);
+                        currentEditedRow = rowIndex;
+                        // might need to wait for available data, register listener if necessary on first try
+                        // if data is available, ensureAvailability will immediately trigger the handler anyway,
+                        // so no need for alternative "immediately available" logic
+                        if (dataAvailableListener == null) {
+                            dataAvailableListener = (event) -> {
+                                Range range = event.getAvailableRows();
+                                if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
+                                    getParent().getWidget().editRow(currentEditedRow);
+                                    waitingForAvailableData = false;
+                                }
+                            };
+                            getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
                         }
+                        waitingForAvailableData = true;
+                        getParent().getDataSource().ensureAvailability(rowIndex, 1);
                     });
                 }
 
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -60,11 +60,12 @@ public class EditorConnector extends AbstractExtensionConnector {
             registerRpc(EditorClientRpc.class, new EditorClientRpc() {
 
                 @Override
-                public void bind(final int rowIndex) {
+                public void bind(final int rowIndex, boolean fromServer) {
                     // call this deferred to avoid issues with editing on init
                     Scheduler.get().scheduleDeferred(new ScheduledCommand() {
                         @Override
                         public void execute() {
+                            serverInitiated = fromServer;
                             getParent().getWidget().editRow(rowIndex);
                         }
                     });
@@ -106,8 +107,20 @@ public class EditorConnector extends AbstractExtensionConnector {
 
         @Override
         public void bind(EditorRequest<JsonObject> request) {
-            startRequest(request);
-            rpc.bind(getRowKey(request.getRow()));
+            if (!handleServerInitiated(request)) {
+                startRequest(request);
+                rpc.bind(getRowKey(request.getRow()));
+            } else {
+                if (getParent().rowHasDetails(request.getRowIndex())) {
+                    rpc.bind(getRowKey(request.getRow()));
+                } else {
+                    getParent().addDetailsRefreshCallback(() -> {
+                        if (getParent().rowHasDetails(request.getRowIndex())) {
+                            rpc.bind(getRowKey(request.getRow()));
+                        }
+                    });
+                }
+            }
         }
 
         @Override
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -19,6 +19,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import com.google.gwt.core.client.Scheduler;
+import com.google.gwt.core.client.Scheduler.ScheduledCommand;
 import com.google.gwt.user.client.ui.Widget;
 import com.vaadin.client.ComponentConnector;
 import com.vaadin.client.ConnectorMap;
@@ -56,6 +58,18 @@ public class EditorConnector extends AbstractExtensionConnector {
 
         public CustomEditorHandler() {
             registerRpc(EditorClientRpc.class, new EditorClientRpc() {
+
+                @Override
+                public void bind(final int rowIndex) {
+                    // call this deferred to avoid issues with editing on init
+                    Scheduler.get().scheduleDeferred(new ScheduledCommand() {
+                        @Override
+                        public void execute() {
+                            getParent().getWidget().editRow(rowIndex);
+                        }
+                    });
+                }
+
                 @Override
                 public void cancel() {
                     serverInitiated = true;
--- a/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java
@@ -27,9 +27,11 @@ import com.vaadin.client.ConnectorMap;
 import com.vaadin.client.ServerConnector;
 import com.vaadin.client.annotations.OnStateChange;
 import com.vaadin.client.extensions.AbstractExtensionConnector;
+import com.vaadin.client.widget.grid.DataAvailableHandler;
 import com.vaadin.client.widget.grid.EditorHandler;
 import com.vaadin.client.widgets.Grid;
 import com.vaadin.client.widgets.Grid.Column;
+import com.vaadin.shared.Range;
 import com.vaadin.shared.data.DataCommunicatorConstants;
 import com.vaadin.shared.ui.Connect;
 import com.vaadin.shared.ui.grid.editor.EditorClientRpc;
@@ -55,6 +57,9 @@ public class EditorConnector extends AbstractExtensionConnector {
         private EditorServerRpc rpc = getRpcProxy(EditorServerRpc.class);
         private EditorRequest<JsonObject> currentRequest = null;
         private boolean serverInitiated = false;
+        private DataAvailableHandler dataAvailableListener = null;
+        private Integer currentEditedRow = null;
+        private boolean waitingForAvailableData = false;
 
         public CustomEditorHandler() {
             registerRpc(EditorClientRpc.class, new EditorClientRpc() {
@@ -66,7 +71,31 @@ public class EditorConnector extends AbstractExtensionConnector {
                         @Override
                         public void execute() {
                             serverInitiated = fromServer;
-                            getParent().getWidget().editRow(rowIndex);
+                            boolean canEdit = true;
+                            if (fromServer) {
+                                currentEditedRow = rowIndex;
+                                if (getParent().getDataSource().getRow(rowIndex) != null) { // already have data
+                                    canEdit = true;
+                                } else {
+                                    // will need to wait for available data, register listener if necessary on first try
+                                    if (dataAvailableListener == null) {
+                                        dataAvailableListener = (event) -> {
+                                            Range range = event.getAvailableRows();
+                                            if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
+                                                getParent().getWidget().editRow(currentEditedRow);
+                                                waitingForAvailableData = false;
+                                            }
+                                        };
+                                        getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
+                                    }
+                                    getParent().getDataSource().ensureAvailability(rowIndex, 1);
+                                    waitingForAvailableData = true;
+                                    canEdit = false;
+                                }
+                            }
+                            if (canEdit) {
+                                getParent().getWidget().editRow(rowIndex);
+                            }
                         }
                     });
                 }
@@ -79,7 +108,10 @@ public class EditorConnector extends AbstractExtensionConnector {
 
                 @Override
                 public void confirmBind(final boolean bindSucceeded) {
-                    endRequest(bindSucceeded);
+                    if (currentRequest != null) {   // this might be a ping-pong after a delayed
+                                                    // client-side callback to a server-side edit request
+                        endRequest(bindSucceeded);
+                    }
                 }
 
                 @Override
@@ -109,18 +141,8 @@ public class EditorConnector extends AbstractExtensionConnector {
         public void bind(EditorRequest<JsonObject> request) {
             if (!handleServerInitiated(request)) {
                 startRequest(request);
-                rpc.bind(getRowKey(request.getRow()));
-            } else {
-                if (getParent().rowHasDetails(request.getRowIndex())) {
-                    rpc.bind(getRowKey(request.getRow()));
-                } else {
-                    getParent().addDetailsRefreshCallback(() -> {
-                        if (getParent().rowHasDetails(request.getRowIndex())) {
-                            rpc.bind(getRowKey(request.getRow()));
-                        }
-                    });
-                }
             }
+            rpc.bind(getRowKey(request.getRow()));
         }
 
         @Override
@@ -137,6 +159,7 @@ public class EditorConnector extends AbstractExtensionConnector {
                 // a confirmation from the server
                 rpc.cancel(afterBeingSaved);
             }
+            currentEditedRow = null;
         }
 
         @Override
--- a/client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java
+++ b/client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java
@@ -184,7 +184,7 @@ public class GridConnector extends AbstractListingConnector
      * @param refreshCallback
      *            Details refreshed callback
      */
-    public void addDetailsRefreshCallback(Runnable refreshCallback) {
+    private void addDetailsRefreshCallback(Runnable refreshCallback) {
         refreshDetailsCallbacks.add(refreshCallback);
     }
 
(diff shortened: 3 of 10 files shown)


tsuoanttila:
Reviewed 5 of 5 files at r1.
Review status: 2 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 373 at r1](https://reviewable.io:443/reviews/vaadin/framework/10040#-KuZCYrSjXlQ5nzzdVUA:-KuZCYrSjXlQ5nzzdVUB:b-vap2bb) ([raw file](https://github.com/vaadin/framework/blob/893fa911ef5695d8e63cd2144c80413b54fbc70e/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L373)):*
> ```Java
>      */
>     @SuppressWarnings({ "rawtypes", "unchecked" })
>     public List<T> fetchItemsWithRange(int offset, int limit) {
> ```

Needs release notes mention about changed visibility (same issue as with the `getDataProviderSize()`)

---

*[server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java, line 292 at r1](https://reviewable.io:443/reviews/vaadin/framework/10040#-KuZC7ct2wciJm5iU8QU:-KuZC7ct2wciJm5iU8QV:bmzr34d) ([raw file](https://github.com/vaadin/framework/blob/893fa911ef5695d8e63cd2144c80413b54fbc70e/server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java#L292)):*
> ```Java
>         T beanToEdit = getParent().getDataCommunicator().
>             fetchItemsWithRange(rowNumber, 1).
>             stream().findFirst().orElse(null);
> ```

I'd recommend using `.orElseThrow()`. Edit with the already edited item, should be just no-op, no need to throw in my opinion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 1 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 373 at r1](https://reviewable.io:443/reviews/vaadin/framework/10040#-KuZCYrSjXlQ5nzzdVUA:-KuZX6Fn8yjdhx3j9PJm:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/893fa911ef5695d8e63cd2144c80413b54fbc70e/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L373)):*

Needs release notes mention about changed visibility (same issue as with the `getDataProviderSize()`)
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java, line 292 at r1](https://reviewable.io:443/reviews/vaadin/framework/10040#-KuZC7ct2wciJm5iU8QU:-KuZX52YWW0d0aqk-r5v:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/893fa911ef5695d8e63cd2144c80413b54fbc70e/server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java#L292)):*

I'd recommend using `.orElseThrow()`. Edit with the already edited item, should be just no-op, no need to throw in my opinion.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Rebased on current master to avoid conflicts in release notes.

pwilkin:
@mvysny could you please comment?

tsuoanttila:
Reviewed 1 of 1 files at r2, 5 of 6 files at r3, 2 of 2 files at r4.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 117 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-KurnkxN_47q7cljixut:-KurnkxOm2N2QemXn7bp:b-ufenr2) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L117)):*
> ```Java
>                     rpc.bind(getRowKey(request.getRow()));
>                 } else {
>                     getParent().addDetailsRefreshCallback(() -> {
> ```

There's a new callback for every row being edited. Can this be made to have only one refresh callback always? (store the relevant information somewhere, use through a field)

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 119 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kurp-_8-aWu0FoyExpJ:-Kurp-_8-aWu0FoyExpK:b6dlqbq) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L119)):*
> ```Java
>                     getParent().addDetailsRefreshCallback(() -> {
>                         if (getParent().rowHasDetails(request.getRowIndex())) {
>                             rpc.bind(getRowKey(request.getRow()));
> ```

why are we calling the server to bind the same item here?

---

*[server/src/main/java/com/vaadin/ui/components/grid/Editor.java, line 111 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-KurpxRnNvFDXwXZK1F5:-KurpxRnNvFDXwXZK1F6:b-7jfxkb) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/server/src/main/java/com/vaadin/ui/components/grid/Editor.java#L111)):*
> ```Java
>      * @param rowNumber
>      *          the row to edit
>      */
> ```

since

---

*[server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java, line 296 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-KurrNyGpsPcW9zi9xkL:-KurrNyGpsPcW9zi9xkM:b-wjhu3) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java#L296)):*
> ```Java
>         if (!beanToEdit.equals(edited)) {
>             if (isBuffered() && edited != null) {
>                 throw new IllegalStateException("Editing item " + beanToEdit
> ```

Still not sure if throwing instead of doing nothing when trying to edit something that is already being edited is the way to go.

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java, line 37 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuroe4fwHFD4t5J4CS1:-Kuroe4fwHFD4t5J4CS2:b-7jfxkb) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java#L37)):*
> ```Java
>      * @param fromServer
>      *            whether the request was server-originated
>      *
> ```

since

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 117 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-KurnkxN_47q7cljixut:-KusYRZ4r_eMx64IvTtG:bnyumk8) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L117)):*

There's a new callback for every row being edited. Can this be made to have only one refresh callback always? (store the relevant information somewhere, use through a field)
</blockquote></details>

This doesn't matter since detailsRefreshed (which gets called whenever a row is refreshed) clears all callbacks anyway, so doing any sort of caching would be simply having extraneous code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 119 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kurp-_8-aWu0FoyExpJ:-KuskTs8WJkE1f_JlKct:beqtb4u) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L119)):*

why are we calling the server to bind the same item here?
</blockquote></details>

Because we need to wait on the scroll event, so we can't edit instantly on the server. On the other hand, if the components aren't set up server-side first, the detailsRefreshed call does not think there are any components to show and does not trigger the detailsRefreshedCallbacks before removing them. This is the only interaction that a) does not throw errors because the row isn't retrieved yet after scrolling (since details retrieval upon scrolling is called server-side) and b) actually shows the editor components

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java, line 296 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-KurrNyGpsPcW9zi9xkL:-KuslsBfwGzl-hGgWZHi:b-i4dhla) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java#L296)):*

Still not sure if throwing instead of doing nothing when trying to edit something that is already being edited is the way to go.
</blockquote></details>

I copied this behavior from Grid v7. I think sticking to known behavior is the way to go, plus throwing an error here actually tells the developer they are doing something wrong - a NOOP might be very confusing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 6 of 9 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/ui/components/grid/Editor.java, line 111 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-KurpxRnNvFDXwXZK1F5:-KusmicoG6l67PBzoSc_:b-8b11j9) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/server/src/main/java/com/vaadin/ui/components/grid/Editor.java#L111)):*

since
</blockquote></details>

Added

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java, line 37 at r4](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuroe4fwHFD4t5J4CS1:-KusmjXakLP5R5umFcgQ:b-8b11j9) ([raw file](https://github.com/vaadin/framework/blob/c10ad90493a3fda07a131a894e722f3e910895ac/shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java#L37)):*

since
</blockquote></details>

Added

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Rebased on current master again.

pwilkin:
Rebased and merged release-notes.html

tsuoanttila:
Reviewed 4 of 6 files at r6, 1 of 2 files at r7, 3 of 3 files at r8.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 114 at r8](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuy1yzYdJCb54OTQkMw:-Kuy1yzZapVditbJNad-:b-icph3z) ([raw file](https://github.com/vaadin/framework/blob/29dd624f76ec980e2c4b04d529f4822c8091bb80/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L114)):*
> ```Java
>                 rpc.bind(getRowKey(request.getRow()));
>             } else {
>                 if (getParent().rowHasDetails(request.getRowIndex())) {
> ```

I still don't get why binding the editor is looking for `JSONKEY_DETAILS_VISIBLE` and it's value. I'm pretty sure there is some better way to handle this specific interaction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 114 at r8](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuy1yzYdJCb54OTQkMw:-Kuy7yn3pdCn-xG21PvE:b6zoqd7) ([raw file](https://github.com/vaadin/framework/blob/29dd624f76ec980e2c4b04d529f4822c8091bb80/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L114)):*

I still don't get why binding the editor is looking for `JSONKEY_DETAILS_VISIBLE` and it's value. I'm pretty sure there is some better way to handle this specific interaction.
</blockquote></details>

You're right, I'll try to rewrite this to utilize dataChangeHandler instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

tsuoanttila:
*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 114 at r8](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuy1yzYdJCb54OTQkMw:-Kuy9kVXT5N7_AN6Z1n8:b-p3j1ku) ([raw file](https://github.com/vaadin/framework/blob/29dd624f76ec980e2c4b04d529f4822c8091bb80/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L114)):*
<details><summary><i>Previously, pwilkin (Piotr Wilkin) wrote…</i></summary><blockquote>

You're right, I'll try to rewrite this to utilize dataChangeHandler instead.
</blockquote></details>

I'd recommend using `getParent().getWidget().addDataAvailableHandler()` instead.

Only provides the necessary information without bothering with extra information.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

tsuoanttila:
*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 114 at r8](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuy1yzYdJCb54OTQkMw:-Kuy9uaumZFJHD3phCB1:b-s4ohs6) ([raw file](https://github.com/vaadin/framework/blob/29dd624f76ec980e2c4b04d529f4822c8091bb80/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L114)):*

I'd recommend using `getParent().getWidget().addDataAvailableHandler()` instead.

Only provides the necessary information without bothering with extra information.
</blockquote></details>

(had to check that it still exists and does what I remembered it to do)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 114 at r8](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kuy1yzYdJCb54OTQkMw:-KuyAawiN0Rqi-P3UL-f:bkfwjjd) ([raw file](https://github.com/vaadin/framework/blob/29dd624f76ec980e2c4b04d529f4822c8091bb80/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L114)):*

(had to check that it still exists and does what I remembered it to do)
</blockquote></details>

Yeah, thanks, that makes it easier as I don't have to expose SimpleDataChangeHandler.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Okay, had to rewrite the logic (pull it earlier in the client call tree) because otherwise non-editable column details were not being updated properly. Tested that this works fine both in buffered and unbuffered mode.

tsuoanttila:
*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kv17tydgVd-vgvlX-PX:-Kv17tydgVd-vgvlX-PY:beyyevl) (no related file):*
Looking at the size of this change, I think it might be clearer and less `if then else` blocks if we implement this the same way it was done in Vaadin 7. Basically what server edit meant there was:
1. Check if index is in range `[0, container.size())`
2. Tell client to edit row based on index
3. Client waits for data, tells server to start editing.
4. Server behaves exactly the same as with normal client-side initiated editing.

There will be a small duration where the edited row is visible, but the editor is not open. This can be prevented by adding an overloaded API to the GridEditorServerRpc so it accepts both index or item key. And in this server initiated case we'd use the index to find out the item on the correct row while also processing the data to display altogether.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 7 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kv17tydgVd-vgvlX-PX:-Kv74dpT_YIMwO9Whtpa:bgiy40g) (no related file):*

Looking at the size of this change, I think it might be clearer and less `if then else` blocks if we implement this the same way it was done in Vaadin 7. Basically what server edit meant there was:
1. Check if index is in range `[0, container.size())`
2. Tell client to edit row based on index
3. Client waits for data, tells server to start editing.
4. Server behaves exactly the same as with normal client-side initiated editing.

There will be a small duration where the edited row is visible, but the editor is not open. This can be prevented by adding an overloaded API to the GridEditorServerRpc so it accepts both index or item key. And in this server initiated case we'd use the index to find out the item on the correct row while also processing the data to display altogether.
</blockquote></details>

Okay, I rewrote the logic:
* removed the ping-pong by eliminating the "server initiated flag", making the deferred client request similar to the normal client request
* removed the prerendering of the editor on server-side - now it's handled just as you suggested, as a normal response to a client-side callback
* rewrote the connector bind logic to omit the if-then-else, taking advantage of the fact that ensureAvailability calls the dataAvailableListener anyway
* added a comment explaining the above to potential contributors :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

tsuoanttila:
Review status: 7 of 9 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 70 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBXOZ4981RDWoI7usn:-KvBXOZ4981RDWoI7uso:bz8n9to) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L70)):*
> ```Java
>                     // call this deferred to avoid issues with editing on init
>                     Scheduler.get().scheduleDeferred(() -> {
>                         if (fromServer) {
> ```

see comment about the variable `fromServer` in the RPC Interface.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 79 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBYsCfXHiMMMwc4-V8:-KvBYsCfXHiMMMwc4-V9:b-mgs61m) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L79)):*
> ```Java
>                                     Range range = event.getAvailableRows();
>                                     if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
>                                         getParent().getWidget().editRow(currentEditedRow);
> ```

`Editor::editRow` can be used when data is not yet available. This will scroll to the correct row and calls the `EditorHandler::bind` method implemented in this class.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 83 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBXXEYB5L6s-HNKeg3:-KvBXXEYB5L6s-HNKeg4:bmevjkc) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L83)):*
> ```Java
>                                     }
>                                 };
>                                 getParent().getWidget().addDataAvailableHandler(dataAvailableListener);
> ```

is the data available handler ever removed? The Pattern I've used the `DataAvailableHandler`is basically: 
```
// field
HandlerRegistration registration;

// in my code block
if (registration != null)
    registration.removeHandler();

registration = grid.addDataAvailableHandler(event -> {
    if (matchesMyCondition(event)) {
        doTheStuff();
        registration.removeHandler();
        registration = null;
    }
});
```

Alternatively you can register one implementation that will always listen to the available row range, and act on it when needed. In every case you should make sure any `HandlerRegistration`

---

*[server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java, line 300 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBcQ7uMc4pxuwMJ-W4:-KvBcQ7uMc4pxuwMJ-W5:bjr7erz) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java#L300)):*
> ```Java
>                     + edited);
>             } else {
>                 rpc.bind(rowNumber, true);
> ```

unnecessary parameter, see comment in RPC Interface.

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java, line 41 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBXElXPGGLUtLKntTw:-KvBXElXPGGLUtLKntTx:bjfp9sq) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java#L41)):*
> ```Java
>      *
>      */
>     void bind(int rowIndex, boolean fromServer);
> ```

This method is only called from one place, and the `fromServer` is always `true`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 83 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBXXEYB5L6s-HNKeg3:-KvC0LCRhL1ch7NwMj5B:bbi2nus) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L83)):*

is the data available handler ever removed? The Pattern I've used the `DataAvailableHandler`is basically: 
```
// field
HandlerRegistration registration;

// in my code block
if (registration != null)
    registration.removeHandler();

registration = grid.addDataAvailableHandler(event -> {
    if (matchesMyCondition(event)) {
        doTheStuff();
        registration.removeHandler();
        registration = null;
    }
});
```

Alternatively you can register one implementation that will always listen to the available row range, and act on it when needed. In every case you should make sure any `HandlerRegistration`
</blockquote></details>

That pattern doesn't work (try it!). Variables used in lambdas have to be effectively final, meaning if you assign a value to it later, you cannot use it within the lambda. In other words - you cannot use a handler to deregister itself. 

I'm using a single listener for all calls, hence the dataAvailableListener field. Since I never remove the listener (it'll be garbage collected anyways when the RPC is thrown out, there is never a valid reason to unregister it), there is no reason to keep the registration.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 70 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBXOZ4981RDWoI7usn:-KvC10fZgFPb5F1sniby:bgk7ezg) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L70)):*

see comment about the variable `fromServer` in the RPC Interface.
</blockquote></details>

Okay, the method was there in the first place so I thought it's somehow used, but it must have been a relic from the Vaadin 7 RPC.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 79 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBYsCfXHiMMMwc4-V8:-KvC1JCmX4jvQ1SqZWeV:b83dexl) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L79)):*

`Editor::editRow` can be used when data is not yet available. This will scroll to the correct row and calls the `EditorHandler::bind` method implemented in this class.
</blockquote></details>

It can, but it will not work correctly (as a matter of fact, that's the first thing I've tried). Since the data is not yet retrieved, trying to open the editor will throw an error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java, line 300 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBcQ7uMc4pxuwMJ-W4:-KvC1gLqxbojtOo6xm0I:bss4a2b) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/server/src/main/java/com/vaadin/ui/components/grid/EditorImpl.java#L300)):*

unnecessary parameter, see comment in RPC Interface.
</blockquote></details>

Removed.

---

*[shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java, line 41 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBXElXPGGLUtLKntTw:-KvC1hhL7t1IbA9_8tsb:bss4a2b) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/shared/src/main/java/com/vaadin/shared/ui/grid/editor/EditorClientRpc.java#L41)):*

This method is only called from one place, and the `fromServer` is always `true`.
</blockquote></details>

Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

tsuoanttila:
Reviewed 2 of 5 files at r13, 1 of 2 files at r14, 1 of 2 files at r15, 3 of 3 files at r19.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 79 at r11](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvBYsCfXHiMMMwc4-V8:-KvWKFSSV0mMZW0SZXk3:b-opmhi) ([raw file](https://github.com/vaadin/framework/blob/13483f63b4063e72b12320c76466033e29f6d8b0/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L79)):*
<details><summary><i>Previously, pwilkin (Piotr Wilkin) wrote…</i></summary><blockquote>

It can, but it will not work correctly (as a matter of fact, that's the first thing I've tried). Since the data is not yet retrieved, trying to open the editor will throw an error.
</blockquote></details>

If you do the `Editor::editRow` first, you need to move the range check in to the `bind` method. But this is ok for now.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 75 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWL1oUAO-S36uUiSnh:-KvWL1oUAO-S36uUiSni:b-qw6ucp) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L75)):*
> ```Java
>                         // so no need for alternative "immediately available" logic
>                         if (dataAvailableListener == null) {
>                             dataAvailableListener = (event) -> {
> ```

Since this is only registered once, might as well do in in the init method of the `extend` method and have it always initialised.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 79 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWM41U6Y3czcbhbpAG:-KvWM41Vdvc1Q_-KlEYX:b-7d8qt0) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L79)):*
> ```Java
>                                 if (waitingForAvailableData && currentEditedRow != null && range.contains(currentEditedRow)) {
>                                     getParent().getWidget().editRow(currentEditedRow);
>                                     waitingForAvailableData = false;
> ```

should the `currentEditedRow` be set to `null` in here as well? 

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 231 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWLgCkMvEk-6O8lPox:-KvWLgCkMvEk-6O8lPoy:b-ncmkqw) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L231)):*
> ```Java
> 
>     @Override
>     protected void extend(ServerConnector target) {
> ```

move data available handler adding here.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java, line 187 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWKe-nCxaNLPH26zBQ:-KvWKe-nCxaNLPH26zBR:bs5d47q) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java#L187)):*
> ```Java
>      *            Details refreshed callback
>      */
>     public void addDetailsRefreshCallback(Runnable refreshCallback) {
> ```

This API change is no longer needed

---

*[client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java, line 197 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWKgKIrF_4tneglnXf:-KvWKgKIrF_4tneglnXg:b-2guby3) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java#L197)):*
> ```Java
>      * @return
>      */
>     public boolean rowHasDetails(int rowIndex) {
> ```

same here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 7 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 75 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWL1oUAO-S36uUiSnh:-KvXNVWRRofuzgxL3_Om:b-837i3v) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L75)):*

Since this is only registered once, might as well do in in the init method of the `extend` method and have it always initialised.
</blockquote></details>

Okay.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 79 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWM41U6Y3czcbhbpAG:-KvXO8rHwzPH2_hE1LFs:b-z17avv) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L79)):*

should the `currentEditedRow` be set to `null` in here as well? 
</blockquote></details>

Not really - it doesn't accomplish anything and setting it to null seems to suggests that we are somehow cancelling the editor. I'd prefer not to change it.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 231 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWLgCkMvEk-6O8lPox:-KvXQFLdujRTR62vyKYF:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L231)):*

move data available handler adding here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Review status: 7 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java, line 187 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWKe-nCxaNLPH26zBQ:-KvXR1aLCIES3ovwxCk0:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java#L187)):*

This API change is no longer needed
</blockquote></details>

Done.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java, line 197 at r19](https://reviewable.io:443/reviews/vaadin/framework/10040#-KvWKgKIrF_4tneglnXf:-KvXR2dAS-wbKxQ7uvA1:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/7e38d97c268a4915bf3c0028692188130652a5b1/client/src/main/java/com/vaadin/client/connectors/grid/GridConnector.java#L197)):*

same here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
This should be all hopefully (would be nice if this makes it in 8.2).

tsuoanttila:
The last thing missing here is a bunch of tests for the new functionality. I'll try to get those done and pushed into your branch on monday.

pwilkin:
Would be great, I don't really have much experience writiing the client-side tests, so could use a hand (not sure how you check client-side data from a server-side test or vice versa).

tsuoanttila:
Reviewed 1 of 8 files at r12, 1 of 1 files at r20, 1 of 1 files at r21.
Review status: 7 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[uitest/src/main/java/com/vaadin/tests/components/grid/basics/GridBasics.java, line 763 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw0Lcz1_803gwQz-yOn:-Kw0Lcz1_803gwQz-yOo:b-z5xvtv) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/uitest/src/main/java/com/vaadin/tests/components/grid/basics/GridBasics.java#L763)):*
> ```Java
>         editorMenu.addItem("Cancel edit", i -> grid.getEditor().cancel());
> 
>         Stream.of(0, 5, 100).forEach(i -> editorMenu.addItem("Edit row " + i,
> ```

I used an already existing test ui called `GridBasics` that has a bunch of editing options already. Added new menu entries for opening various rows for editing.

---

*[uitest/src/test/java/com/vaadin/tests/components/grid/basics/GridEditorTest.java, line 35 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw0M5p-7IYTY3EPe27O:-Kw0M5p-7IYTY3EPe27P:bey152j) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/uitest/src/test/java/com/vaadin/tests/components/grid/basics/GridEditorTest.java#L35)):*
> ```Java
> import com.vaadin.testbench.elements.GridElement.GridEditorElement;
> 
> public abstract class GridEditorTest extends GridBasicsTest {
> ```

This test UI is extended by buffered and unbuffered versions and any test changes here will be run against both variants of the editor.

---

*[uitest/src/test/java/com/vaadin/tests/components/grid/basics/GridEditorTest.java, line 221 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw0LqhPlM4PwVpCIV-W:-Kw0LqhPlM4PwVpCIV-X:b-d30yhc) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/uitest/src/test/java/com/vaadin/tests/components/grid/basics/GridEditorTest.java#L221)):*
> ```Java
> 
>     @Test
>     public void testEditorOpeningFromServer() {
> ```

This test opens the editor through the menu and checks the content of the editor.

---

*[uitest/src/test/java/com/vaadin/tests/components/grid/basics/GridEditorTest.java, line 233 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw0LwJIKpO3bvmZuRCN:-Kw0LwJIKpO3bvmZuRCO:bhkpox2) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/uitest/src/test/java/com/vaadin/tests/components/grid/basics/GridEditorTest.java#L233)):*
> ```Java
> 
>     @Test
>     public void testEditorOpenWithScrollFromServer() {
> ```

This does exactly the same, but with the last row in the editor. This is to verify that the scrolling and displaying the not editable content works also when scrolled.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:
Reviewed 2 of 2 files at r22.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

tsuoanttila:
@hesara want to take the last look at this before merge?

hesara:
Reviewed 1 of 1 files at r2, 1 of 8 files at r12, 1 of 5 files at r13, 1 of 2 files at r14, 1 of 2 files at r15, 2 of 3 files at r19, 1 of 1 files at r20, 1 of 1 files at r21, 2 of 2 files at r22.

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 115 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw3afOusldkjIOqbXU8:-Kw3afOusldkjIOqbXU9:byasl96) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L115)):*
> ```Java
>         @Override
>         public void bind(EditorRequest<JsonObject> request) {
>             if (!handleServerInitiated(request)) {
> ```

What sets serverInitiated for this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

pwilkin:

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 115 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw3afOusldkjIOqbXU8:-Kw4KsnrxjmaH9ad1Zu2:b-nh7fh7) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L115)):*

What sets serverInitiated for this?
</blockquote></details>

It's a relic from the previous version, probably should get edited out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

hesara:
:lgtm:

---

Reviewed 1 of 1 files at r23.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040#-:-Kw4TPQGsDQx4sAs2sVQ:bnfp4nl)*

pwilkin:

---

*[client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java, line 115 at r22](https://reviewable.io:443/reviews/vaadin/framework/10040#-Kw3afOusldkjIOqbXU8:-Kw4UFa705UbBXDOb8nD:b-p8gfkt) ([raw file](https://github.com/vaadin/framework/blob/cb1484c5c86149ea2ccf70acc0fe87abaa817c64/client/src/main/java/com/vaadin/client/connectors/grid/EditorConnector.java#L115)):*
<details><summary><i>Previously, pwilkin (Piotr Wilkin) wrote…</i></summary><blockquote>

It's a relic from the previous version, probably should get edited out.
</blockquote></details>

Okay, cleaned it up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10040)*

PR #10045, opened 2017-09-20
Use String.isEmpty()

asashour:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10045)

### DIFF
--- a/client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java
+++ b/client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java
@@ -143,7 +143,7 @@ public class ReportUsage {
         appendUserAgentProperty(extra, "os.name");
         appendUserAgentProperty(extra, "os.version");
 
-        if (extra.length() > 0) {
+        if (extra.length() != 0) {
             userAgent += " (" + extra.toString() + ")";
         }
 
@@ -154,7 +154,7 @@ public class ReportUsage {
             String propName) {
         String propValue = System.getProperty(propName);
         if (propValue != null) {
-            if (sb.length() > 0) {
+            if (sb.length() != 0) {
                 sb.append(';').append(' ');
             }
             sb.append(propName);
--- a/client/src/main/java/com/vaadin/client/JavaScriptConnectorHelper.java
+++ b/client/src/main/java/com/vaadin/client/JavaScriptConnectorHelper.java
@@ -331,7 +331,7 @@ public class JavaScriptConnectorHelper {
     }
 
     private ServerConnector getConnector(String connectorId) {
-        if (connectorId == null || connectorId.length() == 0) {
+        if (connectorId == null || connectorId.isEmpty()) {
             return connector;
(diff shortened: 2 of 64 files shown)


hesara:
Reviewed 64 of 64 files at r1.

---

*[client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java, line 146 at r1](https://reviewable.io:443/reviews/vaadin/framework/10045#-KuYE6SuYi4YESv-8JCq:-KuYE6SvSQkAh03SAhGe:b-pxlxby) ([raw file](https://github.com/vaadin/framework/blob/2dcd886b574c52ed283db984785e1e11041cdd82/client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java#L146)):*
> ```Java
>         appendUserAgentProperty(extra, "os.version");
> 
>         if (extra.length() != 0) {
> ```

(personally I don't see this as any clearer, but I'm not opposed to it either so I will accept these in review)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10045)*

asashour:

---

*[client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java, line 146 at r1](https://reviewable.io:443/reviews/vaadin/framework/10045#-KuYE6SuYi4YESv-8JCq:-KuYUgJB39Sb8g_kuNHY:b-95ele4) ([raw file](https://github.com/vaadin/framework/blob/2dcd886b574c52ed283db984785e1e11041cdd82/client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java#L146)):*

(personally I don't see this as any clearer, but I'm not opposed to it either so I will accept these in review)
</blockquote></details>

I am also not very inclined for it, however I *hope* that it could be a little faster (at [machine level](https://stackoverflow.com/questions/12135518/is-faster-than)) and for human reading.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10045)*

hesara:

---

*[client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java, line 146 at r1](https://reviewable.io:443/reviews/vaadin/framework/10045#-KuYE6SuYi4YESv-8JCq:-KuYaopoaUYtBE1NtOTm:b6wxmfn) ([raw file](https://github.com/vaadin/framework/blob/2dcd886b574c52ed283db984785e1e11041cdd82/client-compiler/src/main/java/com/vaadin/tools/ReportUsage.java#L146)):*

I am also not very inclined for it, however I *hope* that it could be a little faster (at [machine level](https://stackoverflow.com/questions/12135518/is-faster-than)) and for human reading.
</blockquote></details>

On modern architectures, there should not be any difference in their performance - and even if there were a single clock cycle difference, this is not where any perceivable performance problems come from. In practice, readability matters more.
Submitting this change anyway as neutral on this issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10045)*

PR #10009, opened 2017-09-15
Add role="grid" to grid to support aria-multiselectable

knoobie:
Added role="grid" and role="treegrid" to Grid and Treegrid to support aria-multiselectable. (https://www.w3.org/TR/wai-aria/states_and_properties#aria-multiselectable)

The following has to be done now:

>"Authors SHOULD ensure that selected descendants have the aria-selected attribute set to true, and selectable descendant have the aria-selected attribute set to false. Authors SHOULD NOT use the aria-selected attribute on descendants that are not selectable."

__Done.__

Finally after a lot of try and error everything is working fine.

Sorry, I had to add an additional method (#isMultiSelectionAllowed()) to the SelectionModel, because Grid had no access to the Impl classes and aria-multiselectable needs it. So it's kinda a breaking? change?

Additional, how do you want to continue with the following?

>"A treegrid is considered editable unless otherwise specified. To make a treegrid read-only, set the aria-readonly attribute of the treegrid to true. The value of the treegrid element's aria-readonly attribute is implicitly propagated to all of its owned gridcell elements, and will be exposed through the accessibility API. An author may override an individual gridcell element's propagated aria-readonly value by setting the aria-readonly attribute on the gridcell." (https://www.w3.org/TR/wai-aria/roles#treegrid)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10009)

PR #10046, opened 2017-09-20
Add aria-rowcount to grid

knoobie:
> Depends on the merge of #10009. 

@tsuoanttila, this pull request adds aria-rowcount to the table element of the grid and allow the grid to update itself depending of the amount of rows in the grid. 

The following questions are still open in my mind:

* I'm not sure if the value of aria-rowcount should decrease / increase depending on a filter that is applied to the grid (currently it decreases and increases depending on the filter) or if it should always show the '**full**' ? table row count. 

* Should the header and footer rows be added to the count as well, or should this count only reflect the rows in tbody?

>If all of the rows are present in the DOM, it is not necessary to set this attribute as the user agent can automatically calculate the total number of rows. However, if only a portion of the rows is present in the DOM at a given moment, this attribute is needed to provide an explicit indication of the number of rows in the full table.

>Authors MUST set the value of aria-rowcount to an integer equal to the number of rows in the **full** table. If the total number of rows is unknown, authors MUST set the value of aria-rowcount to -1 to indicate that the value should not be calculated by the user agent.

Source: [WAI-ARIA 1.1#aria-rowcount](https://www.w3.org/TR/wai-aria-1.1/#aria-rowcount)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10046)


hesara:
This PR seems to include changes that have already been integrated and/or are waiting in the PR queue as separate PRs, and possibly some others as well. Would it be possible to split the relevant changes to their own PR(s) and add the missing tests?

---

Reviewed 1 of 1 files at r13.
Review status: 1 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[client/src/main/java/com/vaadin/client/widget/grid/selection/SelectionModel.java, line 131 at r13](https://reviewable.io:443/reviews/vaadin/framework/10046#-KvS7toMhaIAIut28gdt:-KvS7toMhaIAIut28gdu:b-ybs981) ([raw file](https://github.com/vaadin/framework/blob/97479249d0c393a4c805dca67ef3723225a7a541/client/src/main/java/com/vaadin/client/widget/grid/selection/SelectionModel.java#L131)):*
> ```Java
>      * @since 8.2
>      */
>     boolean isMultiSelectionAllowed();
> ```

Does this change belong to this PR?

---

*[client/src/main/java/com/vaadin/client/widget/treegrid/TreeGrid.java, line 48 at r13](https://reviewable.io:443/reviews/vaadin/framework/10046#-KvS6pCVt8OPX0MNJWnV:-KvS6pCVt8OPX0MNJWnW:b-edyl2q) ([raw file](https://github.com/vaadin/framework/blob/97479249d0c393a4c805dca67ef3723225a7a541/client/src/main/java/com/vaadin/client/widget/treegrid/TreeGrid.java#L48)):*
> ```Java
>      * Creates a new instance.
>      */
>     protected TreeGrid() {
> ```

Will this reduce the visibility of the only constructor?
Is this change related to aria-rowcount or an artifact from having this PR share the history of another PR?

---

*[client/src/main/java/com/vaadin/client/widgets/Escalator.java, line 1501 at r13](https://reviewable.io:443/reviews/vaadin/framework/10046#-KvS7fh0UtBaLQ2Pwt5h:-KvS7fh0UtBaLQ2Pwt5i:b2379po) ([raw file](https://github.com/vaadin/framework/blob/97479249d0c393a4c805dca67ef3723225a7a541/client/src/main/java/com/vaadin/client/widgets/Escalator.java#L1501)):*
> ```Java
>         }
> 
>         private void updateAriaRowcountBasedOnRows(Integer rows){
> ```

There don't seem to be any tests for aria-rowcount.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10046#-:-KvS7kKfm5O_BOZd7AjS:b-w143ge)*

knoobie:
Hey @hesara this PR is branched from the other open PR because it depends on some API changes of the other PR. 

I think this PR should be re-written after the other PR gets approved. 
It was a starting point after a mail conversation with @tsuoanttila,

hesara:
The relevant changes from this could now e.g. be picked on top of master as a new PR to have a clear separation between the changes.

knoobie:
I think we can close this pr for now, I'm going to re-create it in a new PR this weekend (hopefully) to implement the feature with all the information I have received from the wai aria team. (How it should be done, see questions in the first msg)

PR #10126, opened 2017-10-02
static import of Assert in tests

asashour:
In some parts of the tests, `static import` is used, and in other places, it is not.

This change unifies all to have `static import`.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10126)


hesara:
An initial observation: some tests still extend the deprecated class `junit.framework.TestCase` rather than using the newer `@org.junit.Test` annotation on test methods, and inherit such methods from `TestCase`. Thus, this change may change the exception thrown by such classes. I hope this does not break tests, but it is hard to be 100% certain without running the full set of tests including all the browser tests.

Ideally, assuming that this change does not break anything, we could submit this and then there could be a follow-up that removes the use of the old junit.framework classes. Note that this affects all subclasses of classes inheriting `TestCase` which will need the `@Test` annotation.

---

Reviewed 133 of 733 files at r1, 1 of 1 files at r2.
Review status: 134 of 733 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10126#-:-KvSSeR4LJd5NAkzfse0:b-jiotyl)*

hesara:
Reviewed 577 of 733 files at r1.
Review status: 711 of 733 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10126)*

asashour:
Only `AbstractFilterTestBase` (and its sub-classes) extends `junit.framework.TestCase`. And all methods already has `@test` annotation.

Removal of the dependency is committed.

Even if we have a failing test detected later, we are here to fix :)

---

Review status: 711 of 735 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10126#-:-KvW0RbfTlJJ3ukv6kgZ:bgud805)*

hesara:
:lgtm:

---

Reviewed 15 of 733 files at r1, 9 of 9 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10126#-:-KvW4jPTZYDXb6EH1FDU:bnfp4nl)*

PR #10128, opened 2017-10-02
Break iteration when method is found

asashour:
Since we are searching for a specific method, `break` the loop when it is found.

Also, since `method` can not be null, then no need for a local variable.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10128)

### DIFF
--- a/server/src/main/java/com/vaadin/event/EventRouter.java
+++ b/server/src/main/java/com/vaadin/event/EventRouter.java
@@ -135,6 +135,7 @@ public class EventRouter implements MethodEventSource {
         for (Method m : methods) {
             if (m.getName().equals(methodName)) {
                 method = m;
+                break;
             }
         }
         if (method == null) {
@@ -152,7 +153,6 @@ public class EventRouter implements MethodEventSource {
                 }
             }
         }
-
     }
 
     /**
@@ -209,7 +209,6 @@ public class EventRouter implements MethodEventSource {
                     listenerMethod.receiveEvent(event);
                 }
             }
-
         }
     }
 
--- a/server/src/main/java/com/vaadin/event/ListenerMethod.java
+++ b/server/src/main/java/com/vaadin/event/ListenerMethod.java
@@ -243,10 +243,10 @@ public class ListenerMethod implements EventListener, Serializable {
             throws IllegalArgumentException {
 
         // Finds the correct method
-        Method method = null;
         for (Method m : target.getClass().getMethods()) {
             if (m.getName().equals(methodName)) {
                 method = m;
+                break;
             }
         }
         if (method == null) {
@@ -271,7 +271,6 @@ public class ListenerMethod implements EventListener, Serializable {
 
         this.eventType = eventType;
         this.target = target;
-        this.method = method;
         this.arguments = arguments;
         this.eventArgumentIndex = eventArgumentIndex;
     }
@@ -355,6 +354,7 @@ public class ListenerMethod implements EventListener, Serializable {
         for (Method m : target.getClass().getMethods()) {
             if (m.getName().equals(methodName)) {
                 method = m;
+                break;
             }
         }
         if (method == null) {
@@ -364,7 +364,6 @@ public class ListenerMethod implements EventListener, Serializable {
 
         this.eventType = eventType;
         this.target = target;
-        this.method = method;
         this.arguments = arguments;
         eventArgumentIndex = -1;
     }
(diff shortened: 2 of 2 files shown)


hesara:
While I agree that it makes no sense to always iterate through the whole collection without even using the information about the other entries, I'm not sure if the change in semantics can break something or not (last match vs. first match). While the javadoc forbids this and states that an exception is thrown, none of the implementations does this. Thus, this change could break a (non-compliant) caller when the last method found in the class would be valid with respect to its parameters whereas the first one isn't. Requesting a second opinion from @tsuoanttila before merging this.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10128#-:-KvS-k83EjHgs-ft3QRF:b-zfuqiy)*

tsuoanttila:
I checked some things about these modified methods and I think it should be pretty safe to merge it. The methods modified are accessed through APIs deprecated in `7.0` so no one should run into issues with them.

---

Reviewed 2 of 2 files at r1.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 129 at r1](https://reviewable.io:443/reviews/vaadin/framework/10128#-KvWOn2pMXkXOcZN9uoJ:-KvWOn2qweDZQ4MbfaZc:bwxjhr5) ([raw file](https://github.com/vaadin/framework/blob/d8dc83a895ee03c1876e33a52071ca84099b1f52/server/src/main/java/com/vaadin/event/EventRouter.java#L129)):*
> ```Java
>      */
>     @Override
>     public void removeListener(Class<?> eventType, Object target,
> ```

This method is only called from `AbstractClientConnector::removeListener` variant that is marked as `@deprecated As of 7.0.`. In my opinion this should not be a blocker, instead we should think about removing the whole method to begin with.

---

*[server/src/main/java/com/vaadin/event/ListenerMethod.java, line 241 at r1](https://reviewable.io:443/reviews/vaadin/framework/10128#-KvWQ98DRoIWHM4rcR46:-KvWQ98DRoIWHM4rcR47:bsdl1fa) ([raw file](https://github.com/vaadin/framework/blob/d8dc83a895ee03c1876e33a52071ca84099b1f52/server/src/main/java/com/vaadin/event/ListenerMethod.java#L241)):*
> ```Java
>      *             <code>target</code>.
>      */
>     public ListenerMethod(Class<?> eventType, Object target, String methodName,
> ```

This Constructor is never called from the Framework code.

---

*[server/src/main/java/com/vaadin/event/ListenerMethod.java, line 350 at r1](https://reviewable.io:443/reviews/vaadin/framework/10128#-KvWQF0iHNS77KxHAsLH:-KvWQF0iHNS77KxHAsLI:bsdl1fa) ([raw file](https://github.com/vaadin/framework/blob/d8dc83a895ee03c1876e33a52071ca84099b1f52/server/src/main/java/com/vaadin/event/ListenerMethod.java#L350)):*
> ```Java
>      *             <code>object</code>.
>      */
>     public ListenerMethod(Class<?> eventType, Object target, String methodName,
> ```

This Constructor is never called from the Framework code.

---

*[server/src/main/java/com/vaadin/event/ListenerMethod.java, line 445 at r1](https://reviewable.io:443/reviews/vaadin/framework/10128#-KvWQizhZM841ln3uGVf:-KvWQizhZM841ln3uGVg:bvnncs2) ([raw file](https://github.com/vaadin/framework/blob/d8dc83a895ee03c1876e33a52071ca84099b1f52/server/src/main/java/com/vaadin/event/ListenerMethod.java#L445)):*
> ```Java
>      *            the method or it contains more than one matching methods
>      *            <code>java.lang.IllegalArgumentException</code> is thrown.
>      * @throws IllegalArgumentException
> ```

Highlight. Same JavaDoc is in each of the modified methods.

---

*[server/src/main/java/com/vaadin/event/ListenerMethod.java, line 449 at r1](https://reviewable.io:443/reviews/vaadin/framework/10128#-KvWQNBsNf4Zy3C3pNBw:-KvWQNBsNf4Zy3C3pNBx:b-v3dgtd) ([raw file](https://github.com/vaadin/framework/blob/d8dc83a895ee03c1876e33a52071ca84099b1f52/server/src/main/java/com/vaadin/event/ListenerMethod.java#L449)):*
> ```Java
>      *             <code>target</code>.
>      */
>     public ListenerMethod(Class<?> eventType, Object target, String methodName)
> ```

This constructor is called by the Framework. From `EventRouter::addListener` which is in turn called by the counterpart of the deprecated remove method in `AbstractClientConnector`. Same arguments apply.

The JavaDoc says this should throw `IllegalArgumentException` if none or many methods are found.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10128#-:-KvWRf7AkJHjhlCkM9jH:b-ge7ck6)*

PR #10130, opened 2017-10-02
Registration to remove state.registeredEventListeners (fixes #9634)

### LINKED ISSUE #9634: EventListeners are not removed from State
It's done in `ComponentStateUtil.removeRegisteredEventListener()` but most of `Registration.remove()` and `removeListener(???)` methods do not do that. That means the listeners are not unregistered properly at client-side, and, thus, sometimes events (ex. context click events) stay not bubbled even if all the listeners for the event type was removed.

asashour:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10130)


hesara:
Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 69 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvSZS8eXu8AHVApnSWW:-KvSZS8fCiOiLf3_pAQ6:b-xrq9or) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L69)):*
> ```Java
> 
>     /**
>      * <p>
> ```

In javadoc, I believe having both the opening and the closing of a `<p>` and having an opening `p` for the first paragraph leads to some tools complaining about invalid first sentence or parsing it incorrectly. As far as I know, in javadoc, only the opening `p` tags are needed, and they aren't needed in front of the first paragraph.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvS_bsE6tFBKr8Opz5o:b-vleg1t) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*
> ```Java
>             needRepaint = true;
>         } else {
>             needRepaint = !hasListeners(eventType);
> ```

Can listeners be added for the same eventType using this method and the one above? Probably not in practice, but if they can, call order would determine if needRepaint is true or false, which might lead to problems.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

asashour:
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 69 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvSZS8eXu8AHVApnSWW:-KvSziy90D4vYpPNbA4s:bs4xtag) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L69)):*

In javadoc, I believe having both the opening and the closing of a `<p>` and having an opening `p` for the first paragraph leads to some tools complaining about invalid first sentence or parsing it incorrectly. As far as I know, in javadoc, only the opening `p` tags are needed, and they aren't needed in front of the first paragraph.
</blockquote></details>

Done. It was actually a copy of the parent javadoc, which is now also changed.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvT0OEAsD_aFsgzYDEn:b-ung3ec) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*

Can listeners be added for the same eventType using this method and the one above? Probably not in practice, but if they can, call order would determine if needRepaint is true or false, which might lead to problems.
</blockquote></details>

I would say, "not in practice".

I believe the original logic of having a Set of state, where many listeners are added is not optimal.

`SharedState.registeredEventListeners` has been be a Map, with a value of the number of event identifier listeners.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

hesara:
Reviewed 6 of 6 files at r2.

---

*[client/src/main/java/com/vaadin/client/ui/AbstractConnector.java, line 502 at r2](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvVl9djYUGmz1DvPGLd:-KvVl9djYUGmz1DvPGLe:b-6g2c1b) ([raw file](https://github.com/vaadin/framework/blob/816ec0a75cd0e869d30708bea18abf65cde39aae/client/src/main/java/com/vaadin/client/ui/AbstractConnector.java#L502)):*
> ```Java
>     public boolean hasEventListener(String eventIdentifier) {
>         Map<String, Integer> reg = getState().registeredEventListeners;
>         return (reg != null && reg.keySet().contains(eventIdentifier));
> ```

this could use reg.containsKey(...), which can be more efficient for many map types

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvVkztvnu9ZlTHgJfvd:b-x7vmn1) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*

I would say, "not in practice".

I believe the original logic of having a Set of state, where many listeners are added is not optimal.

`SharedState.registeredEventListeners` has been be a Map, with a value of the number of event identifier listeners.
</blockquote></details>

This makes more sense from the performance point of view (at the cost of the possibility to validate some broken situations), but should be mentioned under backwards incompatible changes in the release notes as it is possible some add-on has checked the state field (which is public) directly rather than through the standard API.

There should probably be a note in the javadoc of this method not to use the old removeListener methods together with this but only remove it via the Registration returned by addListener(). I'm not sure what would be the result of someone calling eventRouter.removeAllListeners() with this change, but at least that is not more broken now than before.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 75 at r2](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvVnM9X9yeNxdoPjXuR:-KvVnM9X9yeNxdoPjXuS:b-btjwqg) ([raw file](https://github.com/vaadin/framework/blob/816ec0a75cd0e869d30708bea18abf65cde39aae/server/src/main/java/com/vaadin/event/EventRouter.java#L75)):*
> ```Java
>      *
>      * <p>
>      * This method additionally informs the event-api to stop routing events
> ```

(Looking at this and the original documentation in ACC, this is not very clear but I guess we can live with this.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

asashour:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[client/src/main/java/com/vaadin/client/ui/AbstractConnector.java, line 502 at r2](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvVl9djYUGmz1DvPGLd:-KvVvnYUbyHN5foPTeTM:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/816ec0a75cd0e869d30708bea18abf65cde39aae/client/src/main/java/com/vaadin/client/ui/AbstractConnector.java#L502)):*

this could use reg.containsKey(...), which can be more efficient for many map types
</blockquote></details>

Done.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvVwUFoNqqQdK-a6ezJ:b-k3p8b4) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*

This makes more sense from the performance point of view (at the cost of the possibility to validate some broken situations), but should be mentioned under backwards incompatible changes in the release notes as it is possible some add-on has checked the state field (which is public) directly rather than through the standard API.

There should probably be a note in the javadoc of this method not to use the old removeListener methods together with this but only remove it via the Registration returned by addListener(). I'm not sure what would be the result of someone calling eventRouter.removeAllListeners() with this change, but at least that is not more broken now than before.
</blockquote></details>

Done. Added a release note. Regarding removing listener, it is already handled in `AbstractClientConnector.removeListener(String, Class, Object)`.  

This way, when the user add/removeListener from `AbstractClientConnector` with event identifier, it works.

The `EventRouter.addListener(Class, Object, Method, String, SharedState)` doesn't have a corresponding `removeListener`, since we depend only on Registration.

By the way, `EventRouter.removeAllListeners` is not called,  And since `AbstractClientConnector.eventRouter` is private without public/protected access, I believe it should be deprecated. The only potential usage if an add-on uses it.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 75 at r2](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvVnM9X9yeNxdoPjXuR:-KvVyiXdnpP_5x7zlgBY:b2nhll9) ([raw file](https://github.com/vaadin/framework/blob/816ec0a75cd0e869d30708bea18abf65cde39aae/server/src/main/java/com/vaadin/event/EventRouter.java#L75)):*

(Looking at this and the original documentation in ACC, this is not very clear but I guess we can live with this.)
</blockquote></details>

Ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

hesara:
Reviewed 2 of 2 files at r3.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvW-0Zn7mG3gwecZc2k:b428021) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*

Done. Added a release note. Regarding removing listener, it is already handled in `AbstractClientConnector.removeListener(String, Class, Object)`.  

This way, when the user add/removeListener from `AbstractClientConnector` with event identifier, it works.

The `EventRouter.addListener(Class, Object, Method, String, SharedState)` doesn't have a corresponding `removeListener`, since we depend only on Registration.

By the way, `EventRouter.removeAllListeners` is not called,  And since `AbstractClientConnector.eventRouter` is private without public/protected access, I believe it should be deprecated. The only potential usage if an add-on uses it.
</blockquote></details>

EventRouter is also used in other locations than in AbstractClientConnector (such as in Binder and in AbstractDataProvider) and probably even in some add-ons. Thus, I don't think we can really remove API from it without a good reason and a replacement.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

asashour:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvW1WNxkcr_ywRyiReD:b-54b752) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*

EventRouter is also used in other locations than in AbstractClientConnector (such as in Binder and in AbstractDataProvider) and probably even in some add-ons. Thus, I don't think we can really remove API from it without a good reason and a replacement.
</blockquote></details>

Ok, now back to the `removeAllListeners`, the `EventRouter` does not have a reference to the `SharedState`.

So, whoever adds a listener to `EventRouter` has to remove it, and `removeAllListeners` works fine.

So, should we add `SharedState` to the `EventRouter` and change `removeAllListeners`, or we keep things as currently is?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

hesara:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[server/src/main/java/com/vaadin/event/EventRouter.java, line 113 at r1](https://reviewable.io:443/reviews/vaadin/framework/10130#-KvS_bsE6tFBKr8Opz5n:-KvW5VXd140mckX4yivh:b-pgh66r) ([raw file](https://github.com/vaadin/framework/blob/ba6ebb2969a1b353933bd2509ac419177c59cd2a/server/src/main/java/com/vaadin/event/EventRouter.java#L113)):*

Ok, now back to the `removeAllListeners`, the `EventRouter` does not have a reference to the `SharedState`.

So, whoever adds a listener to `EventRouter` has to remove it, and `removeAllListeners` works fine.

So, should we add `SharedState` to the `EventRouter` and change `removeAllListeners`, or we keep things as currently is?
</blockquote></details>

I think a mention in the javadoc that the state registered listener updates by this method must be removed using the Registration, and removeAllListeners() etc. do not clean them up is probably enough as removeAllListeners() is not used within the framework. That just manages the expectations of whoever is using EventRouter for other purposes and might consider calling this method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

hesara:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

hesara:
Now there is a conflict (probably a trivial one) to resolve.

hesara:
Reviewed 2 of 2 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10130)*

PR #10149, opened 2017-10-07
Add aria-rowcount to grid

knoobie:
This pr adds basic support for [aria-rowcount](https://www.w3.org/TR/wai-aria-1.1/#aria-rowcount). 

It's based on discussion from https://github.com/vaadin/vaadin-grid/pull/1023 and exceeds #10046.

 Please tell me what kind of Javadoc you wish for Escalator#AriaGridHelper.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10149)


hesara:
This seems to cause Reviewable to melt down as it sees 6 + 1984 changed files. Based on a quick look, this looks ok to me apart from the javadoc and perhaps checking if some constructor changes can break anything.

knoobie:
@hesara I can create overloading constructors so the old behaviour stays the same and add a simple if(ariaGridHelper != null) before the statements that's add or remove the rows in the AriaGridHelper. 

Should I re-create the pull request (after deleting my fork and forking the current master again) to make reviewable useable again?

hesara:
Rebasing the changes might ease the review if not too much trouble for you. I'm not sure if the constructors are an issue - perhaps @tsuoanttila has an opinion on that, although if the backwards compatible change is simple, it might be worth it to reduce the risk of breaking some add-on.

knoobie:
I will take a look at the rebasing. Thinking about the constructors.. well anything is in Escalator.java - why did I even added the AriaGridHelper in the constructor? Any InnerClass should have access to the field, so I can remove the constructor changes and call the ariaGridHelper field of the class if you don't mind!

hesara:
Superceded by #10167.

PR #10157, opened 2017-10-09
Add binder test to verify that the 8.1 behaviour is still working with 8.2

knoobie:
Hey I stumbled across a weird bug: after editing a value in the current snapshot release all validations of a binder are triggered. Is this a bug/feature? I really don't know. 

I added a test that should work with 8.2.. but currently it fails because password is invalid after the user adds ONLY the username.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10157)


knoobie:
Well 'success' - I can reproduce my problem with this test. 

I would have expected that the password has no error indication, but is has. (See BinderFormValidateTest.java:41)

tsuoanttila:
The part of `Binder` that needs fixing is https://github.com/vaadin/framework/blob/master/server/src/main/java/com/vaadin/data/Binder.java#L1502

Basically the old way was that every binding run validation only on itself, then triggered write to bean, run bean validation without any reverting mechanism. I did some major clean up in that logic and made the saving always use the same path. There's a `TODO` comment about partial writes in the `doWriteIfValid`. The reason I did not solve it immediately was the fact that it could use some new API in the public interface `Binding` instead of relying on implementation internals. But at the same time we don't want to fire extra events. I think the API we want to add is `isValid` in a similar manner as in the `Binder`. (See `Binder::isValid` vs. `Binder::validate`)

tsuoanttila:
Follow up comment: There's already ground work in place and the `Set` of `Binding`s provided to the method are the ones we have changes in and try to save. So running the validation only to that set and then processing only changes for that set should do the trick. After that change, the logic of when to clean up the changed bindings needs some re-thinking.

tsuoanttila (commented):
This could also be tested with unit test on the server-side only.

knoobie:
@tsuoanttila Do you think this could be fixed this week? Otherwise I have to do a rollback and create my own version of 8.1 with my other grid related pull requests to test them next week.

knoobie:
Could this already fix it? Sorry, don't have time to dig deeper into the binder logic. bildschirmfoto 2017-10-10 um 10 46 59

tsuoanttila:
The issue with that fix is that it will fire an extra set of validation status events (which is why I was thinking adding an alternative API for validations in `Binding`). The `bindings.clear` call might be problematic with some cases, make sure that it is always a modifiable collection and it is fine to clear it.

knoobie:
I think I know what you mean. Do you think we can go for a 'quick fix' at the moment and optimize it later on for 8.2? We kinda depend on an asap 'fix' to test the version with the grid changes properly.

tsuoanttila:
@knoobie I took part of your solution, polished it around a bit and changed to a full server-side unit test. See PR #10162 and tell me if it fixes your problem as well.

tsuoanttila:
Closing this as the related PR with a server test has already been merged.

PR #10162, opened 2017-10-10
Fix Binder bean writing to only validate and write given bindings

tsuoanttila:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10162)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -118,14 +118,31 @@ public class Binder<BEAN> implements Serializable {
 
         /**
          * Validates the field value and returns a {@code ValidationStatus}
-         * instance representing the outcome of the validation.
+         * instance representing the outcome of the validation. This method is a
+         * short-hand for calling {@link #validate(boolean)} with
+         * {@code fireEvent} {@code true}.
          *
+         * @see #validate(boolean)
          * @see Binder#validate()
          * @see Validator#apply(Object, ValueContext)
          *
          * @return the validation result.
          */
-        public BindingValidationStatus<TARGET> validate();
+        public default BindingValidationStatus<TARGET> validate() {
+            return validate(true);
+        }
+
+        /**
+         * Validates the field value and returns a {@code ValidationStatus}
+         * instance representing the outcome of the validation.
+         *
+         * @see #validate()
+         * 
+         * @param fireEvent
+         *            {@code true} to fire status event; {@code false} to not
+         * @return the validation result.
+         */
+        public BindingValidationStatus<TARGET> validate(boolean fireEvent);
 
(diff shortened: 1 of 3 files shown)


knoobie:
If the teamcity build passes, this should fix our problem as well! Thank you for looking into this problem so fast! 

Should I close my pull request or do you want to have a client test as well? (just to be sure nothing breaks with the new Implementation)

hesara:
Reviewed 3 of 3 files at r2.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 143 at r2](https://reviewable.io:443/reviews/vaadin/framework/10162#-Kw5yekd7TN-mFFtbdT5:-Kw5yekesdzipU4YLdsb:btrsgnj) ([raw file](https://github.com/vaadin/framework/blob/049288b65366d85675a350e1a7250da817f86fb8/server/src/main/java/com/vaadin/data/Binder.java#L143)):*
> ```Java
>          * @param fireEvent
>          *            {@code true} to fire status event; {@code false} to not
>          * @return the validation result.
> ```

since tag

---

*[server/src/test/java/com/vaadin/data/BinderTest.java, line 964 at r2](https://reviewable.io:443/reviews/vaadin/framework/10162#-Kw6-pxrmWY0vax6vH8C:-Kw6-pxrmWY0vax6vH8D:bci75rg) ([raw file](https://github.com/vaadin/framework/blob/049288b65366d85675a350e1a7250da817f86fb8/server/src/test/java/com/vaadin/data/BinderTest.java#L964)):*
> ```Java
>                 nameField.getComponentError());
>         assertNull(
>                 "Age field should not be in error, since it has not been modified.",
> ```

Whether this should really happen on a field-by-field basis or as a global reaction is a bit unclear to me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10162)*

hesara:
Reviewed 1 of 1 files at r3.

---

*[server/src/test/java/com/vaadin/data/BinderTest.java, line 964 at r2](https://reviewable.io:443/reviews/vaadin/framework/10162#-Kw6-pxrmWY0vax6vH8C:-Kw9up7ytflr9EJe3THy:bujhf3e) ([raw file](https://github.com/vaadin/framework/blob/049288b65366d85675a350e1a7250da817f86fb8/server/src/test/java/com/vaadin/data/BinderTest.java#L964)):*

Whether this should really happen on a field-by-field basis or as a global reaction is a bit unclear to me.
</blockquote></details>

(making this a non-blocker, for others to decide based on feedback)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10162)*

knoobie:
FYI: Tested with our application and everything works like in 8.1! Thank you!

PR #10212, opened 2017-10-19
Fix DateField requires two clicks to open calendar (Fixes #9764)

### LINKED ISSUE #9764: DateField sometimes requires two clicks to open calendar if an invalid string is typed by the user.
Related to #9740 
After merging #9740 effect is visible in _DateFieldElementUI_

asashour:
The reason was: textfield change triggers an event, sent to the server, which sets `ErrorMessage`, and the client-side changes the element to add the error message.

At the same time, there is `click` event on the button, which gets lost, because of the element removal from the hierarchy.

Suggested solution: add a new interface `Orphanable`, and explicitly wait for 100 milliseconds, and the widget sees the 'click' and handles it after re-added to the parent.

Since this may be unique to `VAbstractPopupCalendar`, we may directly check that in Slot, with direct dependency between `Slot` and `VAbstractPopupCalendar`, without the need for `Orphanable`.

Another thing, I couldn't find another way to wait for the `click` event, except by `Scheduler.get().scheduleFixedDelay`, other methods in `Scheduler` are triggered too early.

Lastly, `VAbstractPopupCalendar` is not a `ClickHandler` anymore, and if subclasses need to be notified, they can use `calendarToggle.addClickHandler()`

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10212)


asashour:
I need help.  The tests runs locally with `PhantomJS` 1.9.7, I am not sure if the issue is with Windows/Linux, JDK version, or `local` vs `hub` configurations.

I tried to run the test remotely, but couldn't understand how to run a `hub` server on port `4444` as required by `ParallelTest.getHubURL()`, which is overridden in `PrivateTB3Configuration`

Issue #9713, opened 2017-07-21
Feature: Use slf4j-api as logging frontend


jreznot:
The Simple Logging Facade for Java (SLF4J) serves as a simple facade or abstraction for various logging frameworks (e.g. java.util.logging, logback, log4j) allowing the end user to plug in the desired logging framework at deployment time.

Note that SLF4J-enabling your library implies the addition of only a single mandatory dependency, namely slf4j-api.jar. If no binding is found on the class path, then SLF4J will default to a no-operation implementation.

SLF4J is just API, so libraries can use it safely, app servers will not conflict with it

jreznot:
See also old issue on SLF4J and TomEE: https://github.com/vaadin/framework/issues/3885

PR #10267, opened 2017-10-28
Use slf4j-api as a logging frontend (#9713)

jreznot:
Introduced new dependencies for server:

org.slf4j:slf4j-api
org.slf4j:slf4j-jdk14

Replaced custom 2.4.11.vaadin2 version of
org.atmosphere:atmosphere-runtime with original 2.4.11.

For client side introduced small implementation of SLF4J that uses GWT
JUL as backend: ru.finam:slf4j-gwt.

All usages of java.util.logging.Logger were replaced with SLF4J API.
There are a couple of places where it is still used, such as client side
classes. Since GWT uses JUL as backend everything should work as before.

Now, applications can easily exclude slf4j-jdk14. They can use any
available SLF4J binding, such as slf4j-simple or logback-classic.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10267)

### DIFF
--- a/client-compiler/pom.xml
+++ b/client-compiler/pom.xml
@@ -28,6 +28,16 @@
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-sass-compiler</artifactId>
         </dependency>
+
+        <!-- Logging Facade -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+        </dependency>
     </dependencies>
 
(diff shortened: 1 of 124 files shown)


jreznot:
I think that it is a good time to introduce SLF4J in 8.2, since it is in development right now and only 8.2.0.alpha2 just released

tsuoanttila:
Hi @jreznot and thanks for the PR! Unfortunately we're not going to merge this change for Vaadin 8. We had a discussion with @legioth about this change and came to a conclusion that making a change this major in a minor release is not something we should do. 

Please keep tracking the Flow issue, and let's hope the development team there has time to make the change.

jreznot:
I understand the trade-off, just was curious if it can be done in the upstream. 
Hope, we will be able to maintain this change in our fork )

PR #10271, opened 2017-10-30
Changed retrieval of names of nested properties (#10159)

schaerl:
The getName() function in NestedBeanPropertyDefinition will return only the top level name, which causes problems when adding a nested property that has the same name as a non-nested (e.g "value" & "sub.value")

This change should fix this behaviour by changing the return value of "getName()" of NestedBeanPropertyDefinition and introducing a function "getTopLevelName()" that only returns the last part of the full name for where it's needed

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10271)

Issue #10243, opened 2017-10-24
Adding a column with a nested property to a grid will cause an IllegalArgumentException if the page is reloaded.


schaerl:
https://github.com/vaadin/framework/pull/8916 allows the usage of nested properties in grids, however it breaks when you add a nested property column and then refresh the page. Here is a short example:

```
public class MyUI extends UI {
    @Override
    protected void init(VaadinRequest vaadinRequest) {
        Grid<Person> grid = new Grid(Person.class);
        grid.addColumn("address.street");
        setContent(grid);
    }

    private class Person {
        private String name;
        private Address address;

        public Person(String name, Address adr) {
            this.name = name;
            this.address = adr;
        }

        public String getName() {
            return name;
        }

        public void setName(String name) {
            this.name = name;
        }

        public Address getAddress() {
            return address;
        }

        public void setAddress(Address address) {
            this.address = address;
        }

    }

    private class Address {
        private String street;

        public Address(String str) {
            this.street = str;
        }

        public String getStreet() {
            return street;
        }

        public void setStreet(String street) {
            this.street = street;
        }

    }

    @WebServlet(urlPatterns = "/*", name = "MyUIServlet", asyncSupported = true)
    @VaadinServletConfiguration(ui = MyUI.class, productionMode = false)
    public static class MyUIServlet extends VaadinServlet {
    }

}
```

Running this code and opening the page will show a grid with the columns Address, Name and Street. When reloading(F5) the page, an IllegalArgumentException will be thrown. 

When reloading the page, the grid is constructed a second time, but since the propertySet is already present, it will be reused (c.f. `public static <T> PropertySet<T> get(Class<? extends T> beanType)` in BeanPropertySet).
Since the constructor creates columns for each property definition (using its `getName()` function as input to `addColumn()`), it will try to create one for the street property as well. But `getName()`for this property will return `street` which will not be found in the Person bean and therefore the exception is thrown

schaerl:
I realized that this seems to be a duplicate of https://github.com/vaadin/framework/issues/9960 and it also seems related to https://github.com/vaadin/framework/issues/10159 and https://github.com/vaadin/framework/issues/9609

TatuLund:
There is comment in https://github.com/vaadin/framework/issues/9960 claiming this is fixed in 8.3.0

TatuLund:
Yes, this fixed by https://github.com/vaadin/framework/pull/9925

schaerl:
[...]

PR #10272, opened 2017-10-30
Change which columns of a grid are created when using the constructor "Grid(PropertySet<T> propertySet, DataCommunicator<T> dataCommunicator)" (#10243)

schaerl:
The Grid constructor will create columns for all properties it finds in the property set of the Bean it uses. When adding a nested property (e.g. "id.name") manually, the property set will be extended. On page refresh (F5), the constructor will try to find a property based on the top level name of the nested property (i.e. "name"), which will fail. If this would not fail, the adding of the column later on would fail because the column would already exist.

This change remedies this by not creating nested properties, even if they are in the property set.

The added UI test can be used to check the change. Without the change, opening the test UI in a browser will show a grid with 3 columns ("name", "address", "street") and refreshing the page will cause an exception, saying that the property "street" does not exist on line 13. 
With the change, first it will create the same 3 columns as without the change, but on page refresh it will only show the columns("name", "address") without crash.

**EDIT**: Since the inclusion of #10271 this is not completely correct anymore. See the comment further down.

**EDIT2**: The new issue this fixes is https://github.com/vaadin/framework/issues/10590

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10272)


schaerl:
Because #10271 changed how nested properties are named, the problem has changed as well. Now, the nested properties are found correctly when checking the property set, and the constructor won't crash anymore. However, when the code that originally added the column is called again, obviously the column will already exist and an exception is thrown.

The change still fixes this behaviour and the test has been changed to show this: 
- Without the change, the UI can be opened once without problems, showing a grid with "Address", "Name", "Street". On page refresh (F5) it will fail, because it tries to add a column for a property that already has one.
- With the change, a refresh will work fine and all columns will show.

elmot:
Having only test UI  is not enough to verify the change, an actual TestBench test is needed. 
Moreover, the change looks like server-side only, please make a test in server/src/test/java/com/vaadin/tests/components/grid

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10272#-:-L4eQwHcJ-pebgK9MjGP:bj9iu8k)*

schaerl:
Since I don't have any experience with TestBench, I instead constructed a Unit test which fails without the change and doesn't with it. It shows the same issue without refreshing the page.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10272#-:-L4jfY0nGeJGR1f08Usf:b-rpz8w5)*

elmot:
Thank you for pointing us to the problem, it's exist and is quite important but the way to fix is is not correct. The problem is not in property naming but in the fact that the same instance of `BeanPropertySet` might be used multiple times, and all the added property definitions are accumulated together and may affect several grids and binders. The proper fix of the problem should look like building a copy of `BeanPropertySet` when it's retrieved from the cache instead of reusing of the same instance. And tests should cover reusing of the automatic binding by both Binders and DataSources.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10272#-:-L5JDZeIvchVDRwzQ3o6:ber4qp)*

elmot:
Closed in favor of #10635

Issue #10252, opened 2017-10-25
Binder.setReadOnly(false) makes fields with no setter writeable


aley2003:
Vaadin 8.1.5
A call of Binder.setReadOnly(false) sets all bound fields to read/write. 
This incudes fields wich setter == null. 
Fields bound with no setter should always be read only.

mcollovati:
Should field be read only even when binding to a property with non accessible setter or without a setter?

Legioth:
My gut feeling is that this should work so that `setReadOnly(false)` would throw `IllegalStateException` in cases when the binding can only be in a readonly mode.

aley2003:
If I bind a Component with setter = null, the field is readOnly per default.
If I call binder.setReadOnly(false) afterwards, the field gets read/write suprisingly. 
The create time rule used (fields with no setter are readOnly) will be ignored. That is the problem from my point of view.

An IllegalStateExceptions would make Binder.setReadOnly() useless in all cases with null Setters.

I would suggest to support the bind time rule (no setter ->readOnly) inBinder.setReadOnly().

Legioth:
The entire `setReadOnly()` method is indeed completely redundant for cases where only one mode can work.

It can be annoying if a method throws unchecked exceptions, but it can on the other hand be very confusing if calling `setReadOnly(false)` would be silently ignored.

aley2003:
We have Dialogs were some of the fields should be read only all the time and for some fields read only has to be switched on and off depending of the state of the Dialog. For this case a setReadOnly(false) enabling only the writable fields (with setter) would be very usefull.
I would not call it "silently ignored" if setReadOnly(false) does not enable fields without setter. bind() without setter does the same. It is a question of the contract for Binder.setReadOnly().

mcollovati:
IMO in a mixed mode (RO an RW fields) it makes sense that Binder.setReadonly(...) silently ignores bindings that are **explicitly** set as RO.

For example having a form that contains also a field whose readonly status is base on runtime conditions (eg user role), I would like having its readonly state managed by Binder only if it is no implicitly RO.
It would be nice if one can do that by simply binding the field with a null setter (or with an explicit bind method `Binding<BEAN, TARGET> bindReadonly(ValueProvider<BEAN, TARGET> getter)`), instead of having to manually manage `HasValue.setReadonly` for each field to avoid a potential `IllegalStateException`.

Something like
```
[...]

PR #10351, opened 2017-11-21
WIP: Treat fields as readonly when bound with null setter (#10252)

mcollovati:
Fields bound with a null setter or to a property
with no accessible setter will be marked as readonly
and would be ignore by Binder.setReadonly

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10351)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -33,6 +33,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -226,6 +227,10 @@ public class Binder<BEAN> implements Serializable {
          * TextField nameField = new TextField();
          * binder.forField(nameField).bind(Person::getName, Person::setName);
          * </pre>
+         *
+         * <p>
+         * <strong>Note:</strong> when a {@code null} setter is given the field will be
+         * marked as readonly by invoking ({@link HasValue::setReadOnly}.
          *
          * @param getter
          *            the function to get the value of the property to the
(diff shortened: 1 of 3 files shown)


mcollovati:
`DateTextHandlingTest` fails because `dateField` is bound with null setter but test code tries to fill it.
Should either make `dateField` RW invoking `dateField.setReadonly(false)` after binding in test UI or remove read only logic for null setter from `Binder.bind` introduced by this PR

asashour:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 233 at r1](https://reviewable.io:443/reviews/vaadin/framework/10351#-KzYmuMi_bO0uPNhJ6XE:-KzYmuMi_bO0uPNhJ6XF:b-4z1kvk) ([raw file](https://github.com/vaadin/framework/blob/0d58f14f165485624efe33cd4bd9700a6e691c8e/server/src/main/java/com/vaadin/data/Binder.java#L233)):*
> ```Java
>          * <p>
>          * <strong>Note:</strong> when a {@code null} setter is given the field will be
>          * marked as readonly by invoking ({@link HasValue::setReadOnly}.
> ```

No closing parenthesis, I guess it is better to remove the opening parenthesis.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 775 at r1](https://reviewable.io:443/reviews/vaadin/framework/10351#-KzYnQJP17g7IaSD4xsj:-KzYnQJP17g7IaSD4xsk:b-y7d69i) ([raw file](https://github.com/vaadin/framework/blob/0d58f14f165485624efe33cd4bd9700a6e691c8e/server/src/main/java/com/vaadin/data/Binder.java#L775)):*
> ```Java
>             Setter<BEAN, ?> setter = definition.getSetter().orElse(null);
>             if (setter == null) {
>                 Logger.getLogger(Binder.class.getName())
> ```

Please add `getLogger()` method, as in other classes, e.g. `ListenerMethod`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10351)*

mcollovati:
Closing in order to create two separate PR for the described behaviors

PR #10411, opened 2017-12-05
Pick changes for next 8.2 release

tsuoanttila:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10411)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -18,7 +18,7 @@
 
     <properties>
         <vaadin.spring.version>2.0.1</vaadin.spring.version>
-        <vaadin.testbench.version>5.1.0</vaadin.testbench.version>
+        <vaadin.testbench.version>5.1.1</vaadin.testbench.version>
         <vaadin.cdi.version>2.0.0</vaadin.cdi.version>
         <vaadin.context-menu.version>2.0.0</vaadin.context-menu.version>
     </properties>
--- a/client/src/main/java/com/vaadin/client/ui/FocusUtil.java
+++ b/client/src/main/java/com/vaadin/client/ui/FocusUtil.java
@@ -96,12 +96,27 @@ public class FocusUtil {
         return focusable.getElement().getTabIndex();
     }
 
+    /**
+     * Finds all the focusable children of given parent element.
+     * 
+     * @param parent
+     *            the parent element
+     * @return array of focusable children
+     * @since 8.1.7
+     */
     public static native Element[] getFocusableChildren(Element parent)
     /*-{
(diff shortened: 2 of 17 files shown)


SomeoneToIgnore:
Reviewed 3 of 3 files at r1, 3 of 3 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, 3 of 3 files at r6, 1 of 1 files at r7.

---

*[scripts/BuildHelpers.py, line 148 at r7](https://reviewable.io:443/reviews/vaadin/framework/10411#-L-aSRkNPfajj988onna:-L-aSRkNPfajj988onnb:b-x4w4il) ([raw file](https://github.com/vaadin/framework/blob/8d6a3f3648420f35c9be43ec561c58654f4ae265/scripts/BuildHelpers.py#L148)):*
> ```Python
> LABEL maintainer="FrameworkTeam"
> 
> COPY ./*.war /var/lib/jetty/webapps/
> ```

As a side note, why did you use COPY that early? It looks like you can cache lots of operations from below and reduce docker image creation time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10411)*

tsuoanttila:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---

*[scripts/BuildHelpers.py, line 148 at r7](https://reviewable.io:443/reviews/vaadin/framework/10411#-L-aSRkNPfajj988onna:-L-kArm-BwZgYHI_O24Y:bm1acxj) ([raw file](https://github.com/vaadin/framework/blob/8d6a3f3648420f35c9be43ec561c58654f4ae265/scripts/BuildHelpers.py#L148)):*
<details><summary><i>Previously, SomeoneToIgnore (Kirill Bulatov) wrote…</i></summary><blockquote>

As a side note, why did you use COPY that early? It looks like you can cache lots of operations from below and reduce docker image creation time.
</blockquote></details>

I think that the index-generate.sh is depending on the content being already in place in the jetty auto-deploy folder. You can discuss with @jtomass on further improvements to this if you're interested!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10411)*

jtomass:
@SomeoneToIgnore technically correct about the placement, but docker doesn't really care in what layer you are inserting the wars. In quick tests I didn't notice any speed differences if the COPY was changed around. 
The dockerfile could be compressed to save couple of layers and therefore seconds. But that wouldn't be noticeable considering that the war copy takes most of the time.

SomeoneToIgnore:
Nice to know, thanks for the experiment.

Issue #9000, opened 2017-04-03
BeanValidationBinder says the form is valid even though required (@NotNull) field is not filled


mstahv:
Vaadin 8.0.5

Example code to reproduce

    public static class Entity {

        @NotNull
        private String name;

        public void setName(String name) {
            this.name = name;
        }

        public String getName() {
            return name;
        }

        @Override
        public String toString() {
            return "Entity, name: " + name;
        }

    }

    final TextField name = new TextField();

    @Override
    protected void init(VaadinRequest vaadinRequest) {
        final VerticalLayout layout = new VerticalLayout();
        name.setCaption("Type your name here:");

        Entity entity = new Entity();
        Binder<Entity> b = new BeanValidationBinder<>(Entity.class);
        b.bindInstanceFields(this);

        b.setBean(entity);

        Button button = new Button("Check validity");
        button.addClickListener(e -> {
            ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
            Validator validator = factory.getValidator();
            Set<ConstraintViolation<Entity>> validate = validator.validate(entity);
            String msg = String.format("Binder.isValid: %s, JSR 303 valid: %s", b.isValid(), validate.isEmpty());
            Notification.show(msg);
        });

        layout.addComponents(name);
        layout.addComponent(button);

        setContent(layout);
    }

viydaag:
I might know why the @NotNull does not work on TextField. With new data binding API, TextField value cannot be null. It is always empty at the begining. So as a workaround, adding the @Size(min=1) does the trick.

tsuoanttila:
Indeed the default "null representation" in the case of TextField is one-way change only. There was a long discussion on wether it should change `null -> "" -> null` or not. If you want to use `NotNull`, you should use `binder.forField(textField).withNullRepresentation("")` which performs the change back from empty string to `null`. However this will mean that if the user actually tries to input an empty string, they really can't.

viydaag:
Thx for the precision

jcgueriaud:
I tried to fix this issue but I think it will break some stuff (it does fix NotNull issue but may break NotEmpty or Size min). And there are some differences between withNullRepresentaiton and the "default null representation" that I'm not sure to understand.

I'm not quite sure if it's possible to have exactly the "validation" between BeanValidator and JSR303 validation because of null representation. (but I think Vaadin validation should never be ok if JSR303 is not).

[...]

PR #10422, opened 2017-12-07
BeanValidationBinder and @NotNull, @NotEmpty validations

jcgueriaud:
I tried to fix this issue #9000 
I'm pretty sure that my pull is not enough and have to be changed because it causes breaking change.

Instead of trying to check the field value, if the field value equals the defaultValue then I check null.
Example:
An attribute @Size(min = 1) bound
- Vaadin (before): invalid (it checks "")
- JSR303 Validation: valid (attribute is null)
- Vaadin (with this modification): Valid (it checks null)
So it breaks some old behaviours.

Here some thoughts about that change (why I made this one):
- The main problem is that Vaadin validates an invalid bean (BeanValidator <> JSR303 Validation).
- If I understand it correctly, nullRepresentation is a converter. It convert null to "nullRepresentation" and "nullRepresentation" to null.
--> Instead of check if fieldValue (here "") is useless because the fieldValue will be null (not "").
(For example, asRequired don't check if the field value is null but if the value equals emptyValue.)

Perhaps the discussion should be done in #9000

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10422)


tsuoanttila:
Thank you for the pull request and pointing out the issue. I do apologise for being so late with the review. Can you consider my feedback and respond soon so we can get this to an upcoming release for testing?

---

Reviewed 3 of 3 files at r2.

---

*[server/src/main/java/com/vaadin/data/BeanValidationBinder.java, line 102 at r2](https://reviewable.io:443/reviews/vaadin/framework/10422#-L45Hx0wvRJeVLtLog_1:-L45Hx0wvRJeVLtLog_2:b9yhzxx) ([raw file](https://github.com/vaadin/framework/blob/ccac01df3ed8110b6491a0d4f7ad612901e118ef/server/src/main/java/com/vaadin/data/BeanValidationBinder.java#L102)):*
> ```Java
>         if (requiredConfigurator != null) {
>             configureRequired(binding, definition, validator);
>             validator.setEmptyValue(binding.getField().getEmptyValue());
> ```

We might be able to skip this.

---

*[server/src/main/java/com/vaadin/data/validator/BeanValidator.java, line 121 at r2](https://reviewable.io:443/reviews/vaadin/framework/10422#-L45GdeSH1H7Twfftc9m:-L45GdeSH1H7Twfftc9n:b7n74p8) ([raw file](https://github.com/vaadin/framework/blob/ccac01df3ed8110b6491a0d4f7ad612901e118ef/server/src/main/java/com/vaadin/data/validator/BeanValidator.java#L121)):*
> ```Java
>     @Override
>     public ValidationResult apply(final Object value, ValueContext context) {
>         Object valueToValidate = Objects.equals(emptyValue,value)?null:value;
> ```

This could use `context.getHasValue().isEmpty()` instead of depending on the `emptyValue` provided through the new API.

---

*[server/src/main/java/com/vaadin/data/validator/BeanValidator.java, line 139 at r2](https://reviewable.io:443/reviews/vaadin/framework/10422#-L45H-XyQK2ApuQ7NDse:-L45H-XyQK2ApuQ7NDsf:bpkrmhc) ([raw file](https://github.com/vaadin/framework/blob/ccac01df3ed8110b6491a0d4f7ad612901e118ef/server/src/main/java/com/vaadin/data/validator/BeanValidator.java#L139)):*
> ```Java
>      * @param emptyValue the new empty value
>      */
>     public void setEmptyValue(Object emptyValue) {
> ```

Is this really needed? see suggestion in the `apply`method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10422#-:-L45IJYH2ooeCaB3r9FP:b-3v8jes)*

TatuLund:
I am closing this PR in favor of https://github.com/vaadin/framework/pull/11758

The new fix is made same way the issue has been solved in Flow.

PR #10433, opened 2017-12-12
Add setters to Payload for JsonCodec to be able to encode

### LINKED ISSUE #10428: Problem with setPayload in DnD
Hi! ... I'm using Vaadin 8.1.7, but I think the problem remains in newer versions. I want to implement a validation using `setPayload` & `setDropCriterion` with some components, but whenever the method `setPayload` is executed, I'm getting client side errors like this:

```
Mon Dec 11 16:32:02 GMT-300 2017 com.vaadin.client.ApplicationConfiguration
SEVERE: (TypeError) : Cannot read property 'b' of undefined
com.google.gwt.core.client.JavaScriptException: (TypeError) : Cannot read property 'b' of undefined
	at Unknown.Pc(com.vaadin.DefaultWidgetSet-0.js)
	at Unknown.TTb(com.vaadin.DefaultWidgetSet-0.js)
	at Unknown.Z1b(com.vaadin.DefaultWidgetSet-0.js)
...
```

I'm attaching a small project that reproduces the issue. First let it run (ie: `mvn clean install jetty:run`) as is, without modifications, so you're going to be able to confirm that everything goes ok (try dragging some tasks). But after that, try uncommenting the commented lines that call `setPayload` method, and then run it again, and you will see those errors on the console log.
I've only tested it on an updated chrome, but probably the sames happens in other browsers.
Regards!

[payload-problem.zip](https://github.com/vaadin/framework/files/1549126/payload-problem.zip)

wbadam:
Fixes #10428

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10433)

### DIFF
--- a/shared/src/main/java/com/vaadin/shared/ui/dnd/criteria/Payload.java
+++ b/shared/src/main/java/com/vaadin/shared/ui/dnd/criteria/Payload.java
@@ -76,6 +76,16 @@ public class Payload implements Serializable {
     }
 
     /**
+     * Sets the key of this payload.
+     *
+     * @param key
+     *         key that identifies the payload
+     */
+    public void setKey(String key) {
+        this.key = key;
+    }
+
+    /**
      * Gets the value of this payload.
      *
      * @return value of this payload
@@ -85,12 +95,32 @@ public class Payload implements Serializable {
     }
 
     /**
+     * Sets the value of this payload.
+     *
+     * @param value
+     *         value of the payload
+     */
+    public void setValue(String value) {
+        this.value = value;
+    }
+
+    /**
      * Gets the value type of this payload.
      *
      * @return the type of the value of this payload
      */
     public ValueType getValueType() {
         return valueType;
+    }
+
+    /**
+     * Sets the value type of this payload.
+     *
+     * @param valueType
+     *         type of the payload value
+     */
+    public void setValueType(ValueType valueType) {
+        this.valueType = valueType;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


elmot:
Reviewed 2 of 2 files at r1.

---

*[shared/src/main/java/com/vaadin/shared/ui/dnd/criteria/Payload.java, line 132 at r1](https://reviewable.io:443/reviews/vaadin/framework/10433#-L4AbHrHdSkhTr4J-4o8:-L4AbHrHdSkhTr4J-4o9:b-u2st2b) ([raw file](https://github.com/vaadin/framework/blob/dc2f8f04f236686e335c8a3f440db5697946ea0f/shared/src/main/java/com/vaadin/shared/ui/dnd/criteria/Payload.java#L132)):*
> ```Java
>      * @return the string representation of this payload
>      */
>     public String getPayloadString() {
> ```

Just curious: don't we need setter any kind of setter here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10433)*

wbadam:

---

*[shared/src/main/java/com/vaadin/shared/ui/dnd/criteria/Payload.java, line 132 at r1](https://reviewable.io:443/reviews/vaadin/framework/10433#-L4AbHrHdSkhTr4J-4o8:-L4AwbKU9EkrxTk3uWN-:b-k3c7op) ([raw file](https://github.com/vaadin/framework/blob/dc2f8f04f236686e335c8a3f440db5697946ea0f/shared/src/main/java/com/vaadin/shared/ui/dnd/criteria/Payload.java#L132)):*

Just curious: don't we need setter any kind of setter here?
</blockquote></details>

No because the payload is always constructed using the key/value/valueType triplet and is only used for transferring data via the drag event on the client side. Parsing this data on drop is also handled by this class.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10433)*

PR #10445, opened 2017-12-15
Selection handling fix when adding new item to ComboBox.

### LINKED ISSUE #10284: ComboBox new item and selection handling are inconsistent
#9071 fixed some of the issues when typing a new item to a ComboBox that allows such things and then exiting the field via click instead of tabulator without pressing enter, but there are some problematic cases left, and some are broken with both approaches.

Depending on the order of things you do (e.g. trying to add a new item before or after manual selection of an existing value, re-trying after failure etc.), entering text and moving out of the field without pressing enter can lead to any of these cases:
* new item isn't added, old item is restored to text field
* new item is added, old item is restored to text field
* new item is added and stays in the text field but isn't selected
* new item is added and selected (only in 8.1.6 when using tabulator)

Tested in 8.1.0, 8.1.6, 8.2.0.alpha2, and current master branch using the [test](https://github.com/vaadin/framework/files/1431392/Issue9774UI.txt) for #9774

Ansku:
Fixes #10284

- if new item is added, it should be selected
- if new item isn't added, previous selection should be restored
- it should be possible to prevent default selection of new items

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10445)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VComboBox.java
+++ b/client/src/main/java/com/vaadin/client/ui/VComboBox.java
@@ -1040,6 +1040,8 @@ public class VComboBox extends Composite implements Field, KeyDownHandler,
                     }
                 });
 
+        private String handledNewItem = null;
+
         /**
          * Default constructor
          */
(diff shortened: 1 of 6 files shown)


sirbris:
When can we expect this fix to be merged? Thanks in advance!

tsuoanttila:
Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10445)*

elmot:
Quite hackish but :lgtm: 

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10445#-:-L45Slpf9Ep1tHK8qOoB:bsesgyr)*

PR #10447, opened 2017-12-18
added aria-modal attribute and role='dialog' for modal Window #10424

### LINKED ISSUE #10424: Modal Window doesn't disable background for screen readers
- Tested with Windows 10 + Chrome 62 + NVDA 2017.4 screen reader
- Reproducable in Sampler: https://demo.vaadin.com/sampler/#ui/structure/window
- Open the Window and check the Modal checkbox
- Move mouse over some component behind the modality curtain
- Outcome: components behind the modality curtain are dictated
- Expected: components behind the modality curtain are ignored, possibly with the `aria-disabled` flag set in the root UI element
See https://www.w3.org/WAI/PF/aria/states_and_properties#aria-disabled:
> Indicates that the element is perceivable but disabled, so it is not editable or otherwise operable. See related aria-hidden and aria-readonly.

> For example, irrelevant options in a radio group may be disabled. Disabled elements might not receive focus from the tab order. For some disabled elements, applications might choose not to support navigation to descendants. In addition to setting the aria-disabled attribute, authors SHOULD change the appearance (grayed out, etc.) to indicate that the item has been disabled.

> The state of being disabled applies to the current element and all focusable descendant elements of the element on which the aria-disabled attribute is applied.

Consider also using the [aria-modal](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) property:

> aria-modal=true | div | Tells assistive technologies that the windows underneath the current dialog are not available for interaction (inert).

OlliTietavainenVaadin:
Fixes #10424 

This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10447)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VWindow.java
+++ b/client/src/main/java/com/vaadin/client/ui/VWindow.java
@@ -444,7 +444,7 @@ public class VWindow extends VOverlay implements ShortcutActionHandlerOwner,
                     && nativeEvent.getKeyCode() == KeyCodes.KEY_TAB
                     && nativeEvent.getShiftKey()) {
                 nativeEvent.preventDefault();
-                FocusUtil.focusOnLastFocusableElement(this.getElement());
+                FocusUtil.focusOnLastFocusableElement(getElement());
             }
             if (nativeEvent.getEventTarget().cast() == topTabStop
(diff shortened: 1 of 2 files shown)


tsuoanttila:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10447)*

PR #10470, opened 2017-12-25
Change the order of inheritance so that user.agent property is preserved

### LINKED ISSUE #10469: Widgetset is compiled only for Chrome by default
In Vaadin 8, the widgetset is only compiled for Chrome by default. To compile for e.g. Firefox, one needs to explicitly set `<set-property name="user.agent" value="gecko1_8,safari" />` in the gwt.xml.

`Vaadin.gwt.xml`'s `<set-property name="user.agent" value="gecko1_8,safari" />` setting is ignored and is overwritten by `Elemental.gwt.xml`'s `<set-property name="user.agent" value="safari" />` setting. It is probably due to the order of inheritance in `DefaultWidgetSet.gwt.xml`.

It is a regression in Vaadin 8. Vaadin 7 didn't use the Elemental widgetset and so the setting was not overwritten.

To test, try the `RichTextAreas` test UI with Firefox to see that rich text area is not editable whereas it is on Chrome.

wbadam:
Fixes #10469

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10470)

### DIFF
--- a/client/src/main/resources/com/vaadin/DefaultWidgetSet.gwt.xml
+++ b/client/src/main/resources/com/vaadin/DefaultWidgetSet.gwt.xml
@@ -8,9 +8,9 @@
     <!-- Hint for WidgetSetBuilder not to automatically update the file -->
     <!-- WS Compiler: manually edited -->
 
-    <inherits name="com.vaadin.Vaadin" />
-
     <inherits name="elemental.Elemental" />
+
+    <inherits name="com.vaadin.Vaadin" />
 
     <inherits name="com.google.gwt.precompress.Precompress" />
 


wbadam:

---

*[client/src/main/resources/com/vaadin/DefaultWidgetSet.gwt.xml, line 37 at r1](https://reviewable.io:443/reviews/vaadin/framework/10470#-L1CW3svvy6s-oAXLR2J:-L1CW3svvy6s-oAXLR2K:bixapgg) ([raw file](https://github.com/vaadin/framework/blob/4da155e2ebcf1b368b74f2c92ff0efd7260e21b0/client/src/main/resources/com/vaadin/DefaultWidgetSet.gwt.xml#L37)):*
> ```XML
>         <when-type-is class="com.vaadin.client.widgets.Overlay" />
>     </replace-with>
> ```

Another option could be to add the line `<set-property name="user.agent" value="gecko1_8,safari" />` near here. Not sure which is nicer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10470)*

tsuoanttila:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10470#-:-L1MiSt0xU5NU1I3VTSf:bnfp4nl)*

PR #10471, opened 2017-12-25
Remove objects from ActiveDataHandler that are no longer available in used DataProvider

tomasblahut:
After calling _DataProvider#refreshAll()_, objects, that were previously kept as active by _DataCommunicator.ActiveDataHandler_ (hereafter refered to as "_ActiveDataHandler_"), but are no longer supplied by current _DataProvider_, are not dropped from _ActiveDataHandler_ and stay active. 

Consider a use case where _Grid_ is displaying data from a _ListDataProvider_ and some of its columns use _Components_ in order to render cells content. Then replacing contents of _ListDataProvider#backend_ collection (with objects, for which _DataProvider#getId(object)_ yields a different id than for those previous ones) and calling _DataProvider#refreshAll()_ afterwards results in increasing attached _Component_ count and client side rendering performance drop.

This change causes all active objects inside _ActiveDataHandler_ to be marked as dropped while _DataCommunicator_ is sending data to client with _reset_ field set to `true`. Existing logic for dropping active objects inside _ActiveDataHandler_ prevents objects, requested by client side after reset has occured, from being dropped, even if they are marked that way. Dropping "old no longer available" objects from _ActiveDataHandler_, enables _DataCommunicator_ to destroy them and detach any _Components_ created by e.g. _Grid.Column_ _ValueProviders_.

Bellow is a simple _UI_ describing a way in which this issue can be reproduced. Increasing _Component_ count (_Referenced paintables_ variable using debug window) can be observed with every press of a button.

```java

    private static final int GRID_ROW_COUNT = 20;
    
    private final List<GridRowPojo> gridRowObjects = new ArrayList<>();

    @Override
        refreshGridRowObjects();
        setContent(prepareLayout());
    }

    private void refreshGridRowObjects() {
        gridRowObjects.clear();

        for (int index = 0; index < GRID_ROW_COUNT; index++) {
            gridRowObjects.add(new GridRowPojo(UUID.randomUUID().toString()));
        }
    }

    private VerticalLayout prepareLayout() {
        Grid<GridRowPojo> grid = new Grid<>();
        grid.setSizeFull();
        grid.addColumn(source -> {
            TextField textField = new TextField();
            textField.setValue(source.getName());
            return textField;
        }, new ComponentRenderer());
        grid.setDataProvider(DataProvider.ofCollection(gridRowObjects));

        Button btnRefresh = new Button("Refresh [new objects]", event -> {
            refreshGridRowObjects();
            grid.getDataProvider().refreshAll();
        });

        HorizontalLayout layBtns = new HorizontalLayout(btnRefresh);

        VerticalLayout layout = new VerticalLayout(grid, layBtns);
        layout.setSizeFull();
        layout.setExpandRatio(grid, 1.0f);
        return layout;
    }

    }

    public static class GridRowPojo {

        private final String name;

        public GridRowPojo(String name) {
            this.name = name;
        }

        public String getName() {
            return name;
        }
    }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10471)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -103,7 +103,9 @@ import com.vaadin.ui.components.grid.Editor;
 import com.vaadin.ui.components.grid.EditorImpl;
 import com.vaadin.ui.components.grid.Footer;
 import com.vaadin.ui.components.grid.FooterRow;
+import com.vaadin.ui.components.grid.GridMultiSelect;
 import com.vaadin.ui.components.grid.GridSelectionModel;
+import com.vaadin.ui.components.grid.GridSingleSelect;
 import com.vaadin.ui.components.grid.Header;
 import com.vaadin.ui.components.grid.Header.Row;
 import com.vaadin.ui.components.grid.HeaderCell;
@@ -3904,14 +3906,8 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
      * @throws IllegalStateException
      *             if not using a single selection model
      */
-    public SingleSelect<T> asSingleSelect() {
-        GridSelectionModel<T> model = getSelectionModel();
-        if (!(model instanceof SingleSelectionModel)) {
-            throw new IllegalStateException(
-                    "Grid is not in single select mode, it needs to be explicitly set to such with setSelectionModel(SingleSelectionModel) before being able to use single selection features.");
-        }
-
-        return ((SingleSelectionModel<T>) model).asSingleSelect();
+    public GridSingleSelect<T> asSingleSelect() {
+        return new GridSingleSelect<>(this);
     }
 
     public Editor<T> getEditor() {
@@ -3928,13 +3924,8 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
      * @throws IllegalStateException
      *             if not using a multiselection model
      */
-    public MultiSelect<T> asMultiSelect() {
-        GridSelectionModel<T> model = getSelectionModel();
-        if (!(model instanceof MultiSelectionModel)) {
-            throw new IllegalStateException(
-                    "Grid is not in multiselect mode, it needs to be explicitly set to such with setSelectionModel(MultiSelectionModel) before being able to use multiselection features.");
-        }
-        return ((MultiSelectionModel<T>) model).asMultiSelect();
+    public GridMultiSelect<T> asMultiSelect() {
+        return new GridMultiSelect<>(this);
     }
 
     /**
--- a/server/src/main/java/com/vaadin/ui/components/grid/GridMultiSelect.java
+++ b/server/src/main/java/com/vaadin/ui/components/grid/GridMultiSelect.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright 2000-2018 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.ui.components.grid;
+
+import java.util.Optional;
+import java.util.Set;
+
+import com.vaadin.event.selection.MultiSelectionListener;
+import com.vaadin.event.selection.SelectionListener;
+import com.vaadin.shared.Registration;
+import com.vaadin.ui.Grid;
+import com.vaadin.ui.MultiSelect;
+import com.vaadin.ui.components.grid.MultiSelectionModel.SelectAllCheckBoxVisibility;
+
+/**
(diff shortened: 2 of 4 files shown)


tsuoanttila:
Reviewed 2 of 2 files at r1.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 149 at r1](https://reviewable.io:443/reviews/vaadin/framework/10471#-L1MnDcCSPmqu86ftTAn:-L1MnDcCSPmqu86ftTAo:b-52hbm) ([raw file](https://github.com/vaadin/framework/blob/24cfebbec642c6560b8adc9c1bb08b50c603f7fa/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L149)):*
> ```Java
>             activeData.forEach(this::dropActiveData);
>         }
>         
> ```

extra whitespace 

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 171 at r1](https://reviewable.io:443/reviews/vaadin/framework/10471#-L1MmNhh3kao2PP4TNe_:-L1MmNhidmr0dk9v-goI:b-ufhh7r) ([raw file](https://github.com/vaadin/framework/blob/24cfebbec642c6560b8adc9c1bb08b50c603f7fa/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L171)):*
> ```Java
>          *         {@code false} if not
>          */
>         public boolean isDropped(T dataObject) {
> ```

Where is this method called? Why does it exist? If it's here only for testing purposes, please don't make it public. Calling this method with an object that does not exist will create a new key in the keymapper that will not be removed by the bookkeeping logic in the DataCommunicator. This is practically a method to make a memory leak and should not be needed by an application developer.

Missing a `null` check. Would prefer to add at least `getKeyMapper().has(dataObject)` check before calling the `key` method.

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 174 at r1](https://reviewable.io:443/reviews/vaadin/framework/10471#-L1MnRXpepMBHDdMlIpN:-L1MnRXpepMBHDdMlIpO:b-g1kqjy) ([raw file](https://github.com/vaadin/framework/blob/24cfebbec642c6560b8adc9c1bb08b50c603f7fa/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L174)):*
> ```Java
>             return droppedData.contains(getKeyMapper().key(dataObject));
>         }
>         
> ```

extra whitespace

---

*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 351 at r1](https://reviewable.io:443/reviews/vaadin/framework/10471#-L1MncrtKpiEAHWa4fad:-L1MncrtKpiEAHWa4fae:b-g1kqjy) ([raw file](https://github.com/vaadin/framework/blob/24cfebbec642c6560b8adc9c1bb08b50c603f7fa/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L351)):*
> ```Java
>                 handler.dropAllActiveData();
>             }
>             
> ```

extra whitespace

---

*[server/src/test/java/com/vaadin/data/provider/DataCommunicatorTest.java, line 260 at r1](https://reviewable.io:443/reviews/vaadin/framework/10471#-L1MnUXv2yYtw8ALI2fj:-L1MnUXv2yYtw8ALI2fk:b-g1kqjy) ([raw file](https://github.com/vaadin/framework/blob/24cfebbec642c6560b8adc9c1bb08b50c603f7fa/server/src/test/java/com/vaadin/data/provider/DataCommunicatorTest.java#L260)):*
> ```Java
>                 handler.getActiveData().containsKey(TEST_OBJECT_TWO));
>     }
>     
> ```

extra whitespace

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10471)*

tomasblahut:
*[server/src/main/java/com/vaadin/data/provider/DataCommunicator.java, line 171 at r1](https://reviewable.io:443/reviews/vaadin/framework/10471#-L1MmNhh3kao2PP4TNe_:-L1NDLMwqn_ZSUvs4Z-O:b-yu6ufh) ([raw file](https://github.com/vaadin/framework/blob/24cfebbec642c6560b8adc9c1bb08b50c603f7fa/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java#L171)):*

Where is this method called? Why does it exist? If it's here only for testing purposes, please don't make it public. Calling this method with an object that does not exist will create a new key in the keymapper that will not be removed by the bookkeeping logic in the DataCommunicator. This is practically a method to make a memory leak and should not be needed by an application developer.

Missing a `null` check. Would prefer to add at least `getKeyMapper().has(dataObject)` check before calling the `key` method.
</blockquote></details>

This method was used only for testing purposes. Thank you for pointing out the issues with it. I've replaced it for <code>getDroppedData()</code>, which should resolve all of them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10471)*

tsuoanttila:
I'll need to take another go-through of this change a bit later, but starts to look ok for me. Could you sign the CLA so we can merge it if no further issues rise up?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10471#-:-L44dg7CtCPNIapG_Bk8:b-4bu9zg)*

tomasblahut:
I have signed the CLA.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10471#-:-L47RbCx5q_AeJL_AhV4:bn5ifpp)*

elmot:
oops... the PR was closed automatically as a result of rebasing

elmot:
Closed in favor of #11167

PR #10477, opened 2017-12-28
Treat fields as readonly when bound with null setter

### LINKED ISSUE #10476: Bindings without setter should initialize field as read only
When binding a field with a null setter or to a property with no 
accessible setter the bounded field is not marked as readonly 
even if it is implicitly readonly from Binder point of view.

Currently a developer must set such fields as readonly manually
after binding; this could be done automatically by `BindingBuilder.bind`
invoking `HasValue.setReadOnly(true)` for null setter binding.

This was also discussed in #10252; here's a quote by @Legioth regarding this argument

> A potential follow-up, which should probably be implemented separately, would be that field.setReadOnly(true) would be automatically called when creating a binding with a null or no-op setter. In most cases, existing application code is probably already doing setReadOnly(true) for such fields now. There is however also a small risk that the field component is changed to not emit any new values in some other way, and automatically calling setReadOnly(true) might cause some unintended side effects.

mcollovati:
Fields bound with a null setter or to a property
with no accessible setter will be marked as readonly
Fixes #10476

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10477)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -228,6 +228,10 @@ public class Binder<BEAN> implements Serializable {
          * binder.forField(nameField).bind(Person::getName, Person::setName);
          * </pre>
          *
+         * <p>
+         * <strong>Note:</strong> when a {@code null} setter is given the field will be
+         * marked as readonly by invoking ({@link HasValue::setReadOnly}.
+         *
          * @param getter
          *            the function to get the value of the property to the
          *            field, not null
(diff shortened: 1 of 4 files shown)


mcollovati:
`DateTextHandlingTest.testSpecialValue` fails because field is bound with a null setter in `DateTextHandling`, so the field is readonly and does not accept inputs.

```java
binder.forField(dateField).withStatusLabel(errorLabel).bind(o -> dateField.getValue(), null);
```
May I replace the null setter with a no-op one in `DateTextHandling`?
```java
binder.forField(dateField).withStatusLabel(errorLabel).bind(o -> dateField.getValue(), (aVoid, date) -> {});
```

tsuoanttila:
Replace with no-op setter in `DataTextHandling`. Other than that :lgtm:

---

Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10477#-:-L1WM2YP4K5pV20Rh7my:b5w515u)*

tsuoanttila:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10477)*

PR #10491, opened 2018-01-03
Documentation: Change example to use imported class

anasmi:
In the  example in chapter "Client side application entry point", change
the used class from Button to VButton, as VButton is imported.
Otherwise, the import should be changed to google one's.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10491)

### DIFF
--- a/documentation/clientsideapp/clientsideapp-entrypoint.asciidoc
+++ b/documentation/clientsideapp/clientsideapp-entrypoint.asciidoc
@@ -26,7 +26,7 @@ public class MyEntryPoint implements EntryPoint {
     @Override
     public void onModuleLoad() {
         // Create a button widget
-        Button button = new Button();
+        VButton button = new VButton();
         button.setText("Click me!");
         button.addClickHandler(new ClickHandler() {
             @Override


tsuoanttila:
Reviewed 1 of 1 files at r1.

---

*[documentation/clientsideapp/clientsideapp-entrypoint.asciidoc, line 62 at r1](https://reviewable.io:443/reviews/vaadin/framework/10491#-L1w2R0hYwp1NIJ5EvDZ:-L1w2R0hYwp1NIJ5EvD_:beqg2bf) ([raw file](https://github.com/vaadin/framework/blob/b7c5bf4b86b744425681502aaa3215b978d32e06/documentation/clientsideapp/clientsideapp-entrypoint.asciidoc#L62)):*
> <inherits name="com.vaadin.Vaadin" />
I think the com.vaadin.Vaadin module was removed in Vaadin 8, so this only applies to Vaadin 7. Might be we need to remove documentation here and update the one in V7?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10491)*

anasmi:
*[documentation/clientsideapp/clientsideapp-entrypoint.asciidoc, line 62 at r1](https://reviewable.io:443/reviews/vaadin/framework/10491#-L1w2R0hYwp1NIJ5EvDZ:-L1wKSOhk_9cFLDDC6vq:b-wb7eyd) ([raw file](https://github.com/vaadin/framework/blob/b7c5bf4b86b744425681502aaa3215b978d32e06/documentation/clientsideapp/clientsideapp-entrypoint.asciidoc#L62)):*

> <inherits name="com.vaadin.Vaadin" />
I think the com.vaadin.Vaadin module was removed in Vaadin 8, so this only applies to Vaadin 7. Might be we need to remove documentation here and update the one in V7?
</blockquote></details>

Hi, Teemu , and thanks for the quick reply!
So does it mean that I would better make this change to Vaadin7 documentation?
And yet another question, if this module is removed, what should I inherit in case if I want a pure client side application?  As our documentation states that "If you are developing a pure client-side application, you should instead inherit com.vaadin.Vaadin".  Is " com.google.gwt.user.User"  the only one right module now?
(The biggest part of "client-side development" and "client-side applications" articles contains those misleadings for V8 as they are taken from V7 )
Thanks in advance for your answers!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10491)*

tsuoanttila:

---

*[documentation/clientsideapp/clientsideapp-entrypoint.asciidoc, line 62 at r1](https://reviewable.io:443/reviews/vaadin/framework/10491#-L1w2R0hYwp1NIJ5EvDZ:-L2-9y0X9lT3ApEw231y:b-ftrn2) ([raw file](https://github.com/vaadin/framework/blob/b7c5bf4b86b744425681502aaa3215b978d32e06/documentation/clientsideapp/clientsideapp-entrypoint.asciidoc#L62)):*

Hi, Teemu , and thanks for the quick reply!
So does it mean that I would better make this change to Vaadin7 documentation?
And yet another question, if this module is removed, what should I inherit in case if I want a pure client side application?  As our documentation states that "If you are developing a pure client-side application, you should instead inherit com.vaadin.Vaadin".  Is " com.google.gwt.user.User"  the only one right module now?
(The biggest part of "client-side development" and "client-side applications" articles contains those misleadings for V8 as they are taken from V7 )
Thanks in advance for your answers!
</blockquote></details>

In Vaadin 8, we don't anymore provide the custom widgets package for use with the client-side modules. There were quite a lot of changes to how it should've worked and we deemed it not worth to keep it up to date. So you can try to use the com.vaadin.DefaultWidgetSet or the Vaadin7Compatibility one but chances are that quite a few things will not work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10491)*

Issue #9765, opened 2017-08-01
DateField.setRangeStart can set the range start to a day earlier based on the system time zone


peergynt:
Can reproduce with Vaadin 8.1.0

Steps to reproduce:

- set the system time zone to UTC-5 (US Eastern Time)
- set localDate to LocalDate.of(2017,8,1)
- call DateField.setRangeStart with the localDate
- the start date will actually show as Jul-31 instead of Aug-1

I think that the issue happens in AbstractLocalDateField.convertToDate:

`Date.from(date.atStartOfDay(ZoneOffset.UTC).toInstant());`

Using ZoneOffset.UTC doesn't seem right. 
ZoneId.systemDefault() would probably work.

Another idea would be to use the deprecated Date constructors directly:

`Date date = new Date(localDate.getYear() - 1900, localDate.getMonthValue() -1, localDate.getDayOfMonth());`

since we do not care about time zones (this is actually the implementation from java.sql.Date.valueOf(LocalDate date)).

Most likely, this issue also affects convertFromDate as well as the implementation in AbstractLocalDateTimeField.

peergynt:
Based on [Stephen Colebourne's post on Stack Overflow](https://stackoverflow.com/questions/21242110/convert-java-util-date-to-java-time-localdate), the following conversion seems to do the trick:

```
    @Override
    protected LocalDate convertFromDate(Date date) {
        if (date == null) {
            return null;
        }
        return date.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
    }

    @Override
    protected Date convertToDate(LocalDate date) {
        if (date == null) {
            return null;
        }
        return Date.from(date.atStartOfDay(ZoneId.systemDefault()).toInstant());
    }
```

rayohagan:
See also the conversation on https://vaadin.com/pro/support#2488 which I think is quite relevant. The ZoneId doesn't work for most calculations as you have to take into account if daylight savings were in place at the date specified on the server and also on the client.

peergynt:
@rayohagan I don't have a pro account so I cannot view the conversation.
I do not quite understand the reason you say ZoneId does not work. As far as I know, the server serializes and sends the client a LocalDate which is not tied to a timezone.
In other words, the server sends something like: Feb-27-2018.
Since convertFromDate/convertToDate run on the server, using ZoneId.systemDefault() should be used.

rayohagan:
[...]

PR #10572, opened 2018-01-25
Date conversion should use systemDefault zone, not UTC

peergynt:
Using UTC ZoneOffset when doing time conversion is not correct.
See issue #9765

Note that I do not have TestBench and this issue probably needs a UI test.
Also, reproducing this issue depends on the time zone of the server running this code.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10572)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/AbstractLocalDateField.java
+++ b/server/src/main/java/com/vaadin/ui/AbstractLocalDateField.java
@@ -15,9 +15,8 @@
  */
 package com.vaadin.ui;
 
-import java.time.Instant;
 import java.time.LocalDate;
-import java.time.ZoneOffset;
+import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
 import java.util.Date;
@@ -117,8 +116,7 @@ public abstract class AbstractLocalDateField
         if (date == null) {
             return null;
         }
-        return Instant.ofEpochMilli(date.getTime()).atZone(ZoneOffset.UTC)
-                .toLocalDate();
+        return date.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


elmot:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/10572#-L45dSOUW_oNEcdZoXw7:-L45dSOUW_oNEcdZoXw8:b-8uo0ip) (no related file):*
I am afraid we can not accept the patch because it is behavior-breaking change, and it is going to break existing applications really badly.  Nevertheless, an application developer  is free to override the methods (since they are protected)  in his/her application. Another option is to introduce a field of type `ZoneId` with setter, getter, and put UTC as the default.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10572)*

elmot:

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/framework/10572#-L45dSOUW_oNEcdZoXw7:-L45nLQ6rvFYDN7crwkS:bq1wkmq) (no related file):*

I am afraid we can not accept the patch because it is behavior-breaking change, and it is going to break existing applications really badly.  Nevertheless, an application developer  is free to override the methods (since they are protected)  in his/her application. Another option is to introduce a field of type `ZoneId` with setter, getter, and put UTC as the default.
</blockquote></details>

Moreover, those converting methods are used only to exchange data between of server and client part. Yes, it is legacy and incorrect usage of date values but, nevertheless, there is no any sense  to change the one part only, and no any reason to change both parts. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10572)*

peergynt:
@elmot To address your comments:

> those converting methods are used only to exchange data between of server and client part

That was precisely my point, when the server is sending the date/time to the client, it ought to send the correct date/time.
For instance, with my server set to UTC-5, the following code:
```
DateField dateField = new DateField();
LocalDate localDate = LocalDate.of(2018,1,15);
dateField.setRangeStart(localDate);
```
will end up on the client side as 2018-1-14 being the actual start date.

datefield_start
datefield_outofrange

But if I change my server time zone to UTC (or UTC+n), then the client side will show 2018-1-15 as the start date.

> it is behavior-breaking change, and it is going to break existing applications really badly.

It seems to me that this change would have no effect for people already using UTC time (or time zones east of UTC).
Maybe you could elaborate on how this would break existing applications?

elmot:
Yes, there is a problem with date ranges, but using server time does not seem to be real fix. We have the ticket. I will we fixing it soon.

pinonpierre:
Some news about this bug?

It's a big issue for us too!

PR #10609, opened 2018-02-06
Add  missing @since tag

anasmi:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10609)


tsuoanttila:
Added the same since tag when fixed the test. this is now duplicate.

PR #10626, opened 2018-02-12
removed equals override to not require more overrides in subclasses

basisbit:
this equals method only executes what the default Object.equals does anyways, but causes code smell warnings in eclipse / sonarqube or similar static code analysis once you extend anything related to the AbstractClientConnector.
Also, we don't see how this "isThis" function would help resolving 14639 / #5655 .

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10626)

### DIFF
--- a/server/src/main/java/com/vaadin/server/AbstractClientConnector.java
+++ b/server/src/main/java/com/vaadin/server/AbstractClientConnector.java
@@ -1036,53 +1036,6 @@ public abstract class AbstractClientConnector
         this.errorHandler = errorHandler;
     }
 
-    /*
-     * (non-Javadoc)
-     *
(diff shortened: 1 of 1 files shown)


elmot:
The functionality of the code is to "reverse" the comparison "direction". And that is the way how it helps with CDI proxies. 
There should be better solution, but we can not accept the patch until that better solution is implemented.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10626#-:-L5IJEiaJS7YoEgui7CE:b-mk36pp)*

PR #10644, opened 2018-02-20
Add custom feedback value format for Slider (VSlider).

subshark:
Can you add to next version of Vaadin framework custom value for feedback text? For example slider is seconds and I want see value in format hours:minutes:sec. Now only number avalaible
```
public void setFeedbackValue(double value) {
    feedback.setText(String.valueOf(value)); //feedback is protected, can't override
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10644)


elmot:
The pull request should be from your branch to `master`, not to any of 8.X

elmot:
@subshark, please create an issue in https://github.com/vaadin/framework/issues, pull-requests are for actual code changes only.

PR #10745, opened 2018-03-23
Fix ValueChangeListener order changes with Binder

### LINKED ISSUE #9917: Vaadin 8: order in which change and binder listeners are called gets messed up
A field is bound to a bean property using Binder. The same field also has a valueChangeListener. In the code below, when the value of the combobox changes the value change listener is invoked before the setter on the Data object. It is wrong. However, if the lines of code that registers the listener and the setBean call are reversed the setter on the Data object will be called before the listener (correct behavior)

In this version of the code, the listener method ic invoked before the setValue method:
```
public class ValueChangeView extends VerticalLayout implements View {
  public class Data {
    private String value;
    public String getValue() {
      return value;
    }
    public void setValue(String value) {
      this.value = value;
    }
  }
  private Data data;
 
  @Override
  public void enter(ViewChangeEvent event) {
    data = new Data();
    Binder<Data> binder = new Binder<>(Data.class);
    NativeSelect<String> combo = new NativeSelect<>("Select");
    String items = { "One", "Two" };
    combo.setItems(items);
    addComponent(combo);
    Label comboValue = new Label();
    addComponent(comboValue);
    binder.forField(combo).bind("value");

    combo.addValueChangeListener(e -> comboValue.setValue(data.getValue()));

    binder.setBean(data);
  }
}
```
In the following version, the setValue method is called before the listener:
```
public void enter(ViewChangeEvent event) {
  data = new Data();
  Binder<Data> binder = new Binder<>(Data.class);
  NativeSelect<String> combo = new NativeSelect<>("Select");
  String items = { "One", "Two" };
  combo.setItems(items);
  addComponent(combo);
  Label comboValue = new Label();
  addComponent(comboValue);
  binder.forField(combo).bind("value");
 
  binder.setBean(data);
 
  combo.addValueChangeListener(e -> comboValue.setValue(data.getValue()));
}
```
This happens because of the following code in the Binder class called during the setBean call. The listener associated with the binder is removed from the list and then added. However, if there is another listener in the list (the added value change listener) the order is reversed and the user defined listener is called before the binder's listener. 
```
private void initFieldValue(BEAN bean) {
  assert bean != null;
  assert onValueChange != null;
  onValueChange.remove();
  try {
    getField().setValue(convertDataToFieldType(bean));
  } finally {
    onValueChange = getField() .addValueChangeListener(this::handleFieldValueChange);
  }
}
```
The order may also get messed up during other binder calls as well. The problem is the list of listeners in the EventRouter does not maintain order of methods. It gets messed up and, as a result bean values may be updated either before or after other listeners are invoked. It's inconsistent and unpredictable.
 
To see this in the attached project navigate to
[http://localhost:8080/v8-nativeselect/#!Change](url)

[v8-errors.zip](https://github.com/vaadin/framework/files/1268355/v8-errors.zip)

tsuoanttila:
Fixes #9917

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10745)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -982,8 +982,8 @@ public class Binder<BEAN> implements Serializable {
 
         private boolean readOnly;
 
-        // Not final since we temporarily remove listener while changing values
-        private Registration onValueChange;
+        private final Registration onValueChange;
+        private boolean valueInit = false;
 
         /**
          * Contains all converters and validators chained together in the
@@ -1056,7 +1056,6 @@ public class Binder<BEAN> implements Serializable {
         public void unbind() {
             if (onValueChange != null) {
                 onValueChange.remove();
-                onValueChange = null;
             }
             binder.removeBindingInternal(this);
             binder = null;
@@ -1116,12 +1115,11 @@ public class Binder<BEAN> implements Serializable {
         private void initFieldValue(BEAN bean) {
             assert bean != null;
             assert onValueChange != null;
-            onValueChange.remove();
+            valueInit = true;
             try {
                 getField().setValue(convertDataToFieldType(bean));
             } finally {
-                onValueChange = getField()
-                        .addValueChangeListener(this::handleFieldValueChange);
+                valueInit = false;
             }
         }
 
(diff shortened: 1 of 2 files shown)


elmot:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/framework/10745)*

PR #10793, opened 2018-04-05
Check if old bug is fixed

elmot:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10793)


elmot:
Does not seem working well

PR #10826, opened 2018-04-17
Feature/osgi improvements

mmerruko:
1. Enables DS injection support for UIs in OSGi
1. Fixes two issues in felix http
   1. registration through the WebServlet annotation doesn't work 
   1. fix session creation failures by using the servlet classloader

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10826)

### DIFF
--- a/osgi-integration/bnd.bnd
+++ b/osgi-integration/bnd.bnd
@@ -3,7 +3,7 @@ Bundle-Name: Vaadin OSGi Integration
 Bundle-Version: ${osgi.bundle.version}
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-License: http://www.apache.org/licenses/LICENSE-2.0
-Import-Package: com.vaadin.*;version='[${osgi.bundle.version},${osgi.bundle.version}]',\
+Import-Package: !com.vaadin.osgi.servlet*,com.vaadin.*;version='[${osgi.bundle.version},${osgi.bundle.version}]',\
     * 
 
-Export-Package: com.vaadin.osgi.servlet
+Export-Package: com.vaadin.osgi.servlet, com.vaadin.osgi.servlet.ds
--- a/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java
+++ b/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java
@@ -36,8 +36,11 @@ import org.osgi.service.log.LogService;
 import com.vaadin.osgi.resources.OsgiVaadinResources;
 import com.vaadin.osgi.resources.OsgiVaadinResources.ResourceBundleInactiveException;
 import com.vaadin.osgi.resources.VaadinResourceService;
+import com.vaadin.osgi.servlet.ds.OsgiUIProvider;
+import com.vaadin.osgi.servlet.ds.OsgiVaadinServlet;
 import com.vaadin.server.Constants;
 import com.vaadin.server.VaadinServlet;
+import com.vaadin.ui.UI;
 
 /**
  * This component tracks {@link VaadinServlet} registrations, configures them
@@ -52,7 +55,6 @@ import com.vaadin.server.VaadinServlet;
 public class VaadinServletRegistration {
     private final Map<ServiceReference<VaadinServlet>, ServiceRegistration<Servlet>> registeredServlets = Collections
             .synchronizedMap(new LinkedHashMap<>());
-
     private static final String MISSING_ANNOTATION_MESSAGE_FORMAT = "The property '%s' must be set in a '%s' without the '%s' annotation!";
     private static final String URL_PATTERNS_NOT_SET_MESSAGE_FORMAT = "The property '%s' must be set when the 'urlPatterns' attribute is not set!";
 
@@ -63,6 +65,8 @@ public class VaadinServletRegistration {
 
     private LogService logService;
 
+    private OsgiUIProvider uiProvider = new OsgiUIProvider();
+
     @Reference(cardinality = ReferenceCardinality.MULTIPLE, service = VaadinServlet.class, policy = ReferencePolicy.DYNAMIC)
     void bindVaadinServlet(VaadinServlet servlet, ServiceReference<VaadinServlet> reference)
             throws ResourceBundleInactiveException {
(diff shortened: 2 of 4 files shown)


elmot:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java, line 104 at r1](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHz_ljE-xa8Z0tHBJz:-LAHz_ljE-xa8Z0tHBK-:b-8bmgqm) ([raw file](https://github.com/vaadin/framework/blob/9a0c29f35ba4e3d3f897ef0f08938ff3c1fe9e3f/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java#L104)):*
> ```Java
> 
>     @Reference(cardinality = ReferenceCardinality.MULTIPLE, service = UI.class, policy = ReferencePolicy.DYNAMIC)
>     void bindUI(UI ui, ServiceReference<UI> reference) {
> ```

Usually we do not use default access

---

*[osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java, line 109 at r1](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHze3y5WIE_zW_Kcs4:-LAHze3y5WIE_zW_Kcs5:b-8bmgqm) ([raw file](https://github.com/vaadin/framework/blob/9a0c29f35ba4e3d3f897ef0f08938ff3c1fe9e3f/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java#L109)):*
> ```Java
>     }
> 
>     void unbindUI(UI ui) {
> ```

Usually we do not use default access

---

*[osgi-integration/src/main/java/com/vaadin/osgi/servlet/ds/OsgiUIProvider.java, line 43 at r1](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHzjZy-AoemmrwpgrA:-LAHzjZy-AoemmrwpgrB:bkmkyx8) ([raw file](https://github.com/vaadin/framework/blob/9a0c29f35ba4e3d3f897ef0f08938ff3c1fe9e3f/osgi-integration/src/main/java/com/vaadin/osgi/servlet/ds/OsgiUIProvider.java#L43)):*
> ```Java
>  * @author Vaadin Ltd.
>  *
>  * @since 8.5
> ```

Better to keep this empty

---

*[osgi-integration/src/main/java/com/vaadin/osgi/servlet/ds/OsgiVaadinServlet.java, line 42 at r1](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHzrgmA-Pt4TLSkHDi:-LAHzrgmA-Pt4TLSkHDj:bkmkyx8) ([raw file](https://github.com/vaadin/framework/blob/9a0c29f35ba4e3d3f897ef0f08938ff3c1fe9e3f/osgi-integration/src/main/java/com/vaadin/osgi/servlet/ds/OsgiVaadinServlet.java#L42)):*
> ```Java
>  * @author Vaadin Ltd.
>  *
>  * @since 8.5
> ```

Better to keep this empty

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/vaadin/framework/10826)*

mmerruko:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[a discussion](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHqyrM36utvdfdrB8N:-LAIF7J-EvbcLF0HVIX8:b-m2t3qb) (no related file):*

I have quite many concerns about the feature:
- The PR introduces an OSGi-related UIProvider. We already have one in Liferay support artifact. What is the difference between them? Are they going to clash being used together. Why do we need two?
- I'd prefer to avoid using special servlet for OSGi environments, because of possible conflicts with the original sevlet, Spring Servlet(if used) etc.
- No tests at all
- OSGiUIProvider contains a map of UI's. If a bundle with the UI is unloaded, what about resource leakages? Any tests for it?
</blockquote></details>

1. These bundles are not meant or supposed to be used together. The Liferay bundle uses a 1-1 mapping per provider and UI and creates UI instances from the ServiceObjects as long as they are in prototype scope. This on the other hand tracks all the UIs registered as services and is used as a Singleton to create UIs through ServiceObjects when needed by any servlet. 
The bundles shouldn't be used together but even if they are, they can't clash because they have different package names and even with the same package names in OSGi that's not enough to make the classes clash because they're in different bundles
1. In my opinion this kind of use is beyond the scope of this module and I can't find an alternative way to override the provider in a clean way. If there's a case for Spring Servlet + Spring OSGi I'd suppose it has needs beyond the scope of this
1. This definitely can be an issue, we can try to add some of those.
1. No tests for anything yet. The `@Reference` annotation on `VaadinServletRegistration.bindUI()` will generate an Declarative Services component which calls `unbindUI(UI ui) ` when the service is unregistered (when the bundle is unloaded AFAIK or when the service is unregistered otherwise). This handles the clean up for the map.

---

*[osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java, line 104 at r1](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHz_ljE-xa8Z0tHBJz:-LAIEv5s8l2rGfqORnZ5:b-rqr2e0) ([raw file](https://github.com/vaadin/framework/blob/9a0c29f35ba4e3d3f897ef0f08938ff3c1fe9e3f/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java#L104)):*

Usually we do not use default access
</blockquote></details>

It's typical and even recommended of OSGi DS Components to have access restriction, while this is not a component it's a delegate for one so the code it's not bloated.

---

*[osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java, line 109 at r1](https://beta.reviewable.io/reviews/vaadin/framework/10826#-LAHze3y5WIE_zW_Kcs4:-LAIF29aA-igyrQO4JFW:b-rqr2e0) ([raw file](https://github.com/vaadin/framework/blob/9a0c29f35ba4e3d3f897ef0f08938ff3c1fe9e3f/osgi-integration/src/main/java/com/vaadin/osgi/servlet/VaadinServletRegistration.java#L109)):*

Usually we do not use default access
</blockquote></details>

It's typical and even recommended of OSGi DS Components to have access restriction, while this is not a component it's a delegate for one so the code it's not bloated.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/vaadin/framework/10826)*

arotnov:
1) Maybe better to consume OsgiVaadinServlets in different methods like bindOsgiVaadinServlet(ServiceReference<OsgiVaadinServlet>...), unbindOsgiVaadinServlet(...).
2) There is no way to divide UI services between different OsgiVaadinServlets, I mean all UIs will be available across all servlets, what would be undesirable.

elmot:
Closed in favor of https://vaadin.com/directory/component/osgi-vaadin-servlet-publisher

PR #10899, opened 2018-05-08
Set no cache rules

### LINKED ISSUE #10890: The version (x) of the browser you are using is outdated and not supported
If an user with an old browser version visits our application, he or she gets the error message in the title.
That is fine, but this error page has no cacheing information, so it might be cached. Not only in the user's browser, which again would be fine, but also in proxies.
We have an nginx frontend which caches this response, so as soon as one user with an old browser visits the page, it messes up for subsequent users as well.

( For some reason ctrl-F5 does not fix the issue. We have to purge the nginx cache to get around it. Not sure if this is relevant )

We are on Vaadin 8.3.0.

anasmi:
Set no cache rules for response in UnsupportedBrowserHandler

Fix #10890

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10899)


elmot:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10899#-:-LByhd-49uOXIq2JNd16:bnfp4nl)*

Issue #10709, opened 2018-03-13
Binder needs to support enable/disable rather than just complete removal


bsutton:
Vaadin 8 Binder now supports the concept of removing a binding.

> binder.removeBinding(this.usernameField);

This is great but painful.

If you are buiding a UI which allows a user to show/hide some fields then you need to be able to dynamically add/remove bindings.

Whilst remove does removing the binding, adding the binding back in requires you to re-stipulate the full bind statement.

If we had enable /disable I could just go
> bind.disable(this.usernameField);

and then

> bind.enable(this.usernameField);

rather than:

> binder.removeBinding(this.usernameField);

and then

> binder.forField(this.usernameField)
>	.bind(AccountRequest::getUsername, AccountRequest::setUsername);

tsuoanttila:
If the field itself is in not in the UI, the Binder will update it's value when you change a bean but since the user has no way to modify that value, the value will not change (unless you do it in the application code).

So for example: 
```
TextField name = new TextField();
Binder<Person> binder = new Binder<>();
binder.bind(name, Person::getName, Person::setName);
binder.setBean(new Person("Teemu")); // This will set name TextField to have value "Teemu"

layout.removeComponent(name);
```

Above example will have a TextField that is bound in the Binder, but since the control is not available in the UI the user will not be able to change the value. Selecting which controls you give to the user does not require you to remove anything from the Binder itself.

bsutton:
Teamu,
thanks for the reponse, however its not going to achieve the desired outcome.

The first problem is that you have replaced one complexity with another.
Adding/removing components is problematic as it requires you to track the location of the insertion point for the component when you want to re-add it to the UI. In complex UIs this is not always simple.

The second issue is that its not about the user changing the value, its about not forcing the field to have valid content.

A simple scenario might be a combo box that controls whether a users is sent email notifications.
If the combo is selected then the user MUST enter an email address into a textfield.
So the email text field has an email validator and is required.
If the check box is not selected then we don't care what the email text field contains and it may indeed be blank.

Adding/removing the component won't achieve this. I need to be able to enable/disable the binding based on the combo box selection.

Brett

tsuoanttila:
[...]

PR #10912, opened 2018-05-11
Binder: conditional binding feature

marciopd:
Hello,

This pull request is about making it easier to deal with dynamic forms.

By dynamic form it's meant forms that change (fields are added or removed) depending on the user's input and/or business rules. Use cases that rely on dynamic forms are quite common in web applications.

Currently you have to explicitly deal with the complexity of binding/unbinding and possibly suffer with some side effects as bindings not getting evaluated in the initially expected order.

**Implementation strategy**

In this implementation the _Binder_ is responsible for using all the bindings or only the currently enabled ones in order to provide a service.

**Changes**

1. Enhancing the _Binding_ interface with two methods

```
        /**
         * Sets the binding as enabled depending on the value returned by
         * isEnabledSupplier.
         *
         * @param isEnabledSupplier
         *            supplies if the binding is enabled or not
         * @since
         */
        public void setEnabled(SerializableSupplier<Boolean> isEnabledSupplier);

        /**
         * Returns if the binding is enabled.
         *
         * @return {@code true} if yes; {@code false} if not
         * @since
         */
        public boolean isEnabled();
```
2. Making bindings enabled by default (backwards compatibility)
3. Changing _Binder_ services to use the currently enabled bindings when it makes sense.

**Usage Example**

In practice it looks like this:

```
        binder.forField(nameField).asRequired()
                .bind(Person::getFirstName, Person::setFirstName)
                .setEnabled(() -> nameField.isVisible());
```
In the example above, only when

>  () -> nameField.isVisible()

is true the binding is enabled. So if _nameField.isVisible()_ returns false then nameField's binding isn't validated as a required field.

_Some notes:_
- Notice that setEnabled accepts a expression that returns a boolean, what makes it decoupled from whatever is the logic responsible for enabling or disabling the binding.
- Explicit unbinding/re-binding is not needed anymore and because of that the evaluation order of the bindings doesn't change.

**Final remarks**

Please let me know if you liked the idea and implementation.
I'm open to suggestions, so please feel free to provide feedback.

Best regards!

[x] Valid tests for the pull request
[x] Contributing guidelines implemented

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)


tsuoanttila:
Quite related to Enhancement ticket #10709

---

---

*[a discussion](https://reviewable.io/reviews/vaadin/framework/10912#-LCTB7pf7j73iAEkvFjT:-LCTB7pf7j73iAEkvFjU:bk31urk) (no related file):*
I'll need to go through all the checks where things are done/not done based on the return value of the supplier. Even though the change is quite simple, it affects so many different places in the code and I'm not sure if all of them are needed or if there's a place where it should be.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 216 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT8ht76GT1wcb2A26c:-LCT8ht76GT1wcb2A26d:b-hv993e) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L216)):*
> ```Java
>          * @since
>          */
>         public void setEnabled(SerializableSupplier<Boolean> isEnabledSupplier);
> ```

`setEnabled` and `isEnabled` have mismatching types. if `setEnabled` takes `SerializableSupplier` I'd expect to have a `getEnabled` returning the same type and not a special getter for a `boolean`.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1226 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9rRH0AHo1LujBq-w:-LCT9rRH0AHo1LujBq-x:b-3pkn8n) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L1226)):*
> ```Java
>         @Override
>         public void setEnabled(SerializableSupplier<Boolean> isEnabledSupplier) {
>             if (isEnabledSupplier == null) {
> ```

Use `Objects.requireNonNull` instead of a custom if check.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1235 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCTAWME0s2kMejM7-gR:-LCTAWME0s2kMejM7-gS:bt48r5d) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L1235)):*
> ```Java
>         @Override
>         public boolean isEnabled() {
>             return Boolean.TRUE == isEnabledSupplier.get();
> ```

Why not just return the `supplier.get()`? Supplier should not return a `null` in any case (should be mentioned in the setter JavaDoc as well).

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2348 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9HP69BGaA6nXnQpn:-LCT9HP69BGaA6nXnQpo:b-zc5zr4) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2348)):*
> ```Java
>         return bindings.stream()
>                 .filter(b -> b.isEnabled())
>                 .map(b -> ((BindingImpl<BEAN, ?, ?>) b))
> ```

Is there still some requirement that needs the type to be of the specific implementation?

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2429 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9PrF4ek4f9UutviZ:-LCT9PrF4ek4f9Uutvi_:b-7uq0s8) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2429)):*
> ```Java
>      *
>      * @return set of currently enabled changed bindings
> 	 * @since
> ```

tab instead of spaces

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2447 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9XdJ9-25SMTXdBPk:-LCT9XdJ9-25SMTXdBPl:bjke92j) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2447)):*
> ```Java
>      */
>     public void setReadOnly(boolean readOnly) {
>         getBindingsEnabled().stream().filter(binding -> binding.getSetter() != null)
> ```

Why doesn't this method set the disabled bindings read only? Does it matter somehow if a binding is disabled when the users wants to set it read only?

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912#-:-LCT8UNeETSYDgGVImP9:bta9vml)*

tsuoanttila:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[a discussion](https://reviewable.io/reviews/vaadin/framework/10912#-LCTB7pf7j73iAEkvFjT:-LCTDCe11rSPgAL45eSG:bwpuorv) (no related file):*

I'll need to go through all the checks where things are done/not done based on the return value of the supplier. Even though the change is quite simple, it affects so many different places in the code and I'm not sure if all of them are needed or if there's a place where it should be.
</blockquote></details>

After taking a second look at this, I think the `Supplier` approach might be a bit dangerous one.. Let's say we write a supplier that returns `counter++ % 2 == 0`. During a data write where we hit a bean validation issue and try to revert the changes, but the binding is not enabled when we try to revert the change and actually mess up the whole bean..

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1694 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTBqLW3PVx59zr_r3d:-LCTBqLW3PVx59zr_r3e:b-n2fln1) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1694)):*
> ```Java
>             doRemoveBean(false);
>             this.bean = bean;
>             getBindingsEnabled().forEach(b -> b.initFieldValue(bean));
> ```

How should we handle the value when a binding gets enabled after the initial values have been set? Should we still write in the current "correct" value and not leave in an old one or empty?

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1735 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTC53o97Kslo3eDnq8:-LCTC53o97Kslo3eDnq9:b-z0aitt) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1735)):*
> ```Java
>         } else {
>             changedBindings.clear();
>             getBindingsEnabled()
> ```

Same question as above. Keeping in mind that the bean we read the values from will no longer be available when the field gets dynamically enabled.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1796 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTCKYz3ovQPMr1F1S9:-LCTCKYz3ovQPMr1F1SA:b-i9wbyk) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1796)):*
> ```Java
>      */
>     public boolean writeBeanIfValid(BEAN bean) {
>         return doWriteIfValid(bean, new ArrayList<>(getBindingsEnabled()))
> ```

What happens if a later field "enabledness" is checked from the edited bean and depends on a value that gets written as part of the changes? A field that is considered "enabled" when calling this method is actually "not enabled" when really writing the value. 

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1877 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTCqm0D7dLe2RIAko0:-LCTCqm0D7dLe2RIAko1:bnx3a0c) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1877)):*
> ```Java
>     protected void restoreBeanState(BEAN bean,
>             Map<Binding<BEAN, ?>, Object> oldValues) {
>         getBindingsEnabled().stream().filter(oldValues::containsKey)
> ```

Same issue as above. because of writing a temporary value into the bean we actually disabled a field that we updated earlier. Now the state for that binding is not reverted due to dynamic nature of the `supplier`. This will result in partial data updates lingering when everything should be reverted.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1904 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTD8kn0FsnhUzYmEaz:-LCTD8kn0FsnhUzYmEb-:b-t46weq) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1904)):*
> ```Java
>             Collection<Binding<BEAN, ?>> bindings) {
>         Map<Binding<BEAN, ?>, Object> oldValues = new HashMap<>();
>         getBindingsEnabled().stream().map(binding -> (BindingImpl) binding)
> ```

Same issues as above.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

tsuoanttila:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 216 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT8ht76GT1wcb2A26c:-LCXVBF9Cw0x6yp9DAPR:b-v6avdu) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L216)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

That's really the intention. 

No one should really need to access directly the supplier. The important is the result it supplies as a boolean result.

We can have different setters for enabled, including `void setEnabled(boolean)` as mentioned in a comment above, but a single getter like `isEnabled` should be enough.

</blockquote></details>

This is breaking Java conventions of naming getters and setters. 

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2447 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9XdJ9-25SMTXdBPk:-LCXY07Z1YYSnc0qjJW0:bo77epw) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2447)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

> bled bindings read only? Does it matter somehow if a binding is disabled when the users wants to set it read only?

The documentation states that the method "Sets the read only state to the given value for all current bindings.". 
But yes, I think "current bindings" for this method could be relaxed into simply all bindings.
</blockquote></details>

That's the question. Is the `current bindings` all the bindings or all currently enabled bindings. Keeping in mind that you might want to disable some field based on the read-only state of another binging  where this effectively will actually change the enabled state of a binding. Should that be taken into account as well ?

A `lambda` based `isEnabled` has infinite amount of possible side effects that should not affect how Binder does things, or alternatively we need to document it with cat sized characters that having any kind of side dynamic features in the supplier will make the binder act out in unexpected ways in many occasions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1694 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTBqLW3PVx59zr_r3d:-LCXVVue7yrpdTj3CY3h:b-5tawqd) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1694)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

>  have been set? Should we still write in the current "correct" value and not leave in an old one or empty?

`setBean` should work consistently considering its invocation time. 

After the initialisation is done, when a change happens and a binding becomes enabled (should happen inside a UI Listener) the application is in control to choose to: 
1. let the new field empty (as default); 
2. initialising the field's value directly with `setValue`; 
3. or even calling `setBean` or `readBean` again.

The developer could choose to do whatever fits better the concrete case. It's not a dead end.

What do you think? Are these options good enough?



</blockquote></details>

the problem with the Supplier is that we have 0 ways of knowing when a) the state changes or b) what kind of black magic the application developer puts there.

I do expect to seeing at least the case of enabled depending on some internal value of the bound bean and that being switched by another user in a background thread..

If it for example depends on the value of some checkbox that is part of the binder, the flow goes:
1) Application developer binds the bean, values are set
2) End user checks the checkbox
3) Something that was not enabled is now enabled, but only the application developer can know if something has changed.
4) Developers are going to be confused as things don't update automatically as they do in every other case with the Binder.

We do have `Binding#read(BEAN)` for reading the value but depends on the complexity of the application logic whether this is trivial to do or not.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1877 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTCqm0D7dLe2RIAko0:-LCXXCvS9JVPaNTUF3qr:bwcyynl) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1877)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

> binding is not reverted due to 

If the binding is not enabled then it's not the Binder responsibility to restore the bean's field.
The application should already have taken care of initialising the disabled field.

</blockquote></details>

This is related to the point of something changing in the enabled state based on the values written in the bean. When the initial write is done, there's one field that is enabled, but with the new updated data it will no longer be enabled. At the same time there's a bean validation failure that means we should revert all the changes, but since one of the bindings that we used to save a value is no longer enabled, that change is never reverted, possibly leaving the whole bean in a state where it fails the bean validation..

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

Legioth:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 216 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT8ht76GT1wcb2A26c:-LCXdtis8Dyk-XCh5O9x:bplqs60) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L216)):*

This is breaking Java conventions of naming getters and setters. 
</blockquote></details>

One option could be to split this into two separate features:
* `setEnabledSupplier(SerializableSupplier<Boolean> supplier)`
* `setEnabled(boolean enabled)` which is just a shorthand for `setEnabledSupplier(() -> enabled);`

This would keep setters and getters in sync and also offer a more direct way of disabling a binding (even though the indirect way is probably more practical in most cases).

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1005 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCXexDW5-qO5jsfj6Jd:-LCXexDW5-qO5jsfj6Je:b-dqj3gp) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L1005)):*
> ```Java
>         private boolean readOnly;
> 
>         private SerializableSupplier<Boolean> isEnabledSupplier = ENABLED_TRUE_SUPPLIER;
> ```

It wouldn't be 100% backwards compatible, but I'm wondering whether a more useful default would be to automatically disable binding for components that are either disabled or not visible?

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1235 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCTAWME0s2kMejM7-gR:-LCXcUWFBb6VkE6aQwMS:b8efa27) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L1235)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

As we can't really enforce the supplier to not return `null`, this is just to be sure `isEnabled()` can always be evaluated as a primitive boolean without any `NullPointerException` exceptions.
</blockquote></details>

There's also `java.util.function.BooleanSupplier` that doesn't have the possibility of returning `null`.  We'd just have to include our own serializable variant of it.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2431 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCXgrEJBbfQPmKI1Fuw:-LCXgrEJBbfQPmKI1Fux:bqdff9z) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2431)):*
> ```Java
> 	 * @since
>      */
>     protected Set<Binding<BEAN, ?>> getChangedBindingsEnabled() {
> ```

This method name seems really strange since it's in the format `get<Attribute>Bindings<Attribute>`. Consider renaming to e.g. `getEnabledChangedBindings()` (and then also the `getBindingsEnabled()` method for consistency).

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2447 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9XdJ9-25SMTXdBPk:-LCXcvQ_Bm1P4DhtsLSc:bir718d) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2447)):*

That's the question. Is the `current bindings` all the bindings or all currently enabled bindings. Keeping in mind that you might want to disable some field based on the read-only state of another binging  where this effectively will actually change the enabled state of a binding. Should that be taken into account as well ?

A `lambda` based `isEnabled` has infinite amount of possible side effects that should not affect how Binder does things, or alternatively we need to document it with cat sized characters that having any kind of side dynamic features in the supplier will make the binder act out in unexpected ways in many occasions.
</blockquote></details>

Should `binding` be readonly or not after this sequence of events (where the :
```
binder.setReadonly(true);
binding.setEnabled(() -> false);
binder.setReadonly(false);
binding.setEnabled(() -> true);
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

tsuoanttila:
Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1005 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCXexDW5-qO5jsfj6Jd:-LCXkBE5AUjNFTj6LFyp:bkkg1pn) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L1005)):*

It wouldn't be 100% backwards compatible, but I'm wondering whether a more useful default would be to automatically disable binding for components that are either disabled or not visible?
</blockquote></details>

We should have some convention on how to change some defaults of the Binder. Not really the main thing for this PR, something to keep in mind.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2348 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9HP69BGaA6nXnQpn:-LCXllwtEG4XCX8Dwc4r:b-ne3nr8) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2348)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

This method is replacing some calls to `getBindings()`. So their structure is quite similar. 
</blockquote></details>

This is fine by me. Still related to Leif's question about naming though.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2447 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9XdJ9-25SMTXdBPk:-LCXmVn-5v92dD58yi7F:b-6qt6xp) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2447)):*

Should `binding` be readonly or not after this sequence of events (where the :
```
binder.setReadonly(true);
binding.setEnabled(() -> false);
binder.setReadonly(false);
binding.setEnabled(() -> true);
```
</blockquote></details>

@marciopd how'd you answer Leif's question?

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1694 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTBqLW3PVx59zr_r3d:-LCXkh-C7Z6mOU5vjQm0:bbtz7a3) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1694)):*

the problem with the Supplier is that we have 0 ways of knowing when a) the state changes or b) what kind of black magic the application developer puts there.

I do expect to seeing at least the case of enabled depending on some internal value of the bound bean and that being switched by another user in a background thread..

If it for example depends on the value of some checkbox that is part of the binder, the flow goes:
1) Application developer binds the bean, values are set
2) End user checks the checkbox
3) Something that was not enabled is now enabled, but only the application developer can know if something has changed.
4) Developers are going to be confused as things don't update automatically as they do in every other case with the Binder.

We do have `Binding#read(BEAN)` for reading the value but depends on the complexity of the application logic whether this is trivial to do or not.
</blockquote></details>

I'd suggest calling the `initFieldValue` for all bindings, no matter if they're enabled or not. Just to allow later enabling the binding without having to reset the value manually.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1735 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTC53o97Kslo3eDnq8:-LCXkvtT7GW-cUiTSB14:bbtz7a3) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1735)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

The `readBean` works consistently at the time it was invoked.

If something changes later then maybe you'll need to call it again or init fields directly.
</blockquote></details>

I'd suggest calling the `initFieldValue` for all bindings, no matter if they're enabled or not. Just to allow later enabling the binding without having to reset the value manually.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1877 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTCqm0D7dLe2RIAko0:-LCXlD1066tA_CIhA_s7:bnc1w95) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1877)):*

This is related to the point of something changing in the enabled state based on the values written in the bean. When the initial write is done, there's one field that is enabled, but with the new updated data it will no longer be enabled. At the same time there's a bean validation failure that means we should revert all the changes, but since one of the bindings that we used to save a value is no longer enabled, that change is never reverted, possibly leaving the whole bean in a state where it fails the bean validation..
</blockquote></details>

I think we would need to revert all the bindings that match `oldValues`. Just to avoid the case of a binding getting disabled as part of the write.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1904 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTD8kn0FsnhUzYmEaz:-LCXlSuC94frCB274aTe:b-1r7040) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1904)):*
<details><summary><i>Previously, marciopd (Márcio P. Dantas) wrote…</i></summary><blockquote>

Maybe the same rationale mentioned for the restore could apply here too.
</blockquote></details>

This is fine as it is.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

tsuoanttila:
@marciopd will you have time to look at the proposed changes to this PR?

marciopd:
Yes, I will @tsuoanttila . I almost replied to all the comments.

I want to finish replying and review it all once more before publishing. 
You can expect it for next Monday.

marciopd:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 216 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT8ht76GT1wcb2A26c:-LC_JDZ-5O9uBHHjEr6q:brxp9sq) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L216)):*

One option could be to split this into two separate features:
* `setEnabledSupplier(SerializableSupplier<Boolean> supplier)`
* `setEnabled(boolean enabled)` which is just a shorthand for `setEnabledSupplier(() -> enabled);`

This would keep setters and getters in sync and also offer a more direct way of disabling a binding (even though the indirect way is probably more practical in most cases).
</blockquote></details>

Sounds good.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1235 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCTAWME0s2kMejM7-gR:-LC_EFi07nuSX8xTUelH:b9ztbij) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L1235)):*

There's also `java.util.function.BooleanSupplier` that doesn't have the possibility of returning `null`.  We'd just have to include our own serializable variant of it.
</blockquote></details>

Ok, sounds good. Will do.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2348 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9HP69BGaA6nXnQpn:-LE5yZ1r-9gh6NoHg2Av:b2nhll9) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2348)):*

This is fine by me. Still related to Leif's question about naming though.
</blockquote></details>

Ok.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2431 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCXgrEJBbfQPmKI1Fuw:-LC_P3cu7vgT9QK1cXHm:b9ztbij) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2431)):*

This method name seems really strange since it's in the format `get<Attribute>Bindings<Attribute>`. Consider renaming to e.g. `getEnabledChangedBindings()` (and then also the `getBindingsEnabled()` method for consistency).
</blockquote></details>

Ok, sounds good. Will do.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 2447 at r1](https://reviewable.io/reviews/vaadin/framework/10912#-LCT9XdJ9-25SMTXdBPk:-LC_Qa996ipjruqNDyKo:b-n9knjs) ([raw file](https://github.com/vaadin/framework/blob/a700b1a937d2c0b211e0c177bb0bc1ca642291fa/server/src/main/java/com/vaadin/data/Binder.java#L2447)):*

@marciopd how'd you answer Leif's question?
</blockquote></details>

According to `setReadOnly` docs having `binding` as readonly at the end of the sequence of events would be consistent with the method, as it only acts over the 'current bindings'.

If we think about how the same functionality would be implemented before conditional binding, then the disabled bindings would need to be unbound. As a consequence calling `binder.setReadonly` would never have an effect on them.

So in my opinion the best way of interpreting 'current bindings' is:

* Before conditional binding existed:  `current bindings = all bindings`.

* With conditional binding: `current bindings = enabled bindings`.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1694 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTBqLW3PVx59zr_r3d:-LC_GLTvAFGEnJp2tRQS:b-27hhin) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1694)):*

I'd suggest calling the `initFieldValue` for all bindings, no matter if they're enabled or not. Just to allow later enabling the binding without having to reset the value manually.
</blockquote></details>

When the dynamic form changes it's pretty common updating business beans and then updating the UI to reflect the changes. So initialising fields before the binding gets enabled might not be good enough.

Some ideas on this:

**Binder.setBean() is used**

In this case it should really make sense initialising the field once it's binding becomes enabled.

Maybe keeping track of the enabled bindings and initialising recently enabled ones would do the job. 
I can change the implementation to try this out.

We could use a binder value change listener to check for recently enabled bindings. 
This would only work for bound fields (and implicitly value change events), what should be stated very clearly in the documentation. 

The supplier accepts anything, but still for most use cases it might be good enough relying on value change events.

If events happening outside the Binder's reach are responsible for changing the enabled state and we want to avoid developers using workaround solutions like (hidden bound fields), then we could create a method like `Binder.updateEnabledBindings()` to allow updating and initialising the recently enabled bindings programatically.

**Initialising fields when `Binder.readBean()` is used**

According to `readBean` docs

```
    /**
...
     * The bean is not otherwise associated with this binder; in particular its
     * property values are not bound to the field value changes. To achieve
     * that, use {@link #setBean(BEAN)}.
     *
...
    public void readBean(BEAN bean) {
```

if the application is using `readBean`, it shouldn't be really expected any kind of automatic field initialisation after `readBean` invocation ends. The application is totally in control of initialisation here.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1735 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTC53o97Kslo3eDnq8:-LE5xDuNBgcEn5k72LWY:bebg4ss) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1735)):*
> When the dynamic form changes it's pretty common updating business beans and then updating the UI to reflect the changes. So initialising fields before the binding gets enabled might not be good enough.

I'm not sure initialising the field before the binding gets enabled is good enough.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1877 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTCqm0D7dLe2RIAko0:-LC_OVEr-gci0_wkYB0l:brr7umk) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1877)):*

I think we would need to revert all the bindings that match `oldValues`. Just to avoid the case of a binding getting disabled as part of the write.
</blockquote></details>

Ok. Will change and add a test case.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1904 at r6](https://reviewable.io/reviews/vaadin/framework/10912#-LCTD8kn0FsnhUzYmEaz:-LE5yPf4-qucnEbHwdU2:b2nhll9) ([raw file](https://github.com/vaadin/framework/blob/538ebea0a583f029c9c2baca4d29a76567cec82b/server/src/main/java/com/vaadin/data/Binder.java#L1904)):*

This is fine as it is.
</blockquote></details>

Ok.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

elmot:
Review status: 0 of 3 files reviewed, 9 unresolved discussions (waiting on @tsuoanttila and @marciopd)

---

*[a discussion](https://reviewable.io/reviews/vaadin/framework/10912#-LEoNj1o9tkQkeVS2WyT:-LEoNj1o9tkQkeVS2WyU:bc9ghy) (no related file):*
How badly it might be broken:
```java
    private static class Writer {
        private String firstName;
        private String lastName;
        private String middleName;
        private boolean middleNamed;
...
}

{
      TextField fNameField = new TextField("First Name");
        TextField lNameField = new TextField("Last Name");
        TextField mNameField = new TextField("Middle Name");
        CheckBox flag = new CheckBox("has m. name");

        Binder<Writer> binder = new Binder<>();
        binder.forField(fNameField).bind(Writer::getFirstName, Writer::setFirstName);
        binder.forField(lNameField).bind(Writer::getLastName, Writer::setLastName);
        binder.forField(mNameField).bind(Writer::getMiddleName, Writer::setMiddleName)
                .setEnabledSupplier(mNameField::isVisible);
        flag.addValueChangeListener(e -> mNameField.setVisible(e.getValue()));
        binder.forField(flag).bind(Writer::isMiddleNamed, Writer::setMiddleNamed);
        Writer markes = new Writer("Gabriel", "Garsia", "Markes");
        Writer murakami = new Writer("Haruki", "Murakami");
        ComboBox<Writer> comboBox = new ComboBox<>("Writer");
        comboBox.setItemCaptionGenerator(Writer::getLastName);
        comboBox.setItems(markes, murakami);
        comboBox.addValueChangeListener(e -> binder.readBean(e.getValue()));
        addComponents(comboBox, flag, fNameField, mNameField, lNameField);
}
```


---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1430 at r12](https://reviewable.io/reviews/vaadin/framework/10912#-LEi_8bQ7lC9hTvWYpAJ:-LEi_8bQ7lC9hTvWYpAK:bkoowrw) ([raw file](https://github.com/vaadin/framework/blob/8407c1a40dd5a408357d52d91c56f17827edc64a/server/src/main/java/com/vaadin/data/Binder.java#L1430)):*
> ```Java
>             ValueChangeEvent<?> event) {
>         if (!getEnabledBindings().contains(binding)) {
>             getLogger().log(Level.WARNING,
> ```

Too high  logging level. Should be DEBUG or INFO at worst.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

marciopd:
Review status: 0 of 3 files reviewed, 9 unresolved discussions (waiting on @tsuoanttila, @elmot, and @marciopd)

---

*[a discussion](https://reviewable.io/reviews/vaadin/framework/10912#-LEoNj1o9tkQkeVS2WyT:-LEpOWvg9kOWuNf7TD7k:btiaaet) (no related file):*

How badly it might be broken:
```java
    private static class Writer {
        private String firstName;
        private String lastName;
        private String middleName;
        private boolean middleNamed;
...
}

{
      TextField fNameField = new TextField("First Name");
        TextField lNameField = new TextField("Last Name");
        TextField mNameField = new TextField("Middle Name");
        CheckBox flag = new CheckBox("has m. name");

        Binder<Writer> binder = new Binder<>();
        binder.forField(fNameField).bind(Writer::getFirstName, Writer::setFirstName);
        binder.forField(lNameField).bind(Writer::getLastName, Writer::setLastName);
        binder.forField(mNameField).bind(Writer::getMiddleName, Writer::setMiddleName)
                .setEnabledSupplier(mNameField::isVisible);
        flag.addValueChangeListener(e -> mNameField.setVisible(e.getValue()));
        binder.forField(flag).bind(Writer::isMiddleNamed, Writer::setMiddleNamed);
        Writer markes = new Writer("Gabriel", "Garsia", "Markes");
        Writer murakami = new Writer("Haruki", "Murakami");
        ComboBox<Writer> comboBox = new ComboBox<>("Writer");
        comboBox.setItemCaptionGenerator(Writer::getLastName);
        comboBox.setItems(markes, murakami);
        comboBox.addValueChangeListener(e -> binder.readBean(e.getValue()));
        addComponents(comboBox, flag, fNameField, mNameField, lNameField);
}
```


</blockquote></details>

@elmot we should definitely write some UI tests for this. Can you help me setting it up?

Nowadays you can achieve the same misbehaviour but writing more code and having to deal with the complexity yourself (binding/unbinding and bindings being evaluated in different order).

In the example above it should be much easier changing the supplier condition than having to work as mentioned.

---

*[server/src/main/java/com/vaadin/data/Binder.java, line 1430 at r12](https://reviewable.io/reviews/vaadin/framework/10912#-LEi_8bQ7lC9hTvWYpAJ:-LEpXfFh7YyIRqIKCnlf:b2nhll9) ([raw file](https://github.com/vaadin/framework/blob/8407c1a40dd5a408357d52d91c56f17827edc64a/server/src/main/java/com/vaadin/data/Binder.java#L1430)):*

Too high  logging level. Should be DEBUG or INFO at worst.
</blockquote></details>

Ok.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10912)*

marciopd:
In a discussion with @elmot we agreed to try out this feature changing the approach and restricting it only to validations.

This will be addressed in another PR.

PR #10962, opened 2018-06-04
Restructure reference screenshot folder

tsuoanttila:
This patch moves screenshots into folders based on the browser name.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10962)


tsuoanttila:
This PR is being replaced with split versions of it. #10963

PR #10981, opened 2018-06-13
Graceful servlet shutdown if no VaadinService is initialized

### LINKED ISSUE #10980: NullPointerException in VaadinServlet.destroy
- Vaadin Framework version: 8.4.1
- Browser version: not browser related
- Web container name and version: Payara Embedded 5 / Arquillian

We are using Arquillian and Payara Embedded to execute JUnit tests that have a CDI dependency. We noticed that during shutdown of this container, an NPE was thrown by VaadinServlet.destroy(). This is peculiar, because the Vaadin UI (that is part of the same project as the JUnit tests) is never used/called by any of the JUnit tests.

And this seems to be the source of the problem, too. The Vaadin Servlet is not automatically initialized during container startup, and none of the JUnit tests is ever calling/starting the UI. But nonetheless the container seems to be calling the destroy() method during shutdown, even though the Servlet was never initialized in the first place. And that seems to cause the NPE, because getService() returns null in that case.

It can easily be avoided by including a null check in the destroy method, like this:

```
public void destroy() {
        super.destroy();
        if (getService() != null) {
             getService().destroy();
        }
    }

```

we patched it like this locally and it avoids the exception. But it would be nice to have this integrated into the official codebase.

regards
Mario

elmot:
Fixes #10980

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10981)

### DIFF
--- a/server/src/main/java/com/vaadin/server/VaadinServlet.java
+++ b/server/src/main/java/com/vaadin/server/VaadinServlet.java
@@ -1436,7 +1436,9 @@ public class VaadinServlet extends HttpServlet implements Constants {
     @Override
     public void destroy() {
         super.destroy();
-        getService().destroy();
+        if (getService() != null) {
+            getService().destroy();
+        }
     }
 
     private static void persistCacheEntry(ScssCacheEntry cacheEntry) {


tsuoanttila:
Reviewed 1 of 1 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/10981)*

PR #11018, opened 2018-07-02
Fix expanding an added TreeGrid node

### LINKED ISSUE #9997: TreeGrid client side IndexOutOfBoundsException after expanding new node
After adding a new root and child node to the tree, and expanding the new root node, a client side exception is thrown.
For example, when adding the items to an empty tree, this exception is thrown:
```
SEVERE: Error performing server to client RPC calls
java.lang.IndexOutOfBoundsException: The given index (1) was outside of the current number of rows (0..0)
```

After this exception, the grid refreshes its entire content (dropRows, and requestRows RPC calls), and recovers (so the exception is not visible on the UI, only in console).

Vaadin version:
8.1.4

Code sample to reproduce:
```

    @Override
        TreeGrid<Item> tree = new TreeGrid<>(Item.class);
        Button button = new Button("Add new item", e -> addItem(tree));

        setContent(new VerticalLayout(tree, button));
    }

    private void addItem(TreeGrid<Item> tree) {
        Item parent = new Item("Parent");
        Item child = new Item("Child");

        TreeData<Item> treeData = tree.getTreeData();
        treeData.addItem(null, parent);
        treeData.addItem(parent, child);

        tree.getDataProvider().refreshAll();

        tree.expand(parent);
    }

    public static class Item {
        private String name;

        public Item(String name) {
            this.name = name;
        }

        public String getName() {
            return name;
        }
    }

    }
}
```

tsuoanttila:
Fixes #9997

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11018)

### DIFF
--- a/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
+++ b/server/src/main/java/com/vaadin/data/provider/DataCommunicator.java
@@ -665,7 +665,8 @@ public class DataCommunicator<T> extends AbstractExtension {
         if (isAttached()) {
             attachDataProviderListener();
         }
-        hardReset();
+        reset = true;
+        markAsDirty();
 
         return filter -> {
             if (this.dataProvider != dataProvider) {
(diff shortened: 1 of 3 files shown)


elmot:
Reviewed 3 of 3 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/framework/11018)*

PR #11086, opened 2018-07-26
Fix Grid not using getId on select all

### LINKED ISSUE #11083: com.vaadin.ui.components.grid.MultiSelectionModelImpl: SelectAll behaves strangely
Vaadin Framework version 8.4.5, com.vaadin.ui.components.grid.MultiSelectionModelImpl
scenario:
- create a grid that holds n items that are NOT supposed to implement equals/hashcode (n > 1)
- make it multi-selectable and make the select-all box visible - like this:
MultiSelectionModel selectionModel = (MultiSelectionModel) grid.setSelectionMode(Grid.SelectionMode.MULTI);
		selectionModel.setSelectAllCheckBoxVisibility(MultiSelectionModel.SelectAllCheckBoxVisibility.VISIBLE);
- override 'getId' in the dataProvider of the grid to make the items properly identifyable
- add a selection listener to the grid
- run the test app, watch the events passed to the listener
- click 'select all' -> n items are passed (correct)
- unselect one if the items: n - 1 items are passed (correct)
- click 'select all' again: (n - 1) plus n items are passed (wrong. expected: n items)

Problem:
com.vaadin.ui.components.grid.MultiSelectionModelImpl#onSelectAll (vaadin 8.4.5) line 327 adding to the allItems-Set doesn’t care about the getId-method of the dataProvider but relies on the hashcode-method of the given item

tsuoanttila:
Fixes #11083

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11086)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java
+++ b/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java
@@ -445,9 +445,15 @@ public class MultiSelectionModelImpl<T> extends AbstractSelectionModel<T>
         }
 
         doUpdateSelection(set -> {
+            DataProvider<T, ?> dataProvider = getGrid().getDataProvider();
             // order of add / remove does not matter since no duplicates
-            set.removeAll(removedItems);
-            set.addAll(addedItems);
+            Set<Object> removedItemIds = removedItems.stream()
+                    .map(dataProvider::getId).collect(Collectors.toSet());
+            set.removeIf(
+                    item -> removedItemIds.contains(dataProvider.getId(item)));
+            addedItems.stream().filter(
+                    item -> !selectionContainsId(dataProvider.getId(item)))
+                    .forEach(set::add);
 
             // refresh method is NOOP for items that are not present client side
             DataCommunicator<T> dataCommunicator = getGrid()
--- a/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java
+++ b/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java
@@ -15,29 +15,18 @@
  */
 package com.vaadin.ui.components.grid;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
-import java.util.function.Consumer;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import com.vaadin.data.provider.DataCommunicator;
-import com.vaadin.data.provider.DataProvider;
-import com.vaadin.data.provider.HierarchicalDataProvider;
-import com.vaadin.data.provider.HierarchicalQuery;
-import com.vaadin.data.provider.Query;
+import com.vaadin.data.provider.*;
 import com.vaadin.event.selection.MultiSelectionEvent;
 import com.vaadin.event.selection.MultiSelectionListener;
 import com.vaadin.shared.Registration;
 import com.vaadin.shared.data.selection.GridMultiSelectServerRpc;
 import com.vaadin.shared.ui.grid.MultiSelectionModelState;
 import com.vaadin.ui.MultiSelect;
+
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Multiselection model for grid.
@@ -429,12 +418,20 @@ public class MultiSelectionModelImpl<T> extends AbstractSelectionModel<T>
                     + " although user selection is disallowed");
         }
 
-        // if there are duplicates, some item is both added & removed, just
-        // discard that and leave things as was before
-        addedItems.removeIf(item -> removedItems.remove(item));
+        DataProvider<T, ?> dataProvider = getGrid().getDataProvider();
 
-        if (selection.containsAll(addedItems)
-                && Collections.disjoint(selection, removedItems)) {
+        addedItems.removeIf(item -> {
+            Object id = dataProvider.getId(item);
+            Optional<T> toRemove = removedItems.stream()
+                    .filter(i -> dataProvider.getId(i).equals(id)).findFirst();
+            toRemove.ifPresent(i -> removedItems.remove(i));
+            return toRemove.isPresent();
+        });
+
+        if (addedItems.stream().map(dataProvider::getId)
+                .allMatch(this::selectionContainsId)
+                && removedItems.stream().map(dataProvider::getId)
+                        .noneMatch(this::selectionContainsId)) {
             return;
         }
 
@@ -445,7 +442,6 @@ public class MultiSelectionModelImpl<T> extends AbstractSelectionModel<T>
         }
 
         doUpdateSelection(set -> {
-            DataProvider<T, ?> dataProvider = getGrid().getDataProvider();
             // order of add / remove does not matter since no duplicates
             Set<Object> removedItemIds = removedItems.stream()
                     .map(dataProvider::getId).collect(Collectors.toSet());
--- a/server/src/test/java/com/vaadin/tests/components/grid/GridMultiSelectionModelProxyItemTest.java
+++ b/server/src/test/java/com/vaadin/tests/components/grid/GridMultiSelectionModelProxyItemTest.java
@@ -1,0 +1,99 @@
+package com.vaadin.tests.components.grid;
+
+import com.liferay.portal.kernel.atom.AtomEntryContent;
+import com.vaadin.data.provider.CallbackDataProvider;
+import com.vaadin.ui.Grid;
+import com.vaadin.ui.components.grid.MultiSelectionModel;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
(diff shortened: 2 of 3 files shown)


mehdi-vaadin (changes requested):
What about doing some refactoring and change the type of `selection` to `HashMap` where the ids can be kept as key and the objects as values? It may be better in terms of performance.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086#-:-LIL-jnX5_SKr3D6gGWB:b-jlbdp0)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @tsuoanttila)

___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 436 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKe96x72GkFLcL8jnW:-LIKe96x72GkFLcL8jnX:b3yk66) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L436)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>         addedItems.removeIf(item -> removedItems.remove(item));
>  
>         if (selection.containsAll(addedItems)
>                 && Collections.disjoint(selection, removedItems)) {
>             return;
>         }
> ```
</details>

This part should also be changed.
___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 442 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKeP1A3CtBguJtBcNe:-LIKeP1A3CtBguJtBcNf:bhdnamj) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L442)):*
> ```
> removedItems.isEmpty()
> ```
Although `removedItems` is not empty, nothing may be removed from `selection`, therefore `allSelected` shouldn't be set to false. So, it may be better to set `allSelected` after removing items and make sure that at least one item is really removed.

mehdi-vaadin (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086)* status: all files reviewed, 2 unresolved discussions (waiting on @tsuoanttila)

anasmi:
I am apologize in advance, if it makes no sense, but if we want to make refactoring there, maybe it's better to make it separately in another pr?
Because as far as I understand it, Teemu's fix about another issue and doesn't touch the parts you have reviewed : )

tsuoanttila (commented):
In the default case of we'd store the element into the hashmap twice, since by default the id is the item itself (which is true when the equals and hashcode are implemented properly). Would need to investigate if it makes some of the already existing update logic more complex.

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086#-:-LIL2yOQ4nPdrIb2psgA:b-mkj1dy)* status: all files reviewed, 2 unresolved discussions (waiting on @tsuoanttila)

___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 436 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKe96x72GkFLcL8jnW:-LIL2Rbz1ueXPh6G5aW6:b-1wx9m1) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L436)):*
<details><summary><i>Previously, mehdi-vaadin (Mehdi Javan) wrote…</i></summary><blockquote>

> ```
>         addedItems.removeIf(item -> removedItems.remove(item));
>  
>         if (selection.containsAll(addedItems)
>                 && Collections.disjoint(selection, removedItems)) {
>             return;
>         }
> ```
This part should also be changed.
</blockquote></details>

This code block attempts to be a shortcut for the case where there is actually no changes at all. In the case of actually having different objects is a valid concern, as updating the selection also updates the current cache of selected items.

I'm not even sure if it ever makes sense to shortcut it at all.
___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 442 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKeP1A3CtBguJtBcNe:-LIL1stY4fSoUoc2XVPA:bgp9l68) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L442)):*
<details><summary><i>Previously, mehdi-vaadin (Mehdi Javan) wrote…</i></summary><blockquote>

> ```
> removedItems.isEmpty()
> ```
Although `removedItems` is not empty, nothing may be removed from `selection`, therefore `allSelected` shouldn't be set to false. So, it may be better to set `allSelected` after removing items and make sure that at least one item is really removed.
</blockquote></details>

This PR is not about universally solving the case of all selection updates. Yes, without digging deeper into the code I can't know if there's something maybe broken with this or maybe not. The issue of this PR is about the select all not using the `getId` to properly identify the selected items.

If there's an issue with this, it should be made into a ticket and handled in it's own PR.

mehdi-vaadin (changes requested):
You are right. it needs more investigation.

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086#-:-LIL6M4y1NXPzQzZ1iUf:b8njpnf)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @mehdi-vaadin)

___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 436 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKe96x72GkFLcL8jnW:-LIL4ldM6ova_Dd1SEtz:bfodgaa) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L436)):*

This code block attempts to be a shortcut for the case where there is actually no changes at all. In the case of actually having different objects is a valid concern, as updating the selection also updates the current cache of selected items.

I'm not even sure if it ever makes sense to shortcut it at all.
</blockquote></details>

When using `getId` is ignored here to find out whether we have any change or not, then we may end up in `return` and doing nothing.  Of course, only in some rare cases.
One more thing is that it looks confusing since two different methods are used to compare objects.
___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 442 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKeP1A3CtBguJtBcNe:-LIL5uBt1BiYzk2AVGKj:b32w0x) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L442)):*

This PR is not about universally solving the case of all selection updates. Yes, without digging deeper into the code I can't know if there's something maybe broken with this or maybe not. The issue of this PR is about the select all not using the `getId` to properly identify the selected items.

If there's an issue with this, it should be made into a ticket and handled in it's own PR.
</blockquote></details>

OK

tsuoanttila (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tsuoanttila)

___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 436 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKe96x72GkFLcL8jnW:-LIL7CgE3G9no6iFHepg:b-vvw29q) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L436)):*
<details><summary><i>Previously, mehdi-vaadin (Mehdi Javan) wrote…</i></summary><blockquote>

When using `getId` is ignored here to find out whether we have any change or not, then we may end up in `return` and doing nothing.  Of course, only in some rare cases.
One more thing is that it looks confusing since two different methods are used to compare objects.
</blockquote></details>

True. In the case of removing a not-equals item we might skip it, but not sure how we would end up in that situation. Will fix this.

tsuoanttila (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mehdi-vaadin)

___
*[server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java, line 436 at r1](https://reviewable.io/reviews/vaadin/framework/11086#-LIKe96x72GkFLcL8jnW:-LILclH01Z6oYR9pcnSa:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/69c235f7d4fb90cff54e9be05dccdb818cfc66b1/server/src/main/java/com/vaadin/ui/components/grid/MultiSelectionModelImpl.java#L436)):*

True. In the case of removing a not-equals item we might skip it, but not sure how we would end up in that situation. Will fix this.
</blockquote></details>

Done.

mehdi-vaadin (approved):
:lgtm:

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11086#-:-LILlMYP5QyFAjxs6qCw:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11138, opened 2018-08-23
Ensure resizing CssLayout triggers size check for the contents.

### LINKED ISSUE #11124: TabSheet tabpanel is not resized properly
- Vaadin Framework 8.5.1
- Seems to be browser / OS independent (tested on Windows 10 with latest Chrome, FF and IE11)

- Description of the bug
The bug only occurs with a quite specific set of layouts (see sample project). What happens is that when the browser window is resized horizontally, the `<div role="tabpanel" class="v-tabsheet-content">` element does not react to the changed width. However, when a server round trip has completed (eg. click a button that shows a notification), the tabpanel will snap back to the correct size. 

- Expected behavior
Resizing the browser window should also refresh the TabSheet's contents' size. 

Note that there are some workarounds to the issue; one of them is having a BrowserResizeListener that does `UI.getCurrent().markAsDirty();` in the handler. Another workaround is using a different root-level layout (VerticalLayout instead of CssLayout).

Sample project to reproduce the issue: 
[v8-tabsheet-resize-bug.zip](https://github.com/vaadin/framework/files/2305492/v8-tabsheet-resize-bug.zip)

Ansku:
Fixes #11124

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11138)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VCssLayout.java
+++ b/client/src/main/java/com/vaadin/client/ui/VCssLayout.java
@@ -18,8 +18,10 @@ package com.vaadin.client.ui;
 
 import com.google.gwt.user.client.ui.FlowPanel;
 import com.google.gwt.user.client.ui.Widget;
+import com.vaadin.client.LayoutManager;
 import com.vaadin.client.Profiler;
 import com.vaadin.client.StyleConstants;
+import com.vaadin.client.ui.layout.ElementResizeListener;
 
 /**
  * VCssLayout is a layout which supports configuring it's children with CSS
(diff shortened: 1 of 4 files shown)


Ansku:
These changes aren't something that can be introduced on the framework to CssLayout level as it is by design as lightweight as possible. However, they can be added through local extending or the use of an Extension to alter the behaviour on project level. The purpose of this pull request was to check that the changes wouldn't break any regression tests (which they didn't). Alternatively, adding more programmatic sizing (either fixed or relative) to your layouts can keep your application more attuned to size changes.

PR #11191, opened 2018-09-17
Display the caption of the Empty selection in NativeSelect

### LINKED ISSUE #10937: NativeSelect setEmptySelectionCaption does not work for non-String NativeSelects
*Steps to reproduce*
- Create a new `NativeSelect<Integer>`
- Set the empty selection caption to "Select"
- Add some numbers to the NativeSelect

*Expected outcome*
image
image

*Actual outcome*
image
image

So basically, the empty selection caption occurs in the list, but it's not selected by default, and can not be selected by server-side code. Using `clear()`, `setValue(null)` or `setSelectedItem(null)` sets the NativeSelect back to the empty string caption.

Most likely, this is due to the empty selection caption being shown when an empty string "" is selected, not when null is selected.

Same issue as here (permalink to second to last message):  https://vaadin.com/forum/thread/15654110/16909551

---

Framework version: 8.4.2
Browser version: Chrome 59.0.3071.86

anasmi:
Fixes #10937 
- Previously if selected value is null, then index is set to -1; in current implementation if value is null and emptySelection is allowed then set the index to 0. (The position for the empty selection)

- Also, if changing the allowEmptySelection on the fly, ensure, that either index is to-reset to -1 by setting the selected value to null on the client-side (the value before was null) or preserve the value(value was different than empty).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11191)


ZheSun88 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11191)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @anasmi)

___
*[client/src/main/java/com/vaadin/client/ui/VNativeSelect.java, line 32 at r1](https://reviewable.io/reviews/vaadin/framework/11191#-LOWy7mk5FEC6tmeznr5:-LOWy7ml4YKpEwLXC7tt:bn18e7w) ([raw file](https://github.com/vaadin/framework/blob/b1562986ba37f5a9f3a5078f45ad6ec2952d4aaa/client/src/main/java/com/vaadin/client/ui/VNativeSelect.java#L32)):*
> ```Java
> 
>     private final ListBox listBox = new ListBox();
>     private boolean emptySelectionAllowed = true;
> ```

Should we support the `emptySelectionAllowed` by default? 

according to the commit message, the index is set to `-1` by the framework, if we set `emptySelectionAllowed` to `true` by default, the behavior is going to be changed. 

am i right?

ZheSun88 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11191)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @anasmi)

___
*[client/src/main/java/com/vaadin/client/ui/VNativeSelect.java, line 32 at r1](https://reviewable.io/reviews/vaadin/framework/11191#-LOWy7mk5FEC6tmeznr5:-LOX-OqL6sUuC27Vi_5B:b-s8r9eh) ([raw file](https://github.com/vaadin/framework/blob/b1562986ba37f5a9f3a5078f45ad6ec2952d4aaa/client/src/main/java/com/vaadin/client/ui/VNativeSelect.java#L32)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

Should we support the `emptySelectionAllowed` by default? 

according to the commit message, the index is set to `-1` by the framework, if we set `emptySelectionAllowed` to `true` by default, the behavior is going to be changed. 

am i right?
</blockquote></details>

my reviewable is offline suddenly, and a piece of the message is missing from the comment. 

according to the commit message, the index was set to -1 by the framework previously. And now if we set emptySelectionAllowed to true, the index will be set to 0 instead, so in this case, the behavior is going to be changed.

am i right?

anasmi:
___
*[client/src/main/java/com/vaadin/client/ui/VNativeSelect.java, line 32 at r1](https://reviewable.io/reviews/vaadin/framework/11191#-LOWy7mk5FEC6tmeznr5:-LOX14bx8UT2SzmBMffx:bl68tk1) ([raw file](https://github.com/vaadin/framework/blob/b1562986ba37f5a9f3a5078f45ad6ec2952d4aaa/client/src/main/java/com/vaadin/client/ui/VNativeSelect.java#L32)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

my reviewable is offline suddenly, and a piece of the message is missing from the comment. 

according to the commit message, the index was set to -1 by the framework previously. And now if we set emptySelectionAllowed to true, the index will be set to 0 instead, so in this case, the behavior is going to be changed.

am i right?
</blockquote></details>

Not exactly. The EmptSelectionAllowed was true by default. The value was propagated from the state. https://github.com/vaadin/framework/blob/master/shared/src/main/java/com/vaadin/shared/ui/nativeselect/NativeSelectState.java#L40

Through, previously, when you are trying to set "null" as a value, no value was selected. Currently, it will default to the emptySelectionItem.

ZheSun88 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11191)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @ZheSun88)

___
*[client/src/main/java/com/vaadin/client/ui/VNativeSelect.java, line 32 at r1](https://reviewable.io/reviews/vaadin/framework/11191#-LOWy7mk5FEC6tmeznr5:-LOX2q_N7VeIg9KUSQ6n:blo85ox) ([raw file](https://github.com/vaadin/framework/blob/b1562986ba37f5a9f3a5078f45ad6ec2952d4aaa/client/src/main/java/com/vaadin/client/ui/VNativeSelect.java#L32)):*

Not exactly. The EmptSelectionAllowed was true by default. The value was propagated from the state. https://github.com/vaadin/framework/blob/master/shared/src/main/java/com/vaadin/shared/ui/nativeselect/NativeSelectState.java#L40

Through, previously, when you are trying to set "null" as a value, no value was selected. Currently, it will default to the emptySelectionItem. 
</blockquote></details>

aha, I see. then everything looks fine . let us rebuild the PR and get it in.

mstahv:
What's up with this PR? When will it be in a release? I have been waiting for this fix for a while now, but I guess I need to change the component now to something else.

ZheSun88 (approved):
> Reviewed 4 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11191)* status: :shipit: complete! all files reviewed, all discussions resolved

ZheSun88:
@mstahv currently the framework 8.6.0.rc1 has been built and staged, as this PR containing some APIs changes, according to the SemVer, it is not possible to be included in the 8.6.0 release now.

mstahv:
There is just some new API in the client side widget. Nobody cares about that, it isn't a real API change. This is a bugfix, put it in ASAP.

ZheSun88:
aww.. `8.6.0.rc2`, `8.7.0.alpha1`, which one do you prefer? or we just release 8.6.0 with this included? 

Ping @pleku also.

mstahv:
Next stable release. If there are no more 8.5 series releases coming, then to 8.6.

mstahv:
What is going on with this issue? This is a bug fix (regression), there is finally a fix for it, why can't we just ship it? It has been without a fix for too long, then without a review for too long and now it missed a release and is tagged only to next *feature release*.

PR #11234, opened 2018-10-09
Try to fix the java env variable for fw7

jtomass:
Testing if failing env variable can be fixed with some glue.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11234)


jtomass:
Doesn't accomplish anything, removing.

PR #11252, opened 2018-10-17
Make value changes in color picker user originated

jeremystucki:
This is not a programmatically initiated event, so it should have the
`userOriginated` property set to `true`.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11252)


jeremystucki:
I think you are right. Setting the value twice prevents the `ValueChangeEvent` from beeing fired. See [AbstractField.java](https://github.com/vaadin/framework/blob/master/server/src/main/java/com/vaadin/ui/AbstractField.java#L137)

anasmi:
Indeed, so why to make redundant check?

jeremystucki:
There should only be one call to `setValue`. I am unsure whether the call should come from where I edited it or from the rpc.

anasmi:
We will not modify the rpc call there, so I think the right way would be to modify the line you have worked with

anasmi:
I've continued investigated and then I realize that `rpc` call we are using is actually was not supposed to be there. (It's client to server RPC). So we are back to the initial situation. I have updated my pr, but first need to wait for automated tests to run to see that it doesn't break anything.
As as result, I preserve `setValue` with true for `userOriginated` parameter, then propagating value to the window object itself and updating value of the color. This doesn't contain NPE checks now, so it's definitely work in progress.

Through, we will need to add tests to either of the pull request. ( am thinking of verifying that `addValueChangeListener` returns indeed `true` ). 

That's actually turned to be more complicated issue, but thank you for bringing it up!

anasmi:
Hi,
I've modified your pr a bit and added commit to your branch, but for some reason the pr is not updated. I will also try to create test to add to this pr, but can't promise that will be able to do it today.

anasmi:
Hi,
I've modified your pr a bit and added commit to your branch, but for some reason the pr is not updated. I will also try to create test to add to this pr, but can't promise that will be able to do it today.

anasmi:
Hi,

I've modified your pr a bit and added commit to your branch, but for some reason the pr is not updated. I will also try to create test to add to this pr, but can't promise that will be able to do it today.

anasmi:
Hi @jeremystucki,

I have added test and added commit to your branch, but github hasn't updated pr yet.

Hopefully you are alright with this. I will close then my old pr prior to this one : )

jeremystucki:
Thank you for all this work. Looks good to me

jeremystucki (commented):
:lgtm:

> Reviewed 1 of 1 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11252#-:-LPUrgYD0vpRC8vYe-uy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

jeremystucki:
@anasmi
Can you tell me the current status of this PR?

anasmi:
Hi,

it should be reviewed from some of the team member. Unfortunately, I don't have rights to review/merge it, as I have taken part in it myself.
I will ask around,if someone could do it.

Thanks for pointing out! :)

ZheSun88 (approved):
> Reviewed 1 of 1 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11252)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11260, opened 2018-10-18
Avoid column width calculation after inserting the first rows on attach.

ericflock:
The calculation will be scheduled finally.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11260)


TatuLund:
When tested manually I did not find noticeable performance improvement with this. Also there is regression risk, that is difficult to assess.

PR #11306, opened 2018-11-12
Interface for creating default converters for binder API

### LINKED ISSUE #9202: Binder should automatically configure built in converters
Version: 8.0

Binder in V8 is not as handy as related tools in previous Vaadin versions, as developers will need to manually configure all converters, even for basic datatypes like numbers. It now forces developers to write boilerplate code for even simple cases.

Binder could do this automatically at least if name based binding is done using bindInstanceFields method (this is what I most often use).

mrminy:
Add ability to make default converters for any given datatypes and connect it to a Vaadin 8 binder. 

Partly solves the issue discussed here: https://stackoverflow.com/questions/43154002/does-vaadin-8-binderbindinstancefields-only-work-with-string-data-types

Also, it would remove a lot of boilerplate code for converting regular values as Integer and Double from String when using TextField. Similarly, this would make it easier to have forms using the old java.util.Date with the new Vaadin 8 Date[Time]Fields.

Fixes #9202 

This works more like a implementation suggestion. Would require some more work (tests and docs) if it were to be accepted.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11306)


anasmi:
Hi,

and thanks for submitting the pull request! 

Could you, please, run `mvn clean install` in the project? (By running default formatting will be applied)

Than we could go through the changes itself : )

ZheSun88:
Hi, 
Thanks for the contribution again. 
May I ask you to provide a small sample on the usage of this new feature? That would make it easier for the code review.  

Thanks.

ZheSun88:
Hi @mrminy , 

I saw you have closed this PR, may I know about your opinion? 

As Pekka had been working on this PR and tried to get the functionality into the framework a while ago. But according to the different priorities, he must focus on other projects. 

My apologies for not being active in this issue, again.

PR #11329, opened 2018-11-22
Fixing an NPE in MPR

mehdi-vaadin:
In MPR, there are cases when before the initialization of the `UI`, we need to access `Page.location` indirectly and since it is not initialized we get an NPE. To resolve it, the `location` is set to an empty string as its default value in the ctor of `Page`.

Related to https://github.com/vaadin/multiplatform-runtime/issues/26

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11329)


mehdi-vaadin:
To avoid any side effect in the framework, we used another approach inside MPR.

PR #11362, opened 2018-12-10
V7: Improve VMenuBar click handling logic

anasmi:
Backport to V7:

During `updateFromUIDL` inside MenuBarConnector we empty and re-instantiate the components of MenuBar. When we are modifying the Menubar from the BlurEventListener of another component, we ,by this, remove widgets, therefore clickEvent is not fired and the action of the MenuItem is not proceed as a result. (The BlurEvent is fired before the click event in the chain of events. )

To improve the situation, we catch onMouseDown event , which is fired before BlurEvent,by assigning mouseDown flag to true. Then if no click event has yet happened, we delay the execution of update inside `updateFromUIDL` by default 500 ms. Then if click event occurs, it proceeds normally. The time can be increased/decreased using setter.

There is no delay, if we are clicking on the MenuBar as usual or no Blur listener is set.

This change allows setting descriptions preserving the action from the MenuItem

(cherry picked from commit 22cc85c)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11362)


OlliTietavainenVaadin (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11362#-:-LTfyqLY43vpZxLZ7diJ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11410, opened 2019-01-07
If doCometSupport returns Action.CANCELLED HTTP 200 is not correct 

TatuLund:
This patch will change handleRequest return HTTP 400 instead of 200 in case doCometSupport returns Action.CANCELLED

See also discussion here: https://stackoverflow.com/questions/46234679/what-is-the-correct-http-status-code-for-a-cancelled-request

According to Atmosphere code https://github.com/Atmosphere/atmosphere/blob/master/modules/runtime/src/main/java/org/atmosphere/runtime/AtmosphereFramework.java#L2267 it is always failure when value Action.CANCELLED is returned.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11410)


mehdi-vaadin:
@TatuLund Can we delete the related branch and set the milestone to invalid?

TatuLund:
> TatuLund Can we delete the related branch and set the milestone to invalid?

Yes you can and should.

Issue #11354, opened 2018-12-03
Combobox dropdown stays open/showed, when you input data quickly and exit with TAB-Key (Vaadin 7 and 8)


simpletasks:
Tested on Vaadin 7.7.6 and 7.7.15
Tested on Chrome 70.0.3538.110 and  Firefox 63.0.3
Tested on Jetty 9.2.3.v20140905

When using tab for moving from one combobox to another after entering one or
two digit value combobox option list stays open.

Reproduction:
Four Combobox<Integer> components in horizontal layout. Every combobox has list of some integers.
Use tab to focus first combobox, enter one digit number value and follow with fast tab press to focus next combobox.
Almost every time option list on one to all comboboxes stays open.
Value change listeners are not called but blur listeners are (I set valueChangeListener and BlurListener). This result in losing new value entered in combobox.
Use case is better when override Combobox class  and add TextChangeListener in which I can fire valueChangeEvent.
In this case new values are saved and blur listener can use them.
 But option list for combobox still stay open.

![vaadin-bug](https://user-images.githubusercontent.com/27790652/49457710-955bd980-f7eb-11e8-81ec-cd8d98d3d0e1.png)

  I expect that even in fast typing new values in comboboxes followed with fast tab usage will show correct combobox
 behavior. On blur option list should be hidden or removed from page.
 
 Now, combobox do not save new value, that is first problem. Adding TextChangeListeners looks like help but I'm not sure  is this solution good enough for every case.
 Second problem is that combobox option list stays open when using keyboard to fast.

simpletasks:
Problem can also be seen here:
https://demo.vaadin.com/book-examples-vaadin7/book/#component.select.common.basic

Type fast: 1) enter any letter and immediately press tab. Option list stays open even empty.

anasmi:
This issue should be fixed by this pull request : https://github.com/vaadin/framework/pull/11436

PR #11436, opened 2019-01-24
Ensure pop-up is not opened, when tabbing out fast from Combobox

anasmi:
Currently tabbing out from Combobox quickly into another one, opens the popup window. 
The event chain is following: 

- Pressing any letter. `A`, for example( `KeyDownEvent` 1. CB)
- Pressing tab (`KeyDownEvent` 1.CB)
- Blur Event happens (the 2.nd CB is focused)
- `KeyUpEvent` for letter `A`
- `KeyUpEvent` for `Tab`

As `KeyUpEvent` is caught only by focused component, the popup is opened incorrectly for the currently focused component, not the one where button was originally pressed. To prevent filtering (which is called from KeyUpEvent handler) check that the `keyDown` event was called prior in the current combobox. Also, prevent showing the popup, if field is not focused anymore.

Related to https://github.com/vaadin/framework/issues/11354 

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11436)


tepi (commented):
The fix seems to fix the issue when testing manually, and looks otherwise ok as well. However the test does not seem to be ok: It passes whether I have the fix applied or not. It should fail when the fix is not applied.

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11436#-:-LYSGMW_00vXANWOiYeB:b740wqo)* status: 0 of 3 files reviewed, all discussions resolved

anasmi:
@tepi sorry for long delay. I've tried to come up with the idea, how to test it reliable, but I am not sure, we can.
- Trying to set value via JS into combobox doesn't fire filtering, therefore no pop-up is showing. 
- `sendKeys` doesn't imitate the situation either.(I think because it's too slow, but I am not sure)
- The screenshot comparison will not have any effect either, as I can get broken screen automatically. 
Therefore, would you pass the pull request if I remove the test all in all? Or should I still try to figure out a way to test is reliably?

tepi:
Yeah, looks like this is one of those cases you can't really test reliably or at all. I am fine with removing the test, makes no sense to spend too much time on it.

tepi (commented):
> Reviewed 1 of 2 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11436)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #10259, opened 2017-10-26
Issue3922disableditem

Ansku:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/10259)

PR #11442, opened 2019-01-30
Add lost containerElement getter to VMenuBar

TobseF:
Adds access to the `VMenuBar#containerElement` which was removed in the `VMenuBar` restructuring 
of #10259.  There the `getContainerElement()` was removed because it wasn't mandatory anymore from the parent `SimplePanel`.
But as you can see on the users of the field `containerElement` it's still very handy if you want to add an new element at the first position.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11442)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VMenuBar.java
+++ b/client/src/main/java/com/vaadin/client/ui/VMenuBar.java
@@ -316,6 +316,15 @@ public class VMenuBar extends FocusableFlowPanel implements
     }
 
     /**
+     * Returns the containing element of the menu.
+     *
+     * @return containerElement
+     */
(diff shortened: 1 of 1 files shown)


TobseF:
Yes `getElement()` is a logical and working alternative. But isn't better to make it clear what i get?
`getContainerElement()` //Returns the containing element of the menu. **vs**
`getElement()`  // the object's browser element
But it's true, even in the `PopupPanel` the `getContainerElement()` returns `getElement()`. In our case the #10259 was a breaking change, which was easy to fix. Se yes if its the goal to get rid of the `getContainerElement()` with the same value and breaking changes are no problem, we should close my MRE.

PR #11444, opened 2019-02-01
Fix an NPE caused by the previous fix to Grid's frozen columns

### LINKED ISSUE #10653: Grid frozen columns state updating
- Vaadin Framework version (e.g. 8.3.1)
- Chrome
- Description:
  Grid doesn't update frozen columns after removing columns and setting 
the same 'frozenColumnCount'

- Minimal reproducible example:

        Grid<Person> grid = new Grid<>(Person.class);

        addGridColumns();
        grid.setFrozenColumnCount(3);   // all is working, Grid displays frozen columns

        // attach Grid

        grid.removeAllColumns();
        addGridColumns();
        grid.setFrozenColumnCount(3);   // Grid doesn't show frozen columns

        // after this, 'grid.getFrozenColumnCount()' still returns '3';
        // grid state also still keeps '3' as 'frozenColumnCount', but there is no result...
        // NOTE: setting the different 'frozenColumnCount' value updates the Grid

mlindfors:
If the user managed to get two frozen column count changes into a single server round-trip before the component had been attached, the previous fix lead to a null pointer exception since there was no diff state available.

This fix adds a null check before removing the frozen column count value from the diff state so that this will not happen.

Closes #10653

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11444)


anasmi:
The check against null is a valid enhancement and fix for the issue (this particular pr has no problems). Through regarding the original fix, why you are removing the property, but not using `                                    updateDiffstate("frozenColumnCount", Json.createObject());` ? I didn't check further yet,but is property re-added somewhere then ?

mlindfors:
> why you are removing the property, but not using ` updateDiffstate("frozenColumnCount", Json.createObject());` ? I didn't check further yet,but is property re-added somewhere then ?

It's been a while, but if memory serves the property gets put back there with the next round-trip if you have set a frozenColumnCount. I just wanted to make sure all the things happen properly at client side as well, since both the server side and the client side fiddle around with the value whenever all columns are removed, but neither of them touch the diffState, which led to the original issue. I also had already wasted too much time on it, and just removing the property was the first thing which came to mind and worked. I suppose emptying the value would work just as well, but barring this one NPE issue I haven't noticed any problems with the approach.

anasmi (commented):
> Reviewed 1 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11444)* status: 1 of 3 files reviewed, all discussions resolved

anasmi (approved):
> Reviewed 2 of 3 files at r1, 3 of 3 files at r2, 4 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11444)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #11437, opened 2019-01-25
Setting autocomplete to a random number prevents setting autocomplete to 'off'


mcsherrylabs:
Google Chrome 71.0.3578.80 (Official Build) on Ubuntu
Chromium Version 71.0.3578.98 (Official Build) on Ubuntu
Vaadin 8.5.2 (but I note the code has not changed in latest 8.x release)
ComboBox  see #7820

![combo_problem](https://user-images.githubusercontent.com/6160346/51733633-d3ea7380-2079-11e9-82d4-104390fac0d1.png)

Problem -
In the above setup chrome autocomplete  boxes overshadow the dropdown making for a horrible UX. 
Using viritin HtmlElementPropertySetter doesn't override the random number set in referenced PR

Desired Outcome - 
Setting the autocomplete property to "off" by editing the DOM in chrome developer tools prevents the chrome options popping up. This is the desired outcome.

This seems like a 'simple fix' ... replace the random number generation with "off". I have not checked this on Firefox or IE. --reaches for helmet--

mcsherrylabs:
Guys, I don't mind fixing this, it's a simple fix, but ... 

Can I get some indication of whether setting the 

```autocomplete=off```

 instead of a random number is an acceptable fix please?

TatuLund:
The inconvenience about this is that Chrome is not supporting autocomplete="off" at the moment as we would wish. This is a bit of a problem with all frameworks, not just Vaadin. There are lengthy discussions about this in various places, notably this thread at StackOverflow

https://stackoverflow.com/questions/15738259/disabling-chrome-autofill

as well as the change request in Chromium issue tracker

https://bugs.chromium.org/p/chromium/issues/detail?id=587466

The most common way to tackle this has been using random numbers there, but it is not fool proof.

mcsherrylabs:
Thanks for the reply @TatuLund 

It looks like this was fixed in chrome -> https://codereview.chromium.org/1473733008

This test confirms my experience -> https://stackoverflow.com/a/52082658/10124106

Try it! (at your leisure)

My issue is not with password fields by the way, just combo box dropdowns.

TatuLund:
Note, there still issues with Chrome regarding autocomplete=off setting:

https://bugs.chromium.org/p/chromium/issues/detail?id=928305&can=1&q=component%3AUI%3EBrowser%3EAutofill

mcsherrylabs:
[...]

PR #11472, opened 2019-02-27
Fixing autocompletion issue with ComboBox on newer Chrome versions

TatuLund:
Newer Chrome versions do not work with random number hack to prevent auto completion, but it finally supports autocomplete=off.

Fixes https://github.com/vaadin/framework/issues/11437

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11472)


knoobie:
Would it be worth to investigate if the random number can be removed for all browsers?

TatuLund:
> Would it be worth to investigate if the random number can be removed for all browsers?

Maybe.

Note, it seems that support for autocomplete=off is not yet perfect in Chrome. But I assume that it works better than random number.

https://bugs.chromium.org/p/chromium/issues/detail?id=928305&can=1&q=component%3AUI%3EBrowser%3EAutofill. S

ZheSun88 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11472)* status: all files reviewed, 1 unresolved discussion (waiting on @TatuLund)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11472#-La-CfQ7Bu9RH-y9nac4:-La-CfQ7Bu9RH-y9nac5:b-cu62sw) (no related file):*
The fix is fine, but I think we should add a test in case later there is something wrong that we even don't know. 

you can add a very simple IT test for `chrome` browser:

UI: add a `ComboBox` 

Test: 
```
WebElement input = findElement(By.className("v-filterselect-input"));
assertEquals("autocomplete attribute should be off on Chrome", "off", input.getAttribute("autocomplete") );
```

or it would be better that you can add a test for multi-browsers, for other browsers, you can check the value is not equal to `off`, as the attribute should be a random number. 

---

TatuLund:
I think the proposed method for test case is not working. In the past Chrome did show auto completion even when attribute was off, so assessment of attribute is useless.

ZheSun88 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11472)* status: all files reviewed, 1 unresolved discussion (waiting on @TatuLund)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11472#-La-CfQ7Bu9RH-y9nac4:-La-Hsm_3DdhMRdW5gaQ:bxcoyor) (no related file):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

The fix is fine, but I think we should add a test in case later there is something wrong that we even don't know. 

you can add a very simple IT test for `chrome` browser:

UI: add a `ComboBox` 

Test: 
```
WebElement input = findElement(By.className("v-filterselect-input"));
assertEquals("autocomplete attribute should be off on Chrome", "off", input.getAttribute("autocomplete") );
```

or it would be better that you can add a test for multi-browsers, for other browsers, you can check the value is not equal to `off`, as the attribute should be a random number. 
</blockquote></details>

if there is no autocompletion attribute, there should be an error thrown from the `findElement()` method right?
and on the other hand, currently, the chrome shows this attribute and this test works. so this test is used for testing the current chrome but not the former one maybe and informing us Chrome is doing some changes with this attribute in the future. 

---

TatuLund:
I meant that even if our testing of the attribute value passes, it really does not tell anything for sure how Chrome works. I.e. the auto completion popup may still appear there.

TatuLund:
*[a discussion](https://reviewable.io/reviews/vaadin/framework/11472#-La-CfQ7Bu9RH-y9nac4:-LbWwx_iAbh5Ezxpm7v6:b-m3u2dy) (no related file):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

if there is no autocompletion attribute, there should be an error thrown from the `findElement()` method right?
and on the other hand, currently, the chrome shows this attribute and this test works. so this test is used for testing the current chrome but not the former one maybe and informing us Chrome is doing some changes with this attribute in the future. 
</blockquote></details>

Automated test cannot be done since the autocomplete popup in Chrome is not a dom element that could be detected.

---

TatuLund (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11472)* status: all files reviewed, 1 unresolved discussion (waiting on @ZheSun88)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11472#-La-CfQ7Bu9RH-y9nac4:-LbWx6Ud4z9BOkoSHI6o:b-896fix) (no related file):*
<details><summary><i>Previously, TatuLund (Tatu Lund) wrote…</i></summary><blockquote>

Automated test cannot be done since the autocomplete popup in Chrome is not a dom element that could be detected.
</blockquote></details>

Done.

---

ZheSun88 (approved):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11472)* status: :shipit: complete! all files reviewed, all discussions resolved

ZheSun88 (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11472)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11473, opened 2019-02-27
Fix NPE in v7 compatibility Grid during datasource rebind

TobseF:
This fixes an **NPE** on the `com.vaadin.v7.client.widgets.Grid#setDataSource`. Calling it twice will cause a NPE because `dataSource.addDataChangeHandler(null);` is not allowed.

@see `ListDataSource`
> `com.vaadin.client.widget.grid.datasources.ListDataSource#addDataChangeHandler`
> `Objects.requireNonNull(dataChangeHandler,  "DataChangeHandler can't be null");`

@see `AbstractRemoteDataSource`
> `com.vaadin.client.data.AbstractRemoteDataSource#addDataChangeHandler`
> `Objects.requireNonNull(dataChangeHandler,  "DataChangeHandler can't be null");`
  
This broken behavior was introduced by
@OlliTietavainenVaadin in `8313623dd318c92aa49e9cf96268f5fff6ad6456` as
> v7 Grid performance improvements picked to compatibility package

This error is nasty because it happens only the second time you visit the component and it's only visible in JS Debug. So let's roll this line back to allow data rebind.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11473)

### DIFF
--- a/compatibility-client/src/main/java/com/vaadin/v7/client/widgets/Grid.java
+++ b/compatibility-client/src/main/java/com/vaadin/v7/client/widgets/Grid.java
@@ -7023,10 +7023,6 @@ public class Grid<T> extends ResizeComposite implements HasSelectionHandlers<T>,
 
         selectionModel.reset();
 
-        if (this.dataSource != null) {
-            this.dataSource.addDataChangeHandler((DataChangeHandler) null);
-        }
-
         this.dataSource = dataSource;
         dataSource.addDataChangeHandler(new DataChangeHandler() {
                     @Override
--- a/compatibility-client/src/main/java/com/vaadin/v7/client/widgets/Grid.java
+++ b/compatibility-client/src/main/java/com/vaadin/v7/client/widgets/Grid.java
@@ -4152,6 +4152,7 @@ public class Grid<T> extends ResizeComposite implements HasSelectionHandlers<T>,
      * on initialization, but not after that.
      */
     private DataSource<T> dataSource;
+    private Registration changeHandler;
 
     /**
      * Currently available row range in DataSource.
@@ -7023,8 +7024,13 @@ public class Grid<T> extends ResizeComposite implements HasSelectionHandlers<T>,
 
         selectionModel.reset();
 
+        if (changeHandler != null) {
+            changeHandler.remove();
+            changeHandler = null;
+        }
+
         this.dataSource = dataSource;
-        dataSource.addDataChangeHandler(new DataChangeHandler() {
+		changeHandler = dataSource.addDataChangeHandler(new DataChangeHandler() {
                     @Override
                     public void dataUpdated(int firstIndex, int numberOfItems) {
                         escalator.getBody().refreshRows(firstIndex, numberOfItems);
--- a/uitest/src/main/java/com/vaadin/tests/components/grid/GridRebindDataSourceV7.java
+++ b/uitest/src/main/java/com/vaadin/tests/components/grid/GridRebindDataSourceV7.java
@@ -1,0 +1,42 @@
+package com.vaadin.tests.components.grid;
+
+import com.vaadin.annotations.Widgetset;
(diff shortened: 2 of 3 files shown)


anasmi:
Hi,

thank you for reporting the problem and submitting the pull request to address it! We will try to investigate it as soon as possible. I am not sure, why we are adding the null there overall, but we should definitely fix it, that code doesn't throw NPE.

Thanks again!

TobseF:
I added as second commit the removal of the `DataChangeHandler`. Now it's just as in the v8 Grid. Likely this was the original intended behavior of setting the listener to `null`.

``` java
if (changeHandler != null) {
    changeHandler.remove();
    changeHandler = null;
}
```

pleku (changes requested):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11473)* status: all files reviewed, 1 unresolved discussion (waiting on @TobseF)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11473#-LaKMOEV-5LQun2ZEJ9d:-LaKMOEV-5LQun2ZEJ9e:byjvi4k) (no related file):*
Hi @tobsef . Thanks for the fix, the changes look fine to me.

It is slightly troubling that this will fork the V7 compatibility package Grid even further from the (last) V7 version, but we've recently discovered that has already happened quite (too) much, and thus I do NOT see any good reason why we would use that grounds to block this change that fixes a valid issue in the compatibility package.

However, could you please add a test that fails without the fix to https://github.com/vaadin/framework/tree/master/uitest/src/test/java/com/vaadin/v7/tests/components/grid and with reusing and or adding a new test UI to https://github.com/vaadin/framework/tree/master/uitest/src/main/java/com/vaadin/v7/tests/components/grid

or reuse an existing

---

pleku (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11473)* status: all files reviewed, 1 unresolved discussion (waiting on @TobseF)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11473#-LaKMOEV-5LQun2ZEJ9d:-LaKOXKV6oOtHaRxE_W8:b-kdfcjb) (no related file):*

Hi @tobsef . Thanks for the fix, the changes look fine to me.

It is slightly troubling that this will fork the V7 compatibility package Grid even further from the (last) V7 version, but we've recently discovered that has already happened quite (too) much, and thus I do NOT see any good reason why we would use that grounds to block this change that fixes a valid issue in the compatibility package.

However, could you please add a test that fails without the fix to https://github.com/vaadin/framework/tree/master/uitest/src/test/java/com/vaadin/v7/tests/components/grid and with reusing and or adding a new test UI to https://github.com/vaadin/framework/tree/master/uitest/src/main/java/com/vaadin/v7/tests/components/grid

or reuse an existing
</blockquote></details>

Here is more instructions on how to run an UI test https://github.com/vaadin/framework#running-a-ui-test

---

anasmi:
@pleku tests added

From the repository you have mentioned, the tests are not running locally, so added to the general folder. Also noticed, that there are a couple of other V7 related tests there

pleku (approved):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11473)* status: :shipit: complete! all files reviewed, all discussions resolved

pleku (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11473#-:-LazSsRG7jCuaN_Uyzvb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

anasmi:
@pleku could you merge it also? (As you did the review and then we don't need to update the branch to be even with master) Thanks!

PR #11513, opened 2019-03-26
Fix wrong cursor position in datefield text box after state change 

### LINKED ISSUE #11455: IE11 DateField, the cursor jumps to the beginning of input after first value change
Vaadin Framework version: 8.7.0
Browser version: IE11
Web container name and version: Jetty 9.4.12.v20180830
Description of the bug: After refreshing the page/view containing a datefield, and after changing the value of the DateField first time, the cursor jumps in the beginning of the input field. 
Minimal reproducible example: 
```
DateField df = new DateField();
layout.addComponent(df);
```

Expected behavior: The cursor should stay at the end of the input field after the value change as it works in the other browsers. Tested browsers with expected behavior: Chrome, Edge, Firefox

Actual behavior: See the description above and the attached screeshot: 
image

TatuLund:
This problem happens with IE11 only

Fixes #11455

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11513)


anasmi:
The `executeScript` function could be called without referencing driver directly. This will eliminate the failure why this pr is failing (but I am not sure what is the problem with referencing, it looks good otherwise)
Through, then JS fails with this error https://stackoverflow.com/questions/37297144/syntaxerror-unterminated-regular-expression-literal-javascript

I think the easiest way to get the cursor position is calling `return (long) executeScript("return arguments[0].selectionStart",
                element);`  where `elemen`t is the inputField of the `DateField`
For this to work, the inputField should be focused (only focused component can have the cursor position) and now the textField is not focused, as we are explicitly calling Tab out of it here : https://github.com/vaadin/framework/blob/master/uitest/src/test/java/com/vaadin/tests/components/datefield/DateFieldValidationErrorTest.java#L46. 

I am not sure yet, how to verify/test this properly.  At least, it's not the easiest use case to check for sure.

anasmi:
There is a file with only empty line added. I think this should be removed. Also, I would add UI test with the description, how fix should be verified manually

anasmi:
The fix seems to work for IE properly. On the other hand, now I was able to reproduce the similar problem on Chrome. The cursor position is set correctly for the first time, but not afterwards. For me it's immediately broken on the second time (positioned in the beginning on the input) Can be reproduced at sampler here: https://demo.vaadin.com/sampler/#ui/data-input/dates/popup-date-field

I think we should fix both Chrome and IE behaviour

anasmi:
Can't reproduce the problem in `Firefox`, but the issue occurs both in `Chrome` and `IE11`

TatuLund:
@anasmi I removed the empty line and added the fix to apply to Chrome as well.

anasmi:
Will manually verify the fix, but code seems not to be formatted. 
Also UI test class is missing

anasmi:
I've added the automated test to verify the fix as well as formatted the code.

The test has failed for FF on Firefox 45, but works when checked manually on FF66. This should be addressed based on the results from run here. (Otherwise, tests verifies, at which position the cursor is after selecting the date)

On another note, the cursor change is now visible in Chrome, after the change is applied. The jump is frustrating as from my point of view. But I am not sure, if it's a blocker or not, so @pleku it would be great to get your opinion also.

pleku (commented):
> On another note, the cursor change is now visible in Chrome, after the change is applied. The jump is frustrating as from my point of view. But I am not sure, if it's a blocker or not, so @pleku it would be great to get your opinion also.

@anasmi can you please elaborate WDYM:
1) the bug also manifests in Chrome and this PR only fixes it for IE11 ?
2) because of the changes done here, the bug is now on Chrome too but not in IE11 ? So this PR causes a regression ?

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11513#-:-LcFDiNn3wwgbDpR-d-2:b-rvisx1)* status: 0 of 3 files reviewed, all discussions resolved

anasmi:
@pleku The bug is also present in Chrome without the fix ( so the fix is not causing the regression) Difference is that in IE the bug occurs only once on the first selection, whereas in chrome it occurs **always after** the first one happened. (I tried to reproduce this scenario in the tests)  What I meant is that the fix while works is causing **visible**  jump in an input field in Chrome

pleku (changes requested):
> Reviewed 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11513)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @TatuLund)

___
*[client/src/main/java/com/vaadin/client/ui/datefield/TextualDateConnector.java, line 176 at r10](https://reviewable.io/reviews/vaadin/framework/11513#-LcFERiG8i0unAeZIZv_:-LcFERiG8i0unAeZIZva:b-t286ao) ([raw file](https://github.com/vaadin/framework/blob/9c268c62127bab5b592f8a302768bf80d6377c43/client/src/main/java/com/vaadin/client/ui/datefield/TextualDateConnector.java#L176)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
>         // Force cursor to its right place with IE11 & Chrome (#11455)
>         if ((BrowserInfo.get().isIE11() || BrowserInfo.get().isChrome())
>                 && getWidget().text.getCursorPos() == 0) {
>             getWidget().text.setCursorPos(getWidget().text.getValue().length());
>         }
> ```
</details>


So now, if the cursor is in the beginning, it will be moved to the end for every single state change...

So this is a hack just addressing the symptoms and not trying to fix the actual bug.

How about investigating what is moving the cursor to the beginning instead ?

TatuLund:
___
*[client/src/main/java/com/vaadin/client/ui/datefield/TextualDateConnector.java, line 176 at r10](https://reviewable.io/reviews/vaadin/framework/11513#-LcFERiG8i0unAeZIZv_:-LcaEstK0Ej1F5IbBBOY:bu2de44) ([raw file](https://github.com/vaadin/framework/blob/9c268c62127bab5b592f8a302768bf80d6377c43/client/src/main/java/com/vaadin/client/ui/datefield/TextualDateConnector.java#L176)):*

> ```
>         // Force cursor to its right place with IE11 & Chrome (#11455)
>         if ((BrowserInfo.get().isIE11() || BrowserInfo.get().isChrome())
>                 && getWidget().text.getCursorPos() == 0) {
>             getWidget().text.setCursorPos(getWidget().text.getValue().length());
>         }
> ```

So now, if the cursor is in the beginning, it will be moved to the end for every single state change...

So this is a hack just addressing the symptoms and not trying to fix the actual bug.

How about investigating what is moving the cursor to the beginning instead ?
</blockquote></details>

It is not pretty, I agree. But I think this is just due browser differences, hence quite hard to find cleaner solution. I however added extra condition there so that fix is applied hopefully only when actually needed.

TatuLund (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11513)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @pleku)

___
*[client/src/main/java/com/vaadin/client/ui/datefield/TextualDateConnector.java, line 176 at r10](https://reviewable.io/reviews/vaadin/framework/11513#-LcFERiG8i0unAeZIZv_:-LcaFC8FAAniGos7suGD:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/9c268c62127bab5b592f8a302768bf80d6377c43/client/src/main/java/com/vaadin/client/ui/datefield/TextualDateConnector.java#L176)):*
<details><summary><i>Previously, TatuLund (Tatu Lund) wrote…</i></summary><blockquote>

It is not pretty, I agree. But I think this is just due browser differences, hence quite hard to find cleaner solution. I however added extra condition there so that fix is applied hopefully only when actually needed.
</blockquote></details>

Done.

PR #11519, opened 2019-03-28
Clicking on slider makes handler move

### LINKED ISSUE #1496: Clicking on slider should change slider value
**Originally by @johndevs**

___
Currently clicking on the Slider only moves the keyboard focus to the slider. 

A more common behavior is that clicking on the slider sets the value either to a clicked position or moves the slider in some increment toward the clicked position.

___
 Imported from https://dev.vaadin.com/ issue #6622

anasmi:
Fixes #1496

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)


pleku (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)* status: all files reviewed, 4 unresolved discussions (waiting on @anasmi)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11519#-LbPK_3X-zOLpZ08xix6:-LbPK_3Y08Ld60gOnDNu:boay5wj) (no related file):*
I don't argue against that this change should have been done ages ago for (UX), but it is non a behavior change because with this patch the default way of how the slider works when clicked changes.

Thus any app updating to the version that would include this has this UX change.

Because the default behavior of the slider has been like this for ages, looking from academic perspective, it should be introduced as opt-in. Meaning that there would have to be server side API for enabling the behavior.

But, does @jouni have a strong opinion about this change, mainly like "this is the way it should have always worked, it should be changed now for default too just to improve the UX". Then we could just highlight it in the release notes for next minor, and preferably also have the option to disable this.

---
___
*[client/src/main/java/com/vaadin/client/ui/VSlider.java, line 278 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPMTx059zihNIqKer1:-LbPMTx059zihNIqKer2:b-1r9lut) ([raw file](https://github.com/vaadin/framework/blob/e3912d8f1e13a634a775ec43f88daf6bf3cfb149/client/src/main/java/com/vaadin/client/ui/VSlider.java#L278)):*
> ```
>         } else if (DOM.eventGetType(event) == Event.MOUSEEVENTS) {
>             if (DOM.eventGetType(event) == Event.ONMOUSEDOWN) {
> ```
So`ONMOUSEDOWN` included in `MOUSEEVENTS`, which means that all non-base mouse events go here. Only mouse down gets processed.

But later on there is also the line
```
        } else if (DOM.eventGetType(event) == Event.ONMOUSEDOWN) {
            feedbackPopup.show();
        }
```
Which was never-ever executed if I understand correctly.

So instead of adding more complexity to this horrible piece of code, you could clean it up a bit.  Not sure if anyone will ever end up here, but comments might help too (specially if you've understood something about why/how this works like it does).
___
*[uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java, line 9 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPOaCB-C-atakmyjiG:-LbPOaCB-C-atakmyjiH:bosvrdq) ([raw file](https://github.com/vaadin/framework/blob/e3912d8f1e13a634a775ec43f88daf6bf3cfb149/uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java#L9)):*
> ```Java
> import org.openqa.selenium.interactions.Actions;
> 
> public class SliderHandleBaseClickTest extends MultiBrowserTest {
> ```

This class is not formatted
___
*[uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java, line 11 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPOoM04mBQfgkOnM6J:-LbPOoM04mBQfgkOnM6K:bxqmxum) ([raw file](https://github.com/vaadin/framework/blob/e3912d8f1e13a634a775ec43f88daf6bf3cfb149/uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java#L11)):*
> ```
> testHandlerHasMoved()
> ```
IMO it would be better to also validate that the click does actually move the slider value to something that it should, like for instance clicking on the middle of a 200px wide slider with values of 0->10, current value of 0, should give roughly the value of 5 ?

anasmi (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @anasmi, @jouni, and @pleku)

*[a discussion](https://reviewable.io/reviews/vaadin/framework/11519#-LbPK_3X-zOLpZ08xix6:-Lc10K3ABf25xZbCWMn8:bmd343g) (no related file):*

I don't argue against that this change should have been done ages ago for (UX), but it is non a behavior change because with this patch the default way of how the slider works when clicked changes.

Thus any app updating to the version that would include this has this UX change.

Because the default behavior of the slider has been like this for ages, looking from academic perspective, it should be introduced as opt-in. Meaning that there would have to be server side API for enabling the behavior.

But, does @jouni have a strong opinion about this change, mainly like "this is the way it should have always worked, it should be changed now for default too just to improve the UX". Then we could just highlight it in the release notes for next minor, and preferably also have the option to disable this.
</blockquote></details>

I've added setter and the default value is false, to preserve compatibility.  Through, I would assume it to be the default. Could you check the naming and API? Maybe something else would be more suitable? For now it's `slider.setEnableClickHandler(true);`

---
___
*[client/src/main/java/com/vaadin/client/ui/VSlider.java, line 278 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPMTx059zihNIqKer1:-Lc10cWT98PUScspMSes:bdbhiqj) ([raw file](https://github.com/vaadin/framework/blob/8b3484d09851ad75db6dffc188c5aa4a96ecc4da/client/src/main/java/com/vaadin/client/ui/VSlider.java#L278)):*

> ```
>         } else if (DOM.eventGetType(event) == Event.MOUSEEVENTS) {
>             if (DOM.eventGetType(event) == Event.ONMOUSEDOWN) {
> ```
So`ONMOUSEDOWN` included in `MOUSEEVENTS`, which means that all non-base mouse events go here. Only mouse down gets processed.

But later on there is also the line
```
        } else if (DOM.eventGetType(event) == Event.ONMOUSEDOWN) {
            feedbackPopup.show();
        }
```
Which was never-ever executed if I understand correctly.

So instead of adding more complexity to this horrible piece of code, you could clean it up a bit.  Not sure if anyone will ever end up here, but comments might help too (specially if you've understood something about why/how this works like it does).
</blockquote></details>

The situation, actually, was other way around. The first block was never executed for the MOUSEDOWN event, so I think it's the reason, why the second one was added. Based on this, it was supposed to work like this changes enables right from the beggining, assuming `processBaseEvent` was called from ` else if (DOM.eventGetType(event) == Event.MOUSEEVENTS)`. I've removed it overall, because it adds nothing but complexity to understanding. It didn't work properly. The second one is left, as it's the one causing the popup to show in current implementation of Slider.(If you click on base, the pop-up will be displayed)
___
*[uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java, line 9 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPOaCB-C-atakmyjiG:-Lc11GMUC0bx4t2l2gPq:bb6un1j) ([raw file](https://github.com/vaadin/framework/blob/8b3484d09851ad75db6dffc188c5aa4a96ecc4da/uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java#L9)):*

This class is not formatted
</blockquote></details>

Done
___
*[uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java, line 11 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPOoM04mBQfgkOnM6J:-Lc11Hb_1c5p5tWa9ZF_:b-zahnf5) ([raw file](https://github.com/vaadin/framework/blob/8b3484d09851ad75db6dffc188c5aa4a96ecc4da/uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java#L11)):*

> ```
> testHandlerHasMoved()
> ```
IMO it would be better to also validate that the click does actually move the slider value to something that it should, like for instance clicking on the middle of a 200px wide slider with values of 0->10, current value of 0, should give roughly the value of 5 ?
</blockquote></details>

It didn't work reliably, when I tried it, but I will need to modify test anyway,so will take a look. 
on my todo

pleku (changes requested):
> Reviewed 2 of 3 files at r3, 3 of 4 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)* status: all files reviewed, 4 unresolved discussions (waiting on @anasmi and @pleku)

___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 260 at r5](https://reviewable.io/reviews/vaadin/framework/11519#-Lc1FopO1Xte5EG5JNHf:-Lc1FopO1Xte5EG5JNHg:b-olp2w4) ([raw file](https://github.com/vaadin/framework/blob/3514b44519ee833b529b7a750d649705b622249f/server/src/main/java/com/vaadin/ui/Slider.java#L260)):*
> ```Java
>         getState().resolution = resolution;
>     }
> ```

Missing Javadocs
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 261 at r5](https://reviewable.io/reviews/vaadin/framework/11519#-Lc1Frws9ypt0ZdBpc3c:-Lc1Frws9ypt0ZdBpc3d:b-oy4oxq) ([raw file](https://github.com/vaadin/framework/blob/3514b44519ee833b529b7a750d649705b622249f/server/src/main/java/com/vaadin/ui/Slider.java#L261)):*
> ```
> setEnableClickHandler(boolean enabledClickedHandler)
> ```
How about:
`setUpdateValueOnClick(boolean updateValueOnClick)` ?

I could come up with more alternatives, but this is the most descriptive and implicit.

Problem with `enableClickHandler` is that I don't directly know what click handler it is enabling ?
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 264 at r5](https://reviewable.io/reviews/vaadin/framework/11519#-Lc1FqSI40LZqr5S1BiR:-Lc1FqSI40LZqr5S1BiS:b-po6mmc) ([raw file](https://github.com/vaadin/framework/blob/3514b44519ee833b529b7a750d649705b622249f/server/src/main/java/com/vaadin/ui/Slider.java#L264)):*
> ```Java
>         getState().enableClickHandler = enabledClickedHandler;
>     }
> ```

Missing javadocs

anasmi (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @pleku)

___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 260 at r5](https://reviewable.io/reviews/vaadin/framework/11519#-Lc1FopO1Xte5EG5JNHf:-Lc5WuuI15qOrY3iFLPP:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/b9059006271860f51a7a2ab3fc25316d0f072811/server/src/main/java/com/vaadin/ui/Slider.java#L260)):*

Missing Javadocs
</blockquote></details>

Done.
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 261 at r5](https://reviewable.io/reviews/vaadin/framework/11519#-Lc1Frws9ypt0ZdBpc3c:-Lc5WvbVCa9SOn83fjnu:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/b9059006271860f51a7a2ab3fc25316d0f072811/server/src/main/java/com/vaadin/ui/Slider.java#L261)):*

> ```
> setEnableClickHandler(boolean enabledClickedHandler)
> ```
How about:
`setUpdateValueOnClick(boolean updateValueOnClick)` ?

I could come up with more alternatives, but this is the most descriptive and implicit.

Problem with `enableClickHandler` is that I don't directly know what click handler it is enabling ?
</blockquote></details>

Done.
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 264 at r5](https://reviewable.io/reviews/vaadin/framework/11519#-Lc1FqSI40LZqr5S1BiR:-Lc5Ww8j5xj5R7BFQKzK:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/b9059006271860f51a7a2ab3fc25316d0f072811/server/src/main/java/com/vaadin/ui/Slider.java#L264)):*

Missing javadocs
</blockquote></details>

Done.
___
*[uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java, line 11 at r2](https://reviewable.io/reviews/vaadin/framework/11519#-LbPOoM04mBQfgkOnM6J:-Lc5WxpTBOg0SDwRt-zn:bb6un1j) ([raw file](https://github.com/vaadin/framework/blob/8b3484d09851ad75db6dffc188c5aa4a96ecc4da/uitest/src/test/java/com/vaadin/tests/components/slider/SliderHandleBaseClickTest.java#L11)):*

It didn't work reliably, when I tried it, but I will need modify test,so will take a look. 
on my todo
</blockquote></details>

Done

pleku (changes requested):
> Reviewed 6 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)* status: all files reviewed, 3 unresolved discussions (waiting on @anasmi)

___
*[client/src/main/java/com/vaadin/client/ui/VSlider.java, line 679 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5YAL59ubJS44RJxG9:-Lc5YAL59ubJS44RJxGA:bcbl45i) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/client/src/main/java/com/vaadin/client/ui/VSlider.java#L679)):*
> ```Java
>     }
> 
>     public void setUpdateValueOnClick(boolean updateValueOnClick) {
> ```

Missing javadocs, but it is client side widget code... but this is anyway against the scouts rule
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 262 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5YQUm4qDMnj3mrjnn:-Lc5YQUm4qDMnj3mrjno:b-s9xrjq) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/server/src/main/java/com/vaadin/ui/Slider.java#L262)):*
> ```
> triggered by body.
> ```
Uhm, maybe better phrase it as (and add the default):
"...update its value when the user clicks on the slider. By default, the slider value is only updated by dragging the slider's handle."
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 265 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5YsddBm0TnPqnasK6:-Lc5YsddBm0TnPqnasK7:bstfwg0) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/server/src/main/java/com/vaadin/ui/Slider.java#L265)):*
> ```
>  click event on Slider's body should update the value
> ```
Weird order of words in the sentence, maybe:
`{@code true} to update the value of the slider on click, {@code false} for not`
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 273 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5Z8AoCd1cgRkgrK3n:-Lc5Z8AoCd1cgRkgrK3o:b6z4xtt) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/server/src/main/java/com/vaadin/ui/Slider.java#L273)):*
> ```
> Gets the {@code updateValueOnClick) value.
> ```
No. 

Please write proper javadocs instead of being lazy, this is as good as not having javadocs. You don't need to duplicate everything, you can also use the link `@see #foobarmethod`

anasmi (commented):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @pleku)

___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 262 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5YQUm4qDMnj3mrjnn:-Lc5hqA801BosPzBm6B5:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/server/src/main/java/com/vaadin/ui/Slider.java#L262)):*

> ```
> triggered by body.
> ```
Uhm, maybe better phrase it as (and add the default):
"...update its value when the user clicks on the slider. By default, the slider value is only updated by dragging the slider's handle."
</blockquote></details>

Done.
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 265 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5YsddBm0TnPqnasK6:-Lc5hrMO0DT5Ggqg4aTw:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/server/src/main/java/com/vaadin/ui/Slider.java#L265)):*

> ```
>  click event on Slider's body should update the value
> ```
Weird order of words in the sentence, maybe:
`{@code true} to update the value of the slider on click, {@code false} for not`
</blockquote></details>

Done.
___
*[server/src/main/java/com/vaadin/ui/Slider.java, line 273 at r6](https://reviewable.io/reviews/vaadin/framework/11519#-Lc5Z8AoCd1cgRkgrK3n:-Lc5hsL84ItG2gwuY60G:b-896fix) ([raw file](https://github.com/vaadin/framework/blob/d3c0d72449b4ae281c4174d07dba703721f37f19/server/src/main/java/com/vaadin/ui/Slider.java#L273)):*

> ```
> Gets the {@code updateValueOnClick) value.
> ```
No. 

Please write proper javadocs instead of being lazy, this is as good as not having javadocs. You don't need to duplicate everything, you can also use the link `@see #foobarmethod`
</blockquote></details>

Done.

pleku (approved):
:lgtm:

> Reviewed 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11519#-:-Lc5ilHs7yi9QqYMXq_P:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11556, opened 2019-05-02
Changed the handleSessionExpired logic to return a 404 instead of a 410 and added the no-cache parameter to the reply.

### LINKED ISSUE #4417: Heartbeat should respond 404, not 410, when session expired
**Originally by @jdahlstrom**

___
`HeartbeatHandler` currently sends 410 Gone if the session is expired. This is arguably semantically wrong, as a user agent is within its rights to assume that the resource uniquely identified by the URI (and only the URI - cookies don't matter) will never be available again and any future requests to the same URI should not be made. A 404 Not Found should be returned instead.

___
 Imported from https://dev.vaadin.com/ issue #12526

edler-san:
Fixes #4417

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11556)

### DIFF
--- a/server/src/main/java/com/vaadin/server/communication/HeartbeatHandler.java
+++ b/server/src/main/java/com/vaadin/server/communication/HeartbeatHandler.java
@@ -89,7 +89,14 @@ public class HeartbeatHandler extends SynchronizedRequestHandler
             return false;
         }
 
-        response.sendError(HttpServletResponse.SC_GONE, "Session expired");
+        // Ensure that the browser does not cache expired response.
+        // iOS 6 Safari requires this (#10370)
+        response.setHeader("Cache-Control", "no-cache");
+        // If Content-Type is not set, browsers assume text/html and may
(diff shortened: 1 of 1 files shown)


ZheSun88 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11556#-:-LdsHUOy5pHTcSSiNPG1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

ffdybuster:
There's a problem with this PR: the client behaves differently for 410 and 404: https://github.com/vaadin/framework/blob/master/client/src/main/java/com/vaadin/client/communication/DefaultConnectionStateHandler.java#L149-L156

(Also, there is another place which still returns 410 when the session is expired: https://github.com/vaadin/framework/blob/master/server/src/main/java/com/vaadin/server/VaadinService.java#L1762)

PR #11666, opened 2019-07-26
Update line height when new row height is set for Grid

### LINKED ISSUE #9991: Grid.setRowHeight should also set line-height to avoid uglyness
I use

> myGrid.getColumn("foo").setStyleGenerator(item -> "v-align-middle");

to middle-align my Grid contents after setting row height with:

>myGrid..setRowHeight(42);

However, when focussing a cell, the focused cell it top-aligned and looks like this:
image

The issue seems to be that `height: 42px;` is set, but `line-height` is not.
If I set`line-height: 42px;` it revolves the issue.

This seems to be a sensible default to me anyway, since it prevents text from sticking to the top of the row in any case.

TatuLund:
Fixes https://github.com/vaadin/framework/issues/9991

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11666)


TatuLund:
This approach does not work.

PR #11739, opened 2019-10-02
Firefox for iOS: avoid "too old" error

### LINKED ISSUE #11296: Error Message at startup when people use Firefox on IOS
On Vaadin 8.6.0 with the last Firefox on Iphone (Iphone X - IOS 12.0.1)

When the user connect to my Vaadin Application, a message is displayed:

errorfirefox

ghost:
Fixes the "too old" error shown to Firefox for iOS users. Also handles that all iOS browsers must use Safari.

Fixes #11296

[ ] Valid tests for the pull request
[x] Contributing guidelines implemented

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11739)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -20,7 +20,7 @@
         <vaadin.spring.version>3.2.1</vaadin.spring.version>
         <vaadin.testbench.version>5.2.0</vaadin.testbench.version>
         <vaadin.cdi.version>3.0.1</vaadin.cdi.version>
-        <vaadin.context-menu.version>3.0.0</vaadin.context-menu.version>
+        <vaadin.context-menu.version>3.0.2</vaadin.context-menu.version>
     </properties>
 
     <repositories>
--- a/client/src/main/java/com/vaadin/client/widgets/Grid.java
+++ b/client/src/main/java/com/vaadin/client/widgets/Grid.java
@@ -7751,8 +7751,8 @@ public class Grid<T> extends ResizeComposite implements HasSelectionHandlers<T>,
             return;
         }
 
-        Element e = Element.as(target);
-        RowContainer container = escalator.findRowContainer(e);
+        Element element = Element.as(target);
+        RowContainer container = escalator.findRowContainer(element);
         Cell cell;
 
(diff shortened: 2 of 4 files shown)


anasmi:
Thank you for your contribution and creating a pull request!

TatuLund:
Thanks for your contribution, I checked it and would like you to ask some changes

The change you proposed will skip this version check when on iOS

```
        // Firefox for iOS uses a different versioning scheme and will
        // fail the test. Since it is already covered by the iOS test
        // above, ignore it here.
        if (isFirefox() && !isIOS() && getBrowserMajorVersion() < 45) {
            return true;
        }
```

This would be ok, if all Firefox versions ever released on iOS are not too old. I doubt this is the case. So afer above change we should add also

```
        if (isFirefox() && isIOS() && getBrowserMajorVersion() < X) {
            return true;
        }
```

And now the question is that what is the right value of X?

ffdybuster:
But all Firefox for iOS versions (must) use Safari, and are thus already covered by the check `if ((isSafari() || isIOS()) && getBrowserEngineVersion() < 533) {` a few lines above.

If that ever changes, though, the versions will be new enough that they shouldn't be "too old" :)

TatuLund:
@ffdybuster Thanks for your quick reply. I think then it is ok.

TatuLund (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11739#-:-LqAeqC24vg0Fi0DxUa0:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

TatuLund (approved):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11739)* status: :shipit: complete! all files reviewed, all discussions resolved

ffdybuster:
@anasmi @TatuLund @ZheSun88 thanks a lot for your feedback and for getting this merged! :)

ZheSun88:
@ffdybuster Thanks for the contribution. this will be included in the framework next release. `8.9.1`

PR #11771, opened 2019-10-28
Fix the end limit of the allowed scrollTo rows.

Ansku:
- Row index counts up from zero, data provider size counts up from one,
as one would expect. If the two match we are already past the available
range.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11771)

PR #11774, opened 2019-10-28
Fix the end limit of the allowed scrollTo rows. (#11771)

Ansku:
- Row index counts up from zero, data provider size counts up from one,
as one would expect. If the two match we are already past the available
range.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11774)


ZheSun88 (approved):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11774)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #11718, opened 2019-09-11
ComboBox popup behaves badly when ComboBox is placed to right edge


TatuLund:
Vaadin 8.8.6
Browser: Chrome 76, Edge, probably others too

Steps to replicate

- Create UI where ComboBox is placed on the right edge of the UI. 

- Add items, which are wider than ComboBox itself.

- comboBox.setPopupWidth(null) (or 100% or something that is wider than ComboBox)

- Open the ComboBox

- Hover mouse to first item on popup (popup will shift slightly to right, pushing UI to the left)

- Hover mouse back to the text box (you will see that popup extends outside of the screen)

- Hover mouse back to the first item (popup will shift even more to the right, pushing UI to the left)

Originally reported in Forum (including video)

https://vaadin.com/forum/thread/17826265/combobox-popup-on-far-right-modifies-main-layout

anasmi:
Code to reproduce: 
```
 ComboBox<String> nameComboBox = new ComboBox<>("Long items");
 nameComboBox.setPopupWidth(null);
 nameComboBox.setItems(Arrays.asList("First Very long item to add",
                  "Second very long item to add", "Third very long item to add"));
                
addComponent(nameComboBox);
 Page.getCurrent().getStyles().add(".v-slot.v-slot-positionRight {float:right;}");
 Page.getCurrent().getStyles().add(".removePadding {padding: 0px !important;}");
 getLayout().setMargin(false);
 getLayout().setSpacing(false);
 getLayout().getParent().addStyleName("removePadding");
 getLayout().getParent().removeStyleNames("v-margin-right","v-margin-left");
 nameComboBox.addStyleName("positionRight");
```

PR #11809, opened 2019-11-13
Reduce excess positioning calls for ComboBox popup. (#11808)

Ansku:
- If an open popup is reset to its default position on every update from
the server before getting adjusted again to the actual expected
position, on heavier applications some of those intermediate positions
might get rendered. If the ComboBox is positioned at the right edge and
the popup contents are longer than the input field (i.e. popup should
open to the left, not right) this might cause flickering.
- Setting the default position is only actually needed when the popup is
opened in order to give it a baseline, otherwise it's better to simply
adjust the position if needed.

Continues on #11718

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11809)


TatuLund (approved):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11809)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11811, opened 2019-11-13
Added setButtonCaptionAsHtml functionality to Upload component.

### LINKED ISSUE #11810: Add setButtonCaptionAsHtml method to Upload component
- Vaadin Framework version (e.g. 8.9.2)

Upload button component allows me to customize caption on the upload button only with strings. It is not possible to set arbitrary html into it. 

I want to be able to do this:
````java
Upload upload = new Upload(null, this::receiveUpload);
upload.setButtonCaptionAsHtml(true);
upload.setButtonCaption(VaadinIcons.UPLOAD.getHtml());
````

And see this on screen:
image

Normal buttons have such functionality.

SomMeri:
Add setButtonCaptionAsHtml method to Upload component

Fixes #11810

Usage example:
````java
Upload upload = new Upload(null, this::receiveUpload);
upload.setButtonCaptionAsHtml(true);
upload.setButtonCaption(VaadinIcons.UPLOAD.getHtml());
````

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11811)


SomMeri:
I would take a look at TeamCity failure, but I dont have access there.

Ansku:
This is a valid way to solve the issue but we are still debating whether we'd rather tie the fix to the main setCaptionAsHtml logic (so that both the button caption and the component caption would either be in html mode or not) or, if kept separate, would it be better to add the variable to shared state even if the client-side caption handling has to be done within the legacy uidl logic.

SomMeri:
@Ansku thank you for feedback.

Ansku:
Fixed in #11940

PR #11830, opened 2019-11-23
fix autocomplete in TextField

MidNight-er:
we already have fixed Invalid content for attribute autocomplete for comboboxes in this issue https://github.com/vaadin/framework/issues/7820. let's fix it for TextField.
Please reopen this bug https://github.com/vaadin/framework/issues/7823 and apply my changes in order to fix it

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11830)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/textfield/TextFieldConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/textfield/TextFieldConnector.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.client.ui.textfield;
 
+import com.vaadin.client.BrowserInfo;
 import com.vaadin.client.event.InputEvent;
 import com.vaadin.client.ui.VTextField;
 import com.vaadin.shared.ui.Connect;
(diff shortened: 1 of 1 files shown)


TatuLund:
@MidNight-er thanks for your suggestion and pull request. However we think this should be opt-in feature, since in some cases autocomplete is actually wanted. So this change would break those applications where autocompletion is a desired behavior. We created a small add-on while back for this purpose, which can be found in Directory https://vaadin.com/directory/component/autocompleteoff-extension-add-on

MidNight-er:
> @MidNight-er thanks for your suggestion and pull request. However we think this should be opt-in feature, since in some cases autocomplete is actually wanted. So this change would break those applications where autocompletion is a desired behavior. We created a small add-on while back for this purpose, which can be found in Directory https://vaadin.com/directory/component/autocompleteoff-extension-add-on

@TatuLund I agree with you. There is some cases where autocomplete is really useful. From the other hand the vast majority cases related to autocomplete in a single page application is an annoying feature. For a single page application autocomplete just offer some random words or sentences that user fill once in the past. I try to apply add-on that you offer, but I think it's better to use add-on to enable autocomplete feature in one page application with disabled it by default. What do you think about it?

Kalomo:
I would also appreciate that the autocomplete will be an vaadin framework feature and can be enabled or disabled depending in user needs.

TatuLund:
> I would also appreciate that the autocomplete will be an vaadin framework feature and can be enabled or disabled depending in user needs.

The behavior Chrome has been changing version to version, thus it has been hard to create stable approach in our framework for this. Maintaining add-on solution is more agile way to deal with this at the time being. We may consider integrating the approach that is in the add-on in later time to the framework. Thus I am closing this PR for now.

PR #11832, opened 2019-11-25
Delegate enabled handling to Composite root.

### LINKED ISSUE #11831: Tree looks like it's enabled if set as disabled when attaching it
The following happens with Vaadin 8.9.2, tested on Firefox 70.0.1, Chrome 78.0.3904.97, and Safari 12.1.2:
When creating a new Tree and setting it disabled right away, it behaves like it's disabled, but doesn't look like it. Apparently the proper styles are not applied if the disabling is done when attaching the component.

For the sample UI, see the example code in #11822 .

Steps to reproduce:
* run the UI
* look at the Tree. It's disabled, but looks like normal.
* if you press the button twice, now the Tree is both disabled and looking like it

Expected behaviour:
* the Tree has the disabled style applied always when it's disabled

Actual behaviour:
* the Tree doesn't always look like it's disabled when it should

Ansku:
Otherwise the changed state isn't communicated properly to the
client-side in the initial round trip, as the client-side uses the child
connector's state directly.

Fixes #11831

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11832)

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Composite.java
+++ b/server/src/main/java/com/vaadin/ui/Composite.java
@@ -216,6 +216,16 @@ public class Composite extends AbstractComponent implements HasComponents {
     }
 
     @Override
+    public void setEnabled(boolean enabled) {
+        getRootOrThrow().setEnabled(enabled);
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return getRootOrThrow().isEnabled();
+    }
+
+    @Override
     public float getWidth() {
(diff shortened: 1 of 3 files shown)


TatuLund (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11832#-:-LuaTGZQ-EyaykJSM_ed:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #11672, opened 2019-07-29
Grid Editor is not opening at last row on 110% page scale


AlexByte:
- Vaadin Framework version
8.8.5
- Browser version
Chrome 75
- Description of the bug
Grid Editor is not opening at last row on 110% page scale on double click with mouse.
- Minimal reproducible example
https://demo.vaadin.com/sampler/#ui/grids-and-trees/grid/item-editor
![grid editor2](https://user-images.githubusercontent.com/25251772/62046594-fddf4700-b210-11e9-9713-511d72b112b1.gif)

- Expected behavior
Editor should be opened at last row on double mouse click at 110% page scale.
- Actual behavior
Editor is opening only after double click on last row and pulling the scroll bar to the end or and  scrolling up or and setting scale to 100%.

TatuLund:
I tested this, and I could not reproduce the problem using Chrome Version 75.0.3770.142 (Official Build) (64-bit) and Windows 10

AlexByte:
@TatuLund Can you record video? Have you tested at 110% scale?

AlexByte:
I think this is due to the row bounce after scrolling to the end.
![scrolling](https://user-images.githubusercontent.com/25251772/62047496-50216780-b213-11e9-9fd4-fc3c41c195d8.gif)

SonReimer:
This is the same under Firefox 60.7.0esr (64-Bit) with 100% scale using the Item Editor demo above

Ansku:
There seems to be an uncaught client-side exception. `ScrollbarBundle.getScrollPos()` triggers an assertion error due expected and calculated scroll positions being one pixel off of each other, possibly thanks to the zoom causing a different rounding result.

SonReimer:
After updating vaadin dependencies to 8.9.3 in my application and clearing the cache problem still resides in my application (with 100% scale). It's the same behavior with 
https://demo.vaadin.com/sampler/#ui/grids-and-trees/grid/item-editor (with 100% scale)

SonReimer:
I've tested this under Firefox Version 60.7.0esr (64-Bit) and Chrome Version 78.0.3904.108 (64-Bit)

Ansku:
Hm, interesting. Now that we've got the sampler updated I can indeed still reproduce the problem there, but not in my local dev environment. I'll look into it.

PR #11835, opened 2019-12-02
Simplify Grid scroll handling.

Ansku:
If first attempt at scrolling doesn't succeed it's unlikely that
continuing to wait is going to make any difference. Cache should be
populated before triggering any actions that depend on the row being
visible, otherwise it should be enough to trust that scrollToRow
actually scrolls to row and once scrolling is done the row is as much in
view as it's going to get. This way we don't get into a situation where
Editor never opens because it's still waiting for that one last pixel
that can't be achieved thanks to browser zoom causing rounding errors.

Continues on #11672

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11835)


TatuLund (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11835#-:-LvASgH259dI3CK5JYni:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11846, opened 2019-12-17
Improvements to popup positioning for ComboBox within HorizontalLayout.

Ansku:
Expand ratio and spacing can cause ComboBox to miscalculate its own
position while layouting is still ongoing. Popup should not be
repositioned in such circumstances in order to avoid incorrect
intermediate states.

Continues on #11718

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11846)

PR #11852, opened 2019-12-19
Improvements to popup positioning for ComboBox within HorizontalLayout.

Ansku:
(#11846)

Expand ratio and spacing can cause ComboBox to miscalculate its own
position while layouting is still ongoing. Popup should not be
repositioned in such circumstances in order to avoid incorrect
intermediate states.

Continues on #11718

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11852)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VComboBox.java
+++ b/client/src/main/java/com/vaadin/client/ui/VComboBox.java
@@ -81,6 +81,7 @@ import com.vaadin.client.ui.aria.HandlesAriaRequired;
 import com.vaadin.client.ui.combobox.ComboBoxConnector;
 import com.vaadin.client.ui.menubar.MenuBar;
 import com.vaadin.client.ui.menubar.MenuItem;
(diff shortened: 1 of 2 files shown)


anasmi (approved):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11852)* status: :shipit: complete! all files reviewed, all discussions resolved

dyachera:
I just recently updated to 8.9.4 and it looks like this caused the combo box on one of my windows to no longer function correctly so I cannot upgrade to latest.  Before this I had a combobox on a window in a horizontalLayout at the bottom of the window and when you opened up the options the select options rendered above the combobox so you could see all available options.  After upgrading to 8.9.4 when I go to click on the comboBox I can only see one option and all the others render off the screen.  I have to hover over the first option and press the down arrow key in order for them to be shown.  This does not happen in 8.9.3 or 8.9.2

PR #11865, opened 2020-01-09
Attempt of making test case more stable

TatuLund:
Changing the date field to be not readonly in the beginning and toggling readonly after testing read enabled function before testing readonly.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11865)


TatuLund:
Note: removed initial-date and readonly-date screenshots are equal. So removal of the initial-date screenshot test does not actually reduce testing coverage.

TatuLund:
Superced by https://github.com/vaadin/framework/pull/11867

PR #11900, opened 2020-02-26
Adding missing Aria attributes

### LINKED ISSUE #11897: [Accessibility] ComboBox: missing required wai-aria-attributes
The current implementation of Combobox isn't fully accessible. (V7+V8)

TLDR; Combobox is missing the appropriate aria-* attributes:

* aria-expanded
* aria-owns
* aria-haspopup

[Example how it should behave](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#ex3_label)
[Example how it should be done (source code) ](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#sc3_label)

See the MUST criteria of W3C:

> Authors MUST ensure an element with role combobox contains or owns a text input element with role textbox or searchbox and that the text input has aria-multiline set to false. If the combobox provides autocompletion behavior for the text input as described in aria-autocomplete, authors MUST set aria-autocomplete on the textbox element to the value that corresponds to the provided behavior.
> 
> Typically, the default state of a combobox is collapsed. In the collapsed state, only the textbox element of a combobox is visible. A combobox is said to be expanded when both the textbox and a secondary element that serves as its popup are visible. Authors MUST set aria-expanded to true on an element with role combobox when it is expanded and false when it is collapsed. Elements with the role combobox have an implicit aria-expanded value of false.
> 
> When a combobox is expanded, authors MUST ensure it contains or owns an element that has a role of listbox, tree, grid, or dialog. This element is the combobox popup. When the combobox is expanded, authors MUST set aria-controls on the textbox element to a value that refers to the combobox popup element.
> 
> Elements with the role combobox have an implicit aria-haspopup value of listbox. If the combobox popup element has a role other than listbox, authors MUST specify a value for aria-haspopup that corresponds to the type of its popup.

Source: https://www.w3.org/TR/wai-aria-1.1/#combobox

TatuLund:
Fixes: https://github.com/vaadin/framework/issues/11897

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11900)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VComboBox.java
+++ b/client/src/main/java/com/vaadin/client/ui/VComboBox.java
@@ -26,6 +26,8 @@ import java.util.Set;
 import java.util.logging.Logger;
 
 import com.google.gwt.animation.client.AnimationScheduler;
+import com.google.gwt.aria.client.ExpandedValue;
+import com.google.gwt.aria.client.Id;
 import com.google.gwt.aria.client.Roles;
 import com.google.gwt.core.client.JavaScriptObject;
 import com.google.gwt.core.client.Scheduler;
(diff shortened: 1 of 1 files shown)


knoobie (changes requested):
Blocked by discussion in https://vaadin.com/pro/support#3307

PR #11912, opened 2020-03-06
Trigger re-measure after updating ElementResizeListeners.

### LINKED ISSUE #10734: setExpandRatio breaks layout when component is updated
Seen on multiple browsers (Chrome 65.0, FireFox 58.0.2, Safari 11.0.3), with Vaadin 8.3.2.

Setting an expand ratio on a component in a HorizontalLayout works fine initially but when the component is upgraded (for instance: setting value of a ProgressBar or setting the input prompt of a TextField), the component suddenly becomes wider than the HorizontalLayout.

Example code:
```
@Push
public class MyUI extends UI
{
    @Override
    protected void init(VaadinRequest vaadinRequest)
    {
        ProgressBar progress = new ProgressBar();
        progress.setWidth(100, Unit.PERCENTAGE);
        Button button = new Button("Cancel");

        HorizontalLayout layout = new HorizontalLayout(progress, button);
        layout.setExpandRatio(progress, 1);
        layout.setWidth(100, Unit.PERCENTAGE);
        setContent(layout);

        UI ui = UI.getCurrent();
        Executors.newScheduledThreadPool(1).scheduleAtFixedRate(() -> update(ui, progress), 4, 2, TimeUnit.SECONDS);
    }

    public void update(UI ui, ProgressBar progress)
    {
        ui.access(() -> {
            float value = progress.getValue();
            value = (value >= 1) ? 0 : value + 0.1f;
            System.err.println("new value: " + value);
            progress.setValue(value);
        });
    }

    public static class MyUIServlet extends VaadinServlet
    {
    }
}
```

Initial (correct) screenshot:

before

Issue visible after ProgressBar.setValue() call. Note how the Cancel button is pushed out of the layout.
after

The same effect can be achieved by replacing the `ProgressBar` with a `TextField` and using `TextField::setPlaceHolder` in the `update` method.

Ansku:
Removing ElementResizeListeners from an element makes it unmeasurable
and clears any saved measured values. Adding the listeners back makes
the element measurable again but doesn't add it to measuring queue.
Measuring needs to happen or any updates to expanded components within a
layout (without changes that would trigger full re-measuring of the
layout itself) lead to broken expand size calculations with any fixed
size elements assumed to have no size.

Fixes #10734

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11912)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/orderedlayout/AbstractOrderedLayoutConnector.java
+++ b/client/src/main/java/com/vaadin/client/ui/orderedlayout/AbstractOrderedLayoutConnector.java
@@ -570,11 +570,13 @@ public abstract class AbstractOrderedLayoutConnector
         }
 
         // Add all necessary listeners
+        boolean listenersAdded = false;
         if (needsFixedHeight()) {
             slot.setWidgetResizeListener(childComponentResizeListener);
             if (slot.hasCaption()) {
                 slot.setCaptionResizeListener(slotCaptionResizeListener);
             }
+            listenersAdded = true;
         } else if ((hasChildrenWithRelativeHeight
                 || hasChildrenWithRelativeWidth) && slot.hasCaption()) {
             /*
@@ -586,6 +588,7 @@ public abstract class AbstractOrderedLayoutConnector
              * as the relative size?
              */
             slot.setCaptionResizeListener(slotCaptionResizeListener);
+            listenersAdded = true;
         }
 
         if (needsExpand()) {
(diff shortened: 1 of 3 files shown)


TatuLund (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11912#-:-M1zlHDf6xjY9-ALOTz5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11915, opened 2020-03-11
Remove unnecessary null-check at ValueContext constructors

kos672:
Hi Vaadin devs.
I've found such inconsistency in your sources.
In constructor of the ValueContext there is null-check for Component argument. But the Javadoc says that it can be null. My proposal is to delete such null-check. AFAIK in later versions there are more constructors available, and also those which have more arguments. Some of them allow to give component as null, and others no. As an advice - please check it on newer versions.

Besides that there is also adjusted logic for getting component from VC using Optional#ofNullable.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11915)


Ansku (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11915#-:-M4J8XD3Cp_VZnfKh8yS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11936, opened 2020-03-31
Ensure ChangeListener still works after first upload.

### LINKED ISSUE #10420: Upload component:  ChangeListener only fires once
Component:  `Upload`

Problem: The `Upload.ChangeListener` only fires once.
Vaadin Framework version:   8.1.7  (latest and greatest, also reproduced on 8.2beta)

## How to reproduce:

Use the below minimal example:

```java
public class MyUI extends UI 
        implements Upload.Receiver, Upload.ChangeListener {

    private static final Logger LOGGER = Logger.getLogger(MyUI.class.getName());
    
    @Override
        
        Upload upload = new Upload("Upload file", this);
        upload.setImmediateMode(false);
        upload.addChangeListener(this);
        layout.addComponents(upload);
        setContent(layout);
    }

    @Override
    public OutputStream receiveUpload(String filename, String mimeType) {
        // a NULL outputstream, does nothing
        return new OutputStream() {
            @Override
            public void write(int b) throws IOException {
            }
        };
    }

    @Override
    public void filenameChanged(Upload.ChangeEvent event) {
        LOGGER.log(Level.INFO, "filenameChange() called !!");
    }

    }
}

```

Steps:

- Start the web app.
- Pick a file and press "Upload". Check in the log that the filenameChanged() method has fired.
- Repeat previous step (possibly with another file this time). Notice that filenameChanged() method doesn't fire again.

## Why is this serious ?

The Upload component has a [usability bug](https://github.com/vaadin/framework/issues/10419) in that allows the "Upload" button (in non-immediate mode) to be pressed, even if no file is selected. To workaround this, you can use the `Upload.ChangeListener` (together with `Upload.FinishedListener`) to enable/disable the button that starts the upload. However, if `Upload.ChangeListener` doesn't really work (except first time), then such a workaround won't work.

Ansku:
Fixes #10420

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11936)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/VUpload.java
+++ b/client/src/main/java/com/vaadin/client/ui/VUpload.java
@@ -40,6 +40,7 @@ import com.vaadin.client.ConnectorMap;
 import com.vaadin.client.StyleConstants;
 import com.vaadin.client.ui.upload.UploadConnector;
 import com.vaadin.client.ui.upload.UploadIFrameOnloadStrategy;
+import com.vaadin.shared.EventId;
 import com.vaadin.shared.ui.upload.UploadServerRpc;
 
 /**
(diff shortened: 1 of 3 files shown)


OlliTietavainenVaadin (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11936#-:-M3op4r-ASJQpqXSCcc4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #11944, opened 2020-04-08
Allow UIDL behind reverse proxy by updating hasPathPrefix

cvs-orpheus:
Vaadin-spring behind a reverse proxy did not work, because UIDL requests did not reach the UidlRequestHandler. The problem was that ServerletPortletHelper.hasPathPrefix(...) checked for "/UIDL/" but the incoming prefix was "/UIDL". With the changes vaadin-spring works behind a reverse proxy.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11944)

### DIFF
--- a/server/src/main/java/com/vaadin/server/ServletPortletHelper.java
+++ b/server/src/main/java/com/vaadin/server/ServletPortletHelper.java
@@ -87,7 +87,7 @@ public class ServletPortletHelper implements Serializable {
         }
     }
 
-    private static boolean hasPathPrefix(final VaadinRequest request, String prefix) {
+    private static boolean hasPathPrefix(VaadinRequest request, String prefix) {
         String pathInfo = request.getPathInfo();
 
         if (pathInfo == null) {
--- a/server/src/main/java/com/vaadin/server/ServletPortletHelper.java
+++ b/server/src/main/java/com/vaadin/server/ServletPortletHelper.java
@@ -87,15 +87,27 @@ public class ServletPortletHelper implements Serializable {
         }
     }
 
-    private static boolean hasPathPrefix(VaadinRequest request, String prefix) {
+    private static boolean hasPathPrefix(final VaadinRequest request, String prefix) {
         String pathInfo = request.getPathInfo();
 
         if (pathInfo == null) {
             return false;
         }
 
-        if (!prefix.startsWith("/")) {
-            prefix = '/' + prefix;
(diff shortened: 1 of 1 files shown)


knoobie:
I had never any issues with Vaadin+Spring behind a reverse proxy. 
(Running multiple instances in production) 

Here is an example apache conf snippet: 

```conf
  RedirectMatch ^/web$ /web/
  <Location /web/>
    ProxyPass                                ajp://your-url:8009/web/
    ProxyPassReverse                         ajp://your-url:8009/web/
  </Location>
```

in the application.yml

```yml
server:
  servlet:
    context-path: /web
```

cvs-orpheus:
Well https://github.com/vaadin/spring/issues/30 and https://vaadin.com/forum/thread/17895085/behind-reverse-proxy-uidl-request-urls-from-browser-incorrect suggest that there are still problems with reverse proxies.
In my case the proxy path is not the same as the internal path. Proxy requests to UIDL go to http://my-url/ext-frag1/ext-frag2/int-frag1/UIDL which is then redirected http://localhost:8085/int-frag1/UIDL
This was causing issues, because the UIDL requests were no longer handled by the UidlRequestHandler, because ServletPortletHelper checked if the prefix was "/UIDL/" but it was only "/UIDL".
Anyways, I would need to do the PR via a different account in order to be able to sign the cla. So if there is any chance you may include it, I will gladly do that, but if you do not think its necessary, I wont bother doing that.

knoobie:
@cvs-orpheus thanks for the additional information! Your links explain why I haven't seen the issue in our applications, because we always use the same external context-path and internal-context path. 

> Anyways, I would need to do the PR via a different account in order to be able to sign the cla. So if there is any chance you may include it, I will gladly do that, but if you do not think its necessary, I wont bother doing that.

I'm just a lurker - but I would think the chances are high that this could be added if it fixes the issue. (Needs some tests I guess)

Ansku:
I'm a bit worried that this solution would loosen the requirements too much and cause false positives. I'd rather see something like this after the original startsWith check (disclaimer, haven't tested it):

```
        if (!pathInfo.endsWith("/") && prefix.equals(pathInfo + "/")) {
            return true;
        }
```

clemensvonschwerin:
It has been a while, but please see https://github.com/vaadin/framework/pull/12087 for an update of this PR (with signed cla and more restrictive changes).

PR #11979, opened 2020-04-30
Fix RTA's CreateLink in Firefox & IE11

### LINKED ISSUE #11888: RichTextArea link doesn't work properly in Firefox or IE
In other browsers such as Chrome or IE, when the link button is clicked and link URL is given, the link appears in the RTA, its text being the URL. 

But when the same is done in Firefox 72.0.2 64-bit or IE 11.592, nothing happens. No link appears in the RTA. A link can only be entered by selecting text, and then clicking the link button.

Tested in [sampler](https://demo.vaadin.com/sampler/#ui/data-input/text-input/rich-text-area) and in production running Vaadin 8.

tarekoraby:
In Firefox and IE11, the 'Create Link' button of the RichTextArea (RTA) only works by turning some highlighted text into a link (by adding the inserted URI as the href property of the text). In that, the RTA in these two browsers behave similarly to the way it does in Chrome and Edge. However, in Firefox and IE11, clicking the 'Create Link' button has no effect if no text is pre-selected by the user. This is different from the button's behavior in Chrome and Edge where the user's provided URI is inserted, both, as the displayed text and its href property if no text is highlighted.

This fix enables the RTA's 'Create Link' button to work consistently across the supported browsers. Specifically, (and in addition to enabling adding the href property of a highlighted text), this fix enables Firefox and IE11 to also insert a new Uri as a text and its href property if no text is already highlighted.

fixes #11888

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/11979)

### DIFF
--- a/client/src/main/java/com/vaadin/client/ui/richtextarea/VRichTextToolbar.java
+++ b/client/src/main/java/com/vaadin/client/ui/richtextarea/VRichTextToolbar.java
@@ -242,7 +242,7 @@ public class VRichTextToolbar extends Composite {
                 final String url = Window.prompt("Enter a link URL:",
                         "http://");
                 if (url != null) {
-                    extended.createLink(url);
+                    createLinkViaJSNI(extended, url);
                 }
             } else if (sender == removeLink) {
                 extended.removeLink();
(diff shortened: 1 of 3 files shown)


Ansku (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/11979#-:-M6oJcJA7Qu__lkDeVJ9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12027, opened 2020-05-28
Row Generator fixes

andreybpanfilov:
This PR fixes following issues in row generator feature (https://vaadin.com/api/framework/7.7.17/com/vaadin/ui/Table.RowGenerator.html):

- Class Cast exception on client side when rendering components in generated rows (vaadin#2360)
- "Widget is still attached to the DOM after the connector ... has been unregistered. Widget was removed" errors, caused by superfluous calls of RowGenerator#generateRow method in com.vaadin.v7.ui.Table#parseItemIdToCells (rowGenerator.generateRow(...) and successive  registerComponent(...) poisons pageBuffer and visibleComponents)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12027)

### DIFF
--- a/compatibility-client/src/main/java/com/vaadin/v7/client/ui/VScrollTable.java
+++ b/compatibility-client/src/main/java/com/vaadin/v7/client/ui/VScrollTable.java
@@ -6735,7 +6735,8 @@ public class VScrollTable extends FlowPanel
                             addSpannedCell(uidl, cell.toString(), aligns[0], "",
                                     htmlContentAllowed, false, null, colCount);
                         } else {
-                            addSpannedCell(uidl, (Widget) cell, aligns[0], "",
+                            final ComponentConnector cellContent = client.getPaintable((UIDL) cell);
+                            addSpannedCell(uidl, cellContent.getWidget(), aligns[0], "",
                                     false, colCount);
                         }
                         break;
--- a/compatibility-server/src/main/java/com/vaadin/v7/ui/Table.java
+++ b/compatibility-server/src/main/java/com/vaadin/v7/ui/Table.java
@@ -2282,20 +2282,33 @@ public class Table extends AbstractSelect implements Action.Container,
             }
         }
 
-        GeneratedRow generatedRow = rowGenerator != null
-                ? rowGenerator.generateRow(this, id)
-                : null;
-        cells[CELL_GENERATED_ROW][i] = generatedRow;
+        int index = firstIndex + i;
+        int indexInOldBuffer = index - pageBufferFirstIndex;
+        boolean inPageBuffer = index < firstIndexNotInCache
+                && index >= pageBufferFirstIndex
+                && id.equals(pageBuffer[CELL_ITEMID][indexInOldBuffer]);
 
+        GeneratedRow generatedRow = null;
+        if (rowGenerator != null) {
+            if (inPageBuffer) {
+                generatedRow = (GeneratedRow) pageBuffer[CELL_GENERATED_ROW][indexInOldBuffer];
+            } else {
+                generatedRow = rowGenerator.generateRow(this, cells[CELL_ITEMID][i]);
+            }
+            cells[CELL_GENERATED_ROW][i] = generatedRow;
+        }
+
+        int firstNotCollapsed = -1;
         for (int j = 0; j < cols; j++) {
             if (isColumnCollapsed(colids[j])) {
                 continue;
+            } else if (firstNotCollapsed == -1) {
+                firstNotCollapsed = j;
             }
             Property<?> p = null;
             Object value = "";
-            boolean isGeneratedRow = generatedRow != null;
             boolean isGeneratedColumn = columnGenerators.containsKey(colids[j]);
-            boolean isGenerated = isGeneratedRow || isGeneratedColumn;
+            boolean isGenerated = isGeneratedColumn || generatedRow != null;
 
             if (!isGenerated) {
                 try {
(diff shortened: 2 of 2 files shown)


TatuLund:
Thanks for your contribution idea. We need to take a thorough look at it. However your PR is targeted to wrong branch. You should create PR against master.

andreybpanfilov:
Hi,

I have rebased PR against master branch. May I expect these changes in 8.x branch in case of positive decision?

Ansku:
Sorry for the delay. I've been reading through the changes and couldn't spot any major problems, and the validation build passed so that's definitely a good sign. I updated the branch so the validation build needs to complete again, and I still need to come up with a corresponding regression test, but looking good so far.

Ansku:
Found some issues in more thorough testing (RowGenerators test class breaks), will have to look into them tomorrow.

Ansku:
Turns out this isn't as trivial after all than I first thought (I'm not sure why I expected it to be, it's Table...). It needs a few tweaks to get it to work properly with the existing supported feature which is a Generated Row with either by-cell text or a single cell of text that spans the entire row, but it turns out that the use case for a component was never properly implemented even if the code does contain a few references for potential component contents. Even after these tweaks there are still hierarchy issues which on occasion cause the component contents to get unceremoniously detached from the generated row because they aren't recognized as proper children of the widget. Adding the generated rows works just fine, but some later hierarchy changes (e.g. by continuing to add rows to the Table one-by-one) cause things to go wonky for the older instances. And unfortunately I don't have time right now to figure out what else would need to be done to make the use case plausible.

andreybpanfilov:
@Ansku 

could you please clarify what scenario does not work property?

Ansku:
Continuing to add rows after you already have some. I made a tiny test Table and a row generator that affected every other row (first, third, etc.), and a button for adding one row at a time. Third row was already mostly out of view, and adding fourth row caused the two generated rows to empty their contents. When exactly it happens likely depends on Table height and the row heights.

Ansku:
Or if you meant the issues with the existing test class, I added back a regression test for it (#12069) and it's already available in the master branch, there the problem was within the new extractGeneratedValue which doesn't handle the spanned case correctly if the returned value is a String array with length 1 instead of a component.

andreybpanfilov:
@Ansku 

Thank you for clarification.

The issue with spanned String[1] was trivial, I have fixed it, please check updated PR.

The issue with broken row contents is exactly what this PR is purposed for, however if we want to return Component on GeneratedRow#getValue the implementation of RowGenerator/GeneratedRow should be a bit tricky: application developer must cache Component in GeneratedRow, i.e. naive implementation of GeneratedRow#getValue like:

```java
protected Object getValue() {
    return new Component(...);
}
```

will not work properly because Table calls this method multiple times for the same instance of GeneratedRow, moreover original implementation of Table calls RowGenerator#generateRow method multiple times for the same row, which makes it extremely hard to implement RowGenerator properly.

Ansku:
> application developer must cache Component in GeneratedRow

That's not really robust enough that I'd feel comfortable adding it as a new feature. If it's offered as an option it should be easier to use.

andreybpanfilov:
> That's not really robust enough that I'd feel comfortable adding it as a new feature. If it's offered as an option it should be easier to use.

That is how generated rows were implemented in vaadin, i.e. the minimal code application developer "supposed to" write is:

```java
table.setRowGenerator((tbl, itemId) -> new Table.GeneratedRow() {

    @Override
    protected boolean isSpanColumns() {
        return true;
    }

    @Override
    protected Object getValue() {
        ...
    }

});
```

in my opinion there is no big difference with something like:

```java

table.setRowGenerator((tbl, itemId) -> new Table.GeneratedRow() {

    private Component component = buildComponent();

    @Override
    protected boolean isSpanColumns() {
        return true;
    }
    
    @Override
    protected Object getValue() {
        return component;
    }

    private Component buildComponent() {
        ...
    }

});
```

if Table.GeneratedRow were an interface it would be possible to provide a couple of predefined implementations for application developer, but currently I do see only following option to make it more clear:

```patch
Index: compatibility-server/src/main/java/com/vaadin/v7/ui/Table.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- compatibility-server/src/main/java/com/vaadin/v7/ui/Table.java	(date 1597139054000)
+++ compatibility-server/src/main/java/com/vaadin/v7/ui/Table.java	(date 1597157960000)
@@ -6011,6 +6011,14 @@
         private boolean htmlContentAllowed = false;
         private boolean spanColumns = false;
         private String[] text = null;
+        // intentionally final
+        private final Component component;
+
+        public GeneratedRow(Component component) {
+            setHtmlContentAllowed(true);
+            setSpanColumns(true);
+            this.component = component;
+        }
 
         /**
          * Creates a new generated row. If only one string is passed in, columns
@@ -6022,6 +6030,7 @@
             setHtmlContentAllowed(false);
             setSpanColumns(text == null || text.length == 1);
             setText(text);
+            this.component = null;
         }
 
         /**
@@ -6040,7 +6049,7 @@
         }
 
         protected Object getValue() {
-            return getText();
+            return component == null ? getText() : component;
         }
 
         protected boolean isHtmlContentAllowed() {
```

Ansku:
Alright, we talked this through and decided we'll take the changes so that it's easier to make local adjustments, but we don't consider components within generated rows to be a supported feature as it is. There is quite a bit of potential for issues, both in how you add them and how they affect the Table in general. Table doesn't support variable row heights, for one thing, so the generated rows should still match other rows or things like scrolling _will_ go wonky. I updated the branch so need to wait for the validations to pass again before pushing it through.

Ansku (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12027#-:-MEXaWgoCmUcE4en1mC1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12039, opened 2020-06-23
Refactor how resync request is made

### LINKED ISSUE #11954: Error "Trying to start a new request while another is active" in browser while setting the theme, pushing, and adding a grid in the same request.
### Steps to reproduce

1. Enable @Push
2. In a click listener, do the following
2.1. Set the theme to an existing theme other than the current using `UI#setTheme`
2.2. Manually push using `UI#push`
2.3. Add a `Grid`

### Actual outcome

An error `Trying to start a new request while another is active` is displayed in the browser console.

I believe this is the root cause of sometimes infinite resynchronizations of the page in a real application, but I have not been able to reproduce it completely. The real app does not use manual pushing, but it does have push enabled (long polling) and it accesses the UI from background threads.

### Expected outcome

No errors in the console.

### Example code

After clicking the first button, the theme must be reset before the error occurs again.

    @Theme("mytheme")
    @Push
    
        @Override
            VerticalLayout layout = new VerticalLayout();
            UI ui = UI.getCurrent();
    
            Button button = new Button("Explosion button");
            button.addClickListener(e -> {
                // Must be a different theme, and its style files must exist
                ui.setTheme("my-theme");
                ui.push();
                layout.addComponent(new Grid<>());
            });
    
            Button reset = new Button("Reset theme");
            reset.addClickListener(e -> ui.setTheme("mytheme"));
            
            layout.addComponents(button, reset);
            setContent(layout);
        }
        
        }
    }

TatuLund:
When e.g. UIConnector is requesting resync after theme change, there may be pending requests in the queue.

Send resync request via sendInvocationsToServer(); instead.

This ensures that request is put properly in the queue and handled without conflict.

The same scenario can happen possibly in other ways too.

Fixes #11954

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12039)

### DIFF
--- a/client/src/main/java/com/vaadin/client/communication/MessageSender.java
+++ b/client/src/main/java/com/vaadin/client/communication/MessageSender.java
@@ -47,6 +47,7 @@ public class MessageSender {
 
     private ApplicationConnection connection;
     private boolean hasActiveRequest = false;
+    private boolean resynchronizeRequested = false;
 
     /**
      * Counter for the messages send to the server. First sent message has id 0.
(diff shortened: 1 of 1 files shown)


TatuLund:
Closed in favor of #12043

PR #12053, opened 2020-07-10
Correct example on https://vaadin.com/docs/v8/framework/gwt/gwt-extension.html

oliveryasuna:
Or a default constructor could be added.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12053)


Ansku:
Sorry for the delay in checking this out. You are actually only supposed to use one of the given examples, not copy-paste them all as a ready-made class. If you don't have the first option restricting your access to the default constructor the second two stand alone without issues. I made another PR to add a clarifying note to the example about that: #12062

PR #8861, opened 2017-03-16
Use computed style for Escalator size calculations

Artur-:
The old method of using getBoundingClientRect does not work as expected
if a transform has been applied to the element or one of its parents.
For instance PopupView animates itself using a scale(0) -> scale(1)
animation. When scale(0) is active, getBoundingClientRect will return 0
for all sizes while computed style ignores the transform and returns the
expected value.

Fixes #8793

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/8861)

PR #12138, opened 2020-10-27
Fix to LayoutManager size calculations during transform.

### LINKED ISSUE #11187: Expanding a component in a PopupView
I believe this issue is a bug in the PopupView Component and therefor I open a GitHub issue for this.
I already ask a question in StackOverflow, describing this issue:
https://stackoverflow.com/q/52296259/9687804

I am using Vaadin 8.5.2, and tested it with Chrome 68.0.3440.106 (64-bit) and Firefox 62.0 (64-bit).

I am using an embedded Tomcat Container (version 8.5.32) with Spring Boot (version 2.0.4).

Here you find a working code example to reproduce the issue: https://github.com/HappyLuke/vaadin-issue-1

I expect the button to be inside the PopupView, however, it is shown outside the PopupView. When expanding a tree component or changing the DateField value, the button will move into the PopupView as expected.

Thanks for any support!

Ansku:
- ComputedStyle is slower but more reliable than using
getBoundingClientRect, which does not work as expected if a transform
has been applied to the element or one of its parents. This is a problem
e.g. with PopupView, where getBoundingClientRect will return too small
size (or even zero size) for all the popup contents while the opening
animation is active. ComputedStyle ignores the transform and returns the
expected value.
- The presence of the element in DOM must be checked before the size is
requested from ComputedStyle, if the element has disappeared from DOM
without a warning and calculation is attempted anyway, the browser gets
stuck.
- Possibility to configure LayoutManager to use the less reliable
calculations for applications where the slight performance difference is
more important than layout issues within elements that have transform
animations.
- Manual test, problem isn't reproducible by TestBench.

Fixes: #11187

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12138)

### DIFF
--- a/client/src/main/java/com/vaadin/client/LayoutManager.java
+++ b/client/src/main/java/com/vaadin/client/LayoutManager.java
@@ -69,6 +69,7 @@ public class LayoutManager {
         }
     };
     private boolean everythingNeedsMeasure = false;
+    private boolean thoroughSizeCheck = true;
 
     /**
      * Sets the application connection this instance is connected to. Called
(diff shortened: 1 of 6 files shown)


Ansku:
Related: #8861

OlliTietavainenVaadin (approved):
:lgtm:

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12138#-:-MOpZ7dn4O64KBbKFyJR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12145, opened 2020-11-06
Tweaks to Grid/Escalator column size handling

### LINKED ISSUE #12139: Grid with lots of columns + setMinimumWidthFromContent + setHeightByRows fails with 8.12.0
Vaadin Framework version 8.12.0
Latest Chrome/Firefox browser versions
Liferay CE version 7.3.5 GA6

A grid with columns having setMinimumWithFromContent set to false to see ellipsis if content exceends column width.
So many columns that content exceends column width.
Grid's setHeightByRows set to exact number of data rows.

With Vaadin version <= 8.11.3 grid was shown as expected.
With 8.12.0 the grid is not shown, please wait spinner stays rolling.

Please see below a sample code that fails with 8.12.0 (and works with 8.11.3). 20 columns was well enough for my browser window size.

```
    @Override
    protected void init(VaadinRequest request) {
        
        Random random = new Random();
        
        List<DummyGridRow> gridRows = new ArrayList<DummyGridRow>();
        gridRows.add(new DummyGridRow(random));
        
        Grid<DummyGridRow> grid = new Grid<DummyGridRow>();
        for (int i = 0; i < 20; i++) {
            grid.addColumn(DummyGridRow::getValue)
            .setCaption("[" + i + "] Quite dummy column")
            .setMinimumWidthFromContent(false);
        }
        
        grid.setItems(gridRows);
        grid.setHeightByRows(gridRows.size());
        grid.setWidthFull();

        this.setContent(grid);
        
    }

    class DummyGridRow {
        
        private Random random = null;
        
        public DummyGridRow(Random random) {
            this.random = random;
        }
        
        public int getValue() {
            return this.random.nextInt(1000000000);
        }
        
    }

```

Ansku:
- ScrollbarBundle: removed delays in scroll handling that were only
needed for IE8, added possibility to update offsetSize and scrollSize at
the same time in order to avoid triggering unnecessary scrollbar
visibility change events during the intermediate state.
- ColumnConfigurator: added new method that allows setting column widths
without triggering element size recalculations.
- EscalatorProxy: added implementation of the new method to
ColumnConfigurationProxy.
- Escalator: switched to use new methods in ScrollbarBundle and
ColumnConfigurator, added a pixel to a scrollbar offsetSize calculation
that was for some reason consistently one pixel too low, removed
duplicate method calls from sectionHeightCalculated handling as those
are already handled by the calling method and can cause incorrect
intermediate state and unnecessary scrollbar visibility change events,
added implementation of the new method to ColumnConfigurationImpl with
the element size recalculations made optional.
- Grid: updated column minimum width calculations to take into account
the potential presence of a resize handle, updated expand ratio handling
to not trigger element size recalculations until the entire handling is
finished.
- Test for column width handling when there are multiple columns with
setMinimumWidthFromContent(false)

Fixes #12139

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12145)

### DIFF
--- a/client/src/main/java/com/vaadin/client/widget/escalator/ColumnConfiguration.java
+++ b/client/src/main/java/com/vaadin/client/widget/escalator/ColumnConfiguration.java
@@ -143,7 +143,8 @@ public interface ColumnConfiguration {
     public double getColumnWidth(int index) throws IllegalArgumentException;
 
     /**
-     * Sets widths for a set of columns.
+     * Sets widths for a set of columns. Triggers element size recalculation for
+     * elements that require manual calculations.
      *
      * @param indexWidthMap
      *            a map from column index to its respective width to be set. If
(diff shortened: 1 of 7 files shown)


edler-san (commented):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12145)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12168, opened 2020-12-03
Add API to prevent invalid input when integrated range validator is used in DateField

### LINKED ISSUE #12163: DatePicker.setRangeStart/setRangeEnd doesn't prevent the user from typing in an invalid date
I have the following DatePicker:

```
        final DateField field = new DateField("Type in 31.12.20000, then press TAB. The 31.12.20000 shouldn't pass the validator, but it does and goes to server");
        field.setRangeStart(LocalDate.of(2000, 1, 1));
        field.setRangeEnd(LocalDate.of(2999, 1, 1));
        field.addValueChangeListener(e -> {
            System.out.println("Got date " + e.getValue());
        });
```

I can type in an invalid date such as 31.12.20000 and the date will get transmitted server-side in an value-changed listener, even though it's invalid.

This has a devastating after-effect of not being able to set the very same date back to the `DateField` via `setValue()` (e.g. when using Binder), since that would fail with

```
Caused by: java.lang.IllegalArgumentException: value is not within acceptable range
at com.vaadin.ui.AbstractDateField.setValue(AbstractDateField.java:717)
at com.vaadin.ui.AbstractDateField.setValue(AbstractDateField.java:81)
```

- Vaadin Framework version 8.12.0
- Browser version (Firefox 83.0)
- Web container name and version (Tomcat 9.0.31)

* Expected behavior: the invalid value is not transferred server-side
* Actual behavior: the value change listener is called with the invalid value

TatuLund:
Fixes: #12163 

API naming setPreventInvalidInput(..) was selected based on similar use case in Vaadin 14 TextField pattern validation.

This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12168)


Ansku (approved):
> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12168)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12194, opened 2021-02-02
Update gwt to version 2.9.0

jameshilliard:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12194)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
         <liferay.portal-service.version>6.2.5</liferay.portal-service.version>
         <liferay.portal-kernel.version>2.0.0</liferay.portal-kernel.version>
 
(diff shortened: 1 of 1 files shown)


TatuLund:
@jameshilliard GWT 2.9.0 introduces some backwards breaking API changes, not much, but nevertheless. Thus it is not sufficient to just update the dependency. As these API changes break some functionality we currently like to use, we are not going to update to GWT 2.9.0 yet, but most likely later due other reasons. Before that we will add documentation how to use GWT 2.9.0 in your Vaadin project.

TatuLund:
See details here: https://github.com/TatuLund/gwt290-demo

TatuLund:
More info in blog here: https://vaadin.com/blog/update-your-vaadin-8-project-to-java-11-and-gwt-2.9.0

PR #12214, opened 2021-03-03
update to Atmosphere runtime 2.4.30.vaadin3

### LINKED ISSUE #12205: Vaadin Atmosphere (and eventually push) broken in OSGi environment since 8.12.2
Vaadin Atmosphere runtime dependency version changes from 2.4.30.vaadin1 to 2.4.30.vaadin2. However, the OSGi-Manifest for the latter delclares all Export-Package still being version 2.4.30.vaadin**1**. Vaadin Server artifact declares required version 2.4.30.vaadin**2** for all Import-Package regarding Athmosphere. The net result ist, no matching packages being found. Thus, no push since 8.12.2. There is no OSGi resolution error because all Atmosphere dependencies are optional (that's O.K.)

How to reproduce: Try to use push while running Vaadin server in an OSGi container.

Pull request? I did not find out where the version information for the manifest actually comes from. It does not seem to be in the source code (i.e. Maven poms).

Why is the build-suffix used for matching anyway? Would not omitting it have been simpler? Most OSGi bundles use version ranges for package import in order to introduce some flexibility. You even check the real version at runtime anyway. It would at least not have lead to such a hidden issue (I needed about 3h to dig for the reason).

OlliTietavainenVaadin:
Update Atmosphere runtime version to 2.4.30.vaadin3. Fixes #12205.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12214)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
         <javax.portlet.version>2.0</javax.portlet.version>
         <vaadin.sass.version>0.9.13</vaadin.sass.version>
         <!-- Note that this should be kept in sync with the class Constants -->
-        <atmosphere.runtime.version>2.4.30.vaadin2</atmosphere.runtime.version>
+        <atmosphere.runtime.version>2.4.30.vaadin3</atmosphere.runtime.version>
 
         <!-- OSGi -->
         <osgi.execution.environment>JavaSE-1.8</osgi.execution.environment>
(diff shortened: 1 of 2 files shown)


Ansku (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12214#-:-MUs3Gam0CrJ5rVOESom:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12218, opened 2021-03-05
Prevent NPE in comparison

### LINKED ISSUE #12207: #12132 leads to NPEs
With Vaadin 8.12.1, we get NPE Exceptions in code introduced in #12132.

As also pointed  out in https://github.com/vaadin/framework/pull/12132#discussion_r576280435.

Though it is rather unlikely that a `field_value == null`  leads to a `model_value != null`, there might be cases this happens.

We encountered null -> Optional.empty -> null, but chains like null -> default value -> null might also be possible.
This also questions the null check of `value` in line https://github.com/vaadin/framework/blob/949434a935638cf130535f9d4bf2e6702e2c0a00/server/src/main/java/com/vaadin/data/Binder.java#L1287

Maybe the solution is:
```java
FIELDVALUE converted = convertToFieldType(value);
if (!Objects.equals(field.getValue(), converted)) {
    getField().setValue(converted);
}
```

The code leads to "immediate" normalization of values in the gui by running the conversion back and forth and makes it more difficult to allow users to enter values that have multiple representations e.g. enter decimal values too slow with StringToDoubleConverter and default TextFields is quite challenging when typing slowly. :wink:

TatuLund:
Fixes: https://github.com/vaadin/framework/issues/12207

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12218)

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -1286,7 +1286,7 @@ public class Binder<BEAN> implements Serializable {
                     setter.accept(bean, value);
                     if (value != null) {
                         FIELDVALUE converted = convertToFieldType(value);
-                        if (!field.getValue().equals(converted)) {
+                        if (!Objects.equals(field.getValue(), converted)) {
                             getField().setValue(converted);
                         }
                     }


Ansku (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12218#-:-MV1GL509I0vmSEjZSJ8:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12232, opened 2021-03-08
Fix updating Escalator's physical row index.

Ansku:
- Physical index should take into account that more than one row can
have a spacer. Physical index has no practical effect on how Grid looks
like or behaves, since Escalator by design operates independently of it
and the visual positioning of elements is based on the logical and
visual indices. Physical index only affects the order of elements
within the DOM tree, where a mismatch can cause confusion during
debugging.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12232)


OlliTietavainenVaadin (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vaadin/framework/12232#-:-MVH2no815HA3CNoaiMl:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #12242, opened 2021-03-12
Add detection of HCL Digital Experience rebranding

vjt:
Dear Vaadin team,

On HCL Digital Experience 8.5.5 CF19, the `getPortalInfo()` method returns `"hcl digital experience/8.5"`, breaking detection of the servlet engine. This ultimately leads to methods such as `getHeader()` to return NULL, as the upstream HTTP request is not retrieved by the Vaadin abstraction layer.

In our case, the issue became evident with the MultiFileUploader components, that could not read the `Content-Length` header contents.

Below a debugging session screenshot demonstrating the issue.

Thank you for your consideration.

hcl-portal-vaadin-debug-session

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/framework/12242)


DanieleVistalli:
I confirm the issue. HCL Software acquired the former WebSphere Portal product.

As part of the rebranding the portalInfo string changed and the string based detection is failing. This has impact on any customer updating to HCL DX 8.5 and 9.X with CF19 (Cumulative fix 19)

TatuLund:
Yes, this seems to be already quite old news actually, ref. https://www.hcltech.com/software/acquisition-updates

TatuLund:
@vjt The change looks legit, but have you tested it someway. Luckily createVaadinRequest is protected so you can override that with your own implementation that does the same thing as your fix.

Just pointing out that there is WebSphere Portlet specific code here too: https://github.com/vaadin/framework/blob/master/server/src/main/java/com/vaadin/server/VaadinPortlet.java#L330

Assuming that HCL has not done any actual code level changes, e.g. package renaming, you proposed change could be enough.

vjt:
Hi @TatuLund,

I have tested Vaadin works correctly on a CF19 portal by modifying the portal identity string, so that it returns again `ibm websphere portal`, by unarchiving `PortalServer/base/wp.base/shared/app/wp.base/wp.base.jar` and editing `com/ibm/wps/Version.properties` and changing `server.name`.

Internally we are moving in the direction you pointed out, overriding the Vaadin method and applying this change - I anyway believe it would be valuable for Vaadin to support Portal CF19.

Eventually - my personal opiniion is that HCL should revert this change, as I do not understand breaking software integration compatibility if there are no technical reasons to do so, and without a widespread "breaking change" announcement.

Thanks

DanieleVistalli:
> @vjt The change looks legit, but have you tested it someway. Luckily createVaadinRequest is protected so you can override that with your own implementation that does the same thing as your fix.
> 
> Just pointing out that there is WebSphere Portlet specific code here too: https://github.com/vaadin/framework/blob/master/server/src/main/java/com/vaadin/server/VaadinPortlet.java#L330
> 
> Assuming that HCL has not done any actual code level changes, e.g. package renaming, you proposed change could be enough.

Package renaming didn't happen and is not going to happen (it would break a lot more than Vaadin)

I talked to HCL development and they confirmed the string change due to IBM to HCL rebranding.

The code fix in this PR will make it. If you need a confirmation I can have my team do/test it but it just looks like the fix is just correct.

TatuLund:
@DanieleVistalli Thanks for quick feedback and we appreciate your contribution on this. We simply cannot keep our eyes on everything that moves, and in Java ecosystem there are so many things that moves.

> I can have my team do/test it but it just looks like the fix is just correct.

This would be of course be something that would be greatly appreciated. The code change is neutral in the sense that it wont break anything, but I would assume it would be also nicer for you to get update that surely works, rather than need to re-patch and wait again for the next release.

DanieleVistalli:
I will test this shortly. And provide confirmation by Monday next week

TatuLund:
@DanieleVistalli Did you manage to test it?

TatuLund:
@DanieleVistalli did you have some outcome from your testing?

Ansku (approved):

vjt:
🎉

Issue #12352, opened 2021-07-28
Deadlock can happen during findOrCreateVaadinSession 


TatuLund:
It is possible findOrCreateVaadinSession leaves session locked if HTTP session is invalidated in an another thread between lockSession and unlockSession calls.

Stack trace

2021-05-30 21:29:25,710 SEVERE [org.atmosphere.cpr.AsynchronousProcessor] (bgp-ui I/O-3) Failed to cancel resource: 38b74984-d931-4b90-a052-c3895741b4b6: java.lang.IllegalStateException: UT000010: Session is invalid RSNdMrrtfo2YH3dopUaf0N-irD6miiALGuQPIC9D
   at io.undertow.core@2.0.27.Final//io.undertow.server.session.InMemorySessionManager$SessionImpl.getAttribute(InMemorySessionManager.java:512)
   at io.undertow.servlet@2.0.27.Final//io.undertow.servlet.spec.HttpSessionImpl.getAttribute(HttpSessionImpl.java:122)
   at deployment.ica.war//com.vaadin.server.WrappedHttpSession.getAttribute(WrappedHttpSession.java:54)
   at deployment.ica.war//com.vaadin.server.VaadinService.getSessionLock(VaadinService.java:587)
   at deployment.ica.war//com.vaadin.server.VaadinService.unlockSession(VaadinService.java:656)
   at deployment.ica.war//com.vaadin.server.VaadinService.findOrCreateVaadinSession(VaadinService.java:675)
   at deployment.ica.war//com.vaadin.server.VaadinService.findVaadinSession(VaadinService.java:530)
   at deployment.ica.war//com.vaadin.server.communication.PushHandler.connectionLost(PushHandler.java:321)
   at deployment.ica.war//com.vaadin.server.communication.PushAtmosphereHandler.onStateChange(PushAtmosphereHandler.java:60)
   at deployment.ica.war//org.atmosphere.cpr.AsynchronousProcessor.invokeAtmosphereHandler(AsynchronousProcessor.java:509)
   at deployment.ica.war//org.atmosphere.cpr.AsynchronousProcessor.completeLifecycle(AsynchronousProcessor.java:459)
   at deployment.ica.war//org.atmosphere.cpr.AsynchronousProcessor.endRequest(AsynchronousProcessor.java:566)
   at deployment.ica.war//org.atmosphere.websocket.DefaultWebSocketProcessor.executeClose(DefaultWebSocketProcessor.java:679)
[...]

PR #12353, opened 2021-07-28
fix: Prevent deadlock in findOrCreateVaadinSession

TatuLund:
It is possible findOrCreateVaadinSession leaves session locked if HTTP session is invalidated in an another thread between lockSession and unlockSession calls. This PR changes the implementation so that this can no longer happen.

fixes: https://github.com/vaadin/framework/issues/12352


server/src/main/java/com/vaadin/server/VaadinService.java
```diff
@@ -611,7 +611,7 @@ protected Lock getSessionLock(WrappedSession wrappedSession) {
      * @throws IllegalStateException
      *             if the session is invalidated before it can be locked
```
Legioth, line 612: `@return`

server/src/main/java/com/vaadin/server/VaadinService.java
```diff
@@ -611,7 +611,7 @@ protected Lock getSessionLock(WrappedSession wrappedSession) {
...
      */
-    protected void lockSession(WrappedSession wrappedSession) {
+    protected Lock lockSession(WrappedSession wrappedSession) {
```
Legioth, line 614: Could maybe add a code comment or Javadoc mention of the previous version of this method so that someone who gets errors on upgrading can find an explanation by looking at code also and not only by reading release notes.

server/src/main/java/com/vaadin/server/VaadinService.java
```diff
@@ -674,7 +676,7 @@ private VaadinSession findOrCreateVaadinSession(VaadinRequest request)
             return doFindOrCreateVaadinSession(request,
                     requestCanCreateSession);
         } finally {
-            unlockSession(wrappedSession);
```
Legioth, line 677: Should the now unused `unlockSession` method be removed or deprecated? Removing would be more in line with the approach taken with `lockSession`, i.e. to break it properly.

Another alternative would be to keep it as `unlockSession(WrappedSession, Lock)` so that a totally custom implementation might still override the way unlocking happens in a practical way.

server/src/main/java/com/vaadin/server/VaadinService.java
TatuLund: Yes, I will add one.

server/src/main/java/com/vaadin/server/VaadinService.java
TatuLund: Yes, it was in my plans to add also some note that override of this method is not recommended

server/src/main/java/com/vaadin/server/VaadinService.java
TatuLund: Good pick. It did not notice that findOrCreateVaadinSession was the sole user of unlockSession.

server/src/main/java/com/vaadin/server/VaadinService.java
Legioth: There's nothing that explains the migration scenario, but only a new unrelated comment that blames the user for doing something that might not be right, based on the assumption that it's only about lock storage but not about how the actual locking is performed.

server/src/main/java/com/vaadin/server/VaadinService.java
Legioth: Shouldn't that new method be used here as well? Otherwise, it's quite pointless to change the signature and the implementation.

server/src/main/java/com/vaadin/server/VaadinService.java
```diff
@@ -641,21 +650,27 @@ protected void lockSession(WrappedSession wrappedSession) {
...
      *
      * @param wrappedSession
-     *            The session to unlock
+     *            The session to unlock, used only with assert
+     * @param lock
+     *            Lock instance to unlock
      */
-    protected void unlockSession(WrappedSession wrappedSession) {
+    protected void unlockSession(WrappedSession wrappedSession, Lock lock) {
         assert getSessionLock(wrappedSession) != null;
```
Legioth, line 671: This assert seems quite pointless now, and would make the application throw without unlocking if assertions are enabled in the scenario that we're trying to fix.

server/src/main/java/com/vaadin/server/VaadinService.java
```diff
@@ -641,21 +650,27 @@ protected void lockSession(WrappedSession wrappedSession) {
...
         }
+        return lock;
     }
 
     /**
      * Releases the lock for the given session for this service instance.
      * Typically you want to call {@link VaadinSession#unlock()} instead of this
      * method.
+     * <p>
+     * Note: Overriding this method is not recommended, for custom lock storage
```
Legioth, line 661: This part could also have a note about the changed signature

server/src/main/java/com/vaadin/server/VaadinService.java
TatuLund: Yes, sort of. If I drop it, having WrappedSession will be un-used parameter. But could be point to preserve that in case of someone has indeed overriden this method and used WrappedSession for some other purpose than obtaining the Lock instance.

server/src/main/java/com/vaadin/server/VaadinService.java
Legioth: Still unresolved

Legioth (approved):
LGTM

PR #12360, opened 2021-08-06
fix: Avoid processing value change event due writing back of converted value

### LINKED ISSUE #12356: Two fields with converter in one binder cause ConcurrentModificationException
Vaadin version: 8.13.1
Browser: any, e.g. Chrome Version 92.0.4515.107 (Official Build) (64-bit)
Container: TomEE, J2EE

**Steps to reproduce:**
1. Let's have a Vaadin screen
2. Create Binder<BEAN> for a BEAN 
2. Add bindings to 2 fields, each of them is TextField<BigDecimal>
3. Add a numeric converter to each of the TextFields bindings (to easily cause validation error)
4. Set the bean to the binder
5. Open Vaadin application
6. Set both text fields to non-numeric value (this causes Vaadin not to run converters)
7. Set first field to numeric value (still no converter is executed, because there is still one validation error left on the other field)
8. Set second field to numeric value.

**And the problem happens:**
1. Vaadin now identifies, there are no validation errors and tries to set binding values back to the BEAN. It iterates over "changedBinding" collection (Binder line 2035) and executes writeFieldValue. It also detects, there is convertBackToPresentation flag turned on, so it calls setValue on TextBox (!!!). 
2. This setValue call fires another event, which makes Vaadin execute Binder one more time.
3. So the Binder once again starts to iterate over "changedBinding" collection. When it is done, it clears the changed bindings collection (line 2047, which seems safe, but the parameter is actually changedBindings reference), which has 2 open iterators and boom, concurrent modification exception.

It seems the problem is that changedBindings is modified in recursive call (caused by converters) or that the reference is directly provided as parameter of the call (Binder, line 1569, but thre is a clear reason to pass it directly).
The clean solution seems to be some kind of ability to call setField without raising another event (e.g. doSetValue which directly modifies component state, like in AbstractTextField).

**Workaround:**
Use separate binder per field, if applicable. If you need bean validator or want to modify bean only when all fields are valid, workaround cannot be used.

TatuLund:
This is both a optimization by skipping duplicate validation round and avoids ConcurrentModificationExpectation being thrown certain corner cases.

Fixes: https://github.com/vaadin/framework/issues/12356

### DIFF
--- a/server/src/main/java/com/vaadin/data/Binder.java
+++ b/server/src/main/java/com/vaadin/data/Binder.java
@@ -1103,6 +1103,7 @@ public class Binder<BEAN> implements Serializable {
 
         private Registration onValueChange;
         private boolean valueInit = false;
+        private boolean convertedBack = false;
 
         /**
          * Contains all converters and validators chained together in the
@@ -1283,7 +1284,8 @@ public class Binder<BEAN> implements Serializable {
         private void handleFieldValueChange(
                 ValueChangeEvent<FIELDVALUE> event) {
             // Don't handle change events when setting initial value
-            if (valueInit) {
+            if (valueInit || convertedBack) {
+                convertedBack = false;
                 return;
             }
 
@@ -1313,6 +1315,7 @@ public class Binder<BEAN> implements Serializable {
                     if (convertBackToPresentation && value != null) {
                         FIELDVALUE converted = convertToFieldType(value);
                         if (!Objects.equals(field.getValue(), converted)) {
+                            convertedBack = true;
                             getField().setValue(converted);
                         }
                     }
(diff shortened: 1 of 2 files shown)


Ansku (approved):

ladislavlencucha:
Such a simple solution for such a huge issue. Thanks! You made my day by fixing it so fast!
Any news when 8.13.3 is going to be released?

Thanks again!

Ansku:
Our releases team has already left for the weekend so can't confirm the schedule yet, but the fix is already in 8.14-SNAPSHOT.

Ansku:
There are some other releases that we need to get out first, but currently looks like 8.13.3 could happen towards the end of the week if everything goes well.

PR #12391, opened 2021-08-27
Reduce test flakiness by adding a retry to initial browser resize.

Ansku:


Ansku:
Ineffective as suspected.

PR #12416, opened 2021-10-04
chore: avoid repackaging of google elemental dep and use it directly

denis-anisimov:
The google elemental dependency is not OSGi compatible so it's repackaged
into shared module to use it directly from there. But there is a custom Vaadin version
of google elemental dep and it's OSGi compatible. Flow uses this version and to be able
to use FW in MPR the versions should match. So the Vaadin google elemental dep is added
instead of repackaging.

Thanks for submitting a pull request!
Please confirm that your pull request follows our guidelines found in CONTRIBUTING and that you provide enough information so that we can review your pull request.

<Description of pull request, e.g. "Fix Grid Column not sortable with backend data and sort property">

Fixes #<replace with an issue number>

[ ] Valid tests for the pull request
[ ] Contributing guidelines implemented

### DIFF
--- a/shared/bnd.bnd
+++ b/shared/bnd.bnd
@@ -3,7 +3,6 @@ Bundle-Activator: com.vaadin.osgi.resources.OsgiVaadinResources
 Bundle-Name: Vaadin Shared
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-License: http://www.apache.org/licenses/LICENSE-2.0
-Import-Package: org.osgi*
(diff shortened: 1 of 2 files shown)


TatuLund (approved):

rkovarik:
This change was not visible from https://vaadin.com/download/release/8.14/8.14.2/release-notes.html.
We used to have _gwt-elemental_ as provided dependency so it was not brought into webapps, causing _ClassNotFoundExceptions_ after upgrade.

PR #12421, opened 2021-10-06
Fix for setting dataproviders when overriding getDataProvider

### LINKED ISSUE #12394: CallbackDataProvider cannot be cast to custom DataProvider
We tried to move from Vaadin 8.8.5 to 8.13.3 and encountered the following issue:
The method **internalSetDataProvider** of the class **com.vaadin.ui.Grid** has changed and does not work with our custom **DataProvider** anymore.

Method **internalSetDataProvider** from the specific Vaadin 8 Versions:
**Vaadin 8.8.5 (method is still the same up to 8.12.4)**
```
@Override
protected void internalSetDataProvider(DataProvider<T, ?> dataProvider) {
	super.internalSetDataProvider(dataProvider);
	for (Column<T, ?> column : getColumns()) {
		column.updateSortable();
	}
}
```
**Vaadin 8.13.0 and newer (up to 8.13.3)**
```
@Override
protected void internalSetDataProvider(DataProvider<T, ?> dataProvider) {
	boolean newProvider = getDataProvider() != dataProvider;
	super.internalSetDataProvider(dataProvider);
	if (newProvider) {
		Set<T> oldVisibleDetails = new HashSet<>(
				detailsManager.visibleDetails);
		oldVisibleDetails.forEach(item -> {
			// close all old details even if the same item exists in the new
			// provider
			detailsManager.setDetailsVisible(item, false);
		});
	}
	for (Column<T, ?> column : getColumns()) {
		column.updateSortable();
	}
}
```

Explanation why I think this is a bug:
The method **internalSetDataProvider** does allow us to set a custom data provider of the interface type **DataProvider**.
But when I provide a **DataProvider** which does not implement **CallbackDataProvider** or **AbstractBackEndDataProvider** I get an exception:
```
Caused by: java.lang.ClassCastException: com.vaadin.data.provider.CallbackDataProvider cannot be cast to net.example.JavaListValueDataProvider
```
I assume this is because the check does try to compare the new **DataProvider** with the default one and fails the cast of the data types.
```
boolean newProvider = getDataProvider() != dataProvider;
```

Our **DataProvider** class looks like the following:
As you can see it does implement **DataProvider** over **AbstractDataProvider** but cannot be cast to **CallbackDataProvider**
```
public class JavaListValueDataProvider<T extends JavaValueDataInterface>
extends
	AbstractDataProvider<T, SerializablePredicate<T>>
implements
	InMemoryDataProvider<T> {
...
}
```

pinneri:
Fixes #12394

Do not call an overridable method from the constructor.

### DIFF
--- a/server/src/main/java/com/vaadin/ui/Grid.java
+++ b/server/src/main/java/com/vaadin/ui/Grid.java
@@ -4997,7 +4997,7 @@ public class Grid<T> extends AbstractListing<T> implements HasComponents,
 
     @Override
     protected void internalSetDataProvider(DataProvider<T, ?> dataProvider) {
-        boolean newProvider = getDataProvider() != dataProvider;
+        boolean newProvider = internalGetDataProvider() != dataProvider;
         super.internalSetDataProvider(dataProvider);
         if (newProvider) {
             Set<T> oldVisibleDetails = new HashSet<>(


TatuLund:
Thanks for your contribution.

This PR is created against wrong branch. We will collect all the contributions to master branch first. And after that we do pick them in release branch. Moreover 8.13 release branch is no longer in use.

Also this one needs more study before acceptance as well as a unit test would be preferable.

PR #12423, opened 2021-10-10
When setting the dataprovider in the Grid constructor do not rely on …

pinneri:
…an overridable method. Fixes #12421

Fixes ##12421


TatuLund:
@pinneri We can't proceed with the PR review before you have singed the CLA.

PR #12466, opened 2021-11-02
feat: Add better API to configure maximum allowed rows

TatuLund:
Some components are harder to extend than Grid for overriding max rows limit. However getDataCommunicator is a public method and thus this API would make overriding the limit easier, see example

```
@Theme("mytheme")

    @Override
        
        TwinColSelect<String> twin = new TwinColSelect<>();
        twin.getDataCommunicator().setMaximumAllowedRows(1000);

        ComboBox<String> combo = new ComboBox<>();
        combo.getDataCommunicator().setMaximumAllowedRows(1000);
        
        List<String> items = new ArrayList<>();
        for (int i=0;i<800;i++) {
        	items.add("Item "+i);
        }
        
        twin.setItems(items);
        twin.setItemCaptionGenerator(item -> item.toUpperCase());
        combo.setItems(items);
        combo.setScrollToSelectedItem(true);
        
        layout.addComponents(twin, combo);
        
        setContent(layout);
    }

    }
}

```


server/src/test/java/com/vaadin/data/provider/DataCommunicatorTest.java
```diff
@@ -322,4 +322,12 @@ public void requestTooMuchRowsFail() {
         communicator.onRequestRows(0, communicator.getMaximumAllowedRows() + 10,
                 0, 0);
     }
+
+    @Test
+    public void requestTooMuchRowsOverride() {
+        TestDataCommunicator communicator = new TestDataCommunicator();
```
OlliTietavainenVaadin, line 328: I think this should be tested with an actual Component to make sure the new limit can used for the first load

OlliTietavainenVaadin (approved):

PR #12652, opened 2025-07-01
Switched from java.io to  java.nio for safer temp file generation

edler-san:
Following our own best practices, we should no longer use the older java.io to create temporary files as it uses less safe defaults than java.nio's implementation.


edler-san:
Will re-create the PR there.

PR #12684, opened 2026-09-11
Update CHANGELOG-VAADIN7.md

mlindfors:


CHANGELOG-VAADIN7.md
```diff
@@ -1,5 +1,13 @@
 # Vaadin 7 extended maintenance version changelog
 
+## Vaadin 7.7.54
+
+* The jsoup version being used has been updated to 1.23.2. This new version will change how it handles self-closing tags for components.
```
thevaadinman: Mention the CVE. Also, the new version will not change tag handling in the future - it does not allow self-closing tags in HTML contexts, while the old version allowed them. This can cause some test code or such to no longer work for users. There's also nothing we can do about it.

thevaadinman (changes requested):

thevaadinman (approved):

